PR #165, opened 2014-12-29
Add toggle fullscreen textarea extension

dongweiming:
I add this extension for edit current cell more focused: it can be full screen and hide the other elements. i also add a toggle shortcut `Alt + l`


jcb91:
ok, I'm going to close this as it's now over 18 months old, and the repository's been through some major refactors in that time. Re-open or create anew as appropriate.

Issue #187, opened 2015-02-08
FontAwesome icons depracated and need updating


zach-hartwig:
Hello,

I've just cloned the most recent commit. I'm using the 'hide_input_all.js' and 'read-only.js' extensions, which work great. However, the icons for these extensions show up blank on my IPython Notebook It appears that the icon names need to be updated to FontAwesome4.2.0. For example, in 'read-only'.js, the current code attempts to specify a lock icon with:
```python
icon : 'icon-lock',
```
when for FontAwesome 4.2.0 it should be
```python
icon : 'fa-lock',
```
Making this change gets the icon working perfectly. I only use these two icon-based extensions presently, but, in general, all icons may need updating.

juhasch:
Yes, this has been changed in IPython. Can you make a pull request with the required changes ?

PR #188, opened 2015-02-11
Updating all icons to be compliant with FontAwesome 4.3.0

zach-hartwig:
Many of the icons needed updating to be compliant with IPython's implementation of the FontAwesome 4.X.X series. Those that have not been updated simply show up as blank in the Notebook toolbar. I have gone through the code and updated all icons to the FA 4.3.0 series (the most recent), attempting to keep the graphical images as closely as intended by the original authors. This work done in response to Issue #187. I have verified that these icons work on a subset of all changes made.


juhasch:
Thank you, looks good to me.

PR #190, opened 2015-02-16
fix slidemode for 3.0

Carreau:
Most of the change are indenting to put in a module. 

ping @fperez that requested that.

### DIFF
--- a/slidemode/README.md
+++ b/slidemode/README.md
@@ -1,0 +1,16 @@
+## Legacy Slidemode. 
+
+This folder contains the legacy slidemode
+
+Put the following in your custom.js
+
+
+```
+// Slidemode
+require(['nbextensions/IPython-notebook-extensions/slidemode/main','base/js/events'], function(slidemode, events){
+    events.on('app_initialized.NotebookApp', function(){
+        slidemode.init();
+    });
+});
+// end slidemode
+```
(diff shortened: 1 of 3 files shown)


fperez:
Thanks! Can confirm that the fix works... 

I'd only comment, in the README, that the correct path, if installed in the standard location, is:

```
require(['custom/slidemode/main','base/js/events'], function(slidemode, events){
```

If the full path is put in the `custom.js` file (as currently shown in the README), users will likely see a 404.

In any case, with that small tweak to the README, it's all good. Thanks again!

Carreau:
> I'd only comment, in the README, that the correct path, if installed in the standard location, is:

Well, probably not, as the standard location for extension is not in `profile/custom` anymore, so that extensions can be shared across profile. 

The simplest is for people to clone this repo in their `/nbextension`dir and just pull for update.

fperez:
ok. I can confirm it worked pretty well, though our default CSS is now making lots of whitespace that makes these slideshows sub-optimal without extra tweaking. But that's a separate story. Thanks!

damianavila:
`<hijacking>`[RISE](https://github.com/damianavila/live_reveal)`</hijacking>` :stuck_out_tongue_winking_eye:

damianavila:
Btw, I think this can be merged... so... merging now...

PR #193, opened 2015-03-01
Update README.md

dsblank:
Now that 3.x is released, can change reference back to "checkout 3.x"


juhasch:
Yes, and we should have a 3.x branch, too.

PR #194, opened 2015-03-07
Conda recipe for installing notebook extensions

juhasch:
This recipes allows to create a conda package for the automatic installation of notebook extensions.
It expects all supporting files in the `src` subdirectory:
```
src/extensions
src/nbextensions
src/templates
src/ipython_nbconvert_config.py
src/ipython_notebook_config.py
```
* `extension`: Python part of an extensions, e.g. for server extensions or nbconvert pre/post-processors
* `nbextension`: Javascript part of an extension
* `templates`: jinja or html templates for nbconvert or server side extensions

### DIFF
--- a/conda-recipes/nbextensions/install.py
+++ b/conda-recipes/nbextensions/install.py
@@ -1,0 +1,95 @@
+# Install notebook extensions
+from __future__ import print_function
+import os
+import sys
+import shutil
+
+import IPython
+if IPython.__version__[0] < '3':
+    raise ImportError('IPython version 3 required')
+
+from IPython.utils.path import get_ipython_dir
+
+# http://stackoverflow.com/questions/12683834/how-to-copy-directory-recursively-in-python-and-overwrite-all
+def recursive_overwrite(src, dest, ignore=None):
+    if os.path.isdir(src):
+        if not os.path.isdir(dest):
+            os.makedirs(dest)
+        files = os.listdir(src)
+        if ignore is not None:
+            ignored = ignore(src, files)
+        else:
+            ignored = set()
+        for f in files:
+            if f not in ignored:
+                recursive_overwrite(os.path.join(src, f), 
+                                    os.path.join(dest, f), 
+                                    ignore)
+    else:
+        shutil.copyfile(src, dest)
+
+ipythondir = get_ipython_dir()
+
+# Install extensions
+extensions = os.path.join(ipythondir,'extensions') 
+src = os.path.join('src','extensions')
+print("Install extensions to %s" % extensions)
+recursive_overwrite(src, extensions)
+
+# Install templates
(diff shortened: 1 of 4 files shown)


JanSchulz:
This looks so great! Thanks for your work making ipython extensions useable!

As some kind of dreaming: This as a "install http://wherever.com/something.zip" version... In a gui, together with update and removal... with a server where you can upload, search and rate extensions...

JanSchulz:
What I'm missing is some docs: how would I build the package? How do I point users at the package? I use conda but never build my own packages :-/

conda-recipes/nbextensions/install.py
```diff
@@ -0,0 +1,95 @@
...
+    f = open( config_file, 'r')
+    config = f.read()
+    f.close()   
+    
+    # add config
+    f = open( newconfig_file, 'r')
+    new_config = f.read()
+    f.close()
+
+    marker='#--- nbextensions configuration ---'
```
JanSchulz: So this needs to be a unique marker? If so IMO this should be at the top of the file with a "CHANGE THIS" header

conda-recipes/nbextensions/src/ipython_notebook_config.py
```diff
@@ -0,0 +1,15 @@
+#--- nbextensions configuration ---
```
JanSchulz: Same here, if that should be unique

juhasch:
Take a look at https://binstar.org/juhasch/nbextensions
To build your own conda package, you can use the `conda build` command.

For a user, you either provide a local conda repository (basically just a web server) or you can use binstar.org. Both is supported by conda.

conda-recipes/nbextensions/install.py
juhasch: The idea ist to show what has been added to the configuration and to be able to replace that part it if the conda package is updated.
Currently, IPython is not a very friendly place for configuration management.

juhasch: The better solution would be to parse the configuration file and make the modifications there. This is difficult to do, I don't want to parse Python files.

Ideally, the whole notebook configuration will be JSON based, so it will be easy to get and 
change settings.

JanSchulz: The "show" part is fine, but if one is using two such extensions, the second will overwrite the first :-(

juhasch: This receipe is meant to handle the installation of a bundle of extensions, not a single one.
This is a much larger task. Conda is not the right tool for this kind of granularity.

JanSchulz: oups, I think I misunderstood the meaning of "conda recipe", I actually thought that this is an example how to handle individual extensions. Sorry for the noise... :-/

damianavila:
That's nice @juhasch

PR #208, opened 2015-04-10
fix race condition w/ start-time, use correct event for kernel idle, and...

tristanreid:
... fix css loading

- execution end time was reliant on a non-existent kernel event, probably from a previous version
- CSS file now loaded from a more dynamic path
- Execution start time now getting recorded after the cell execution rather than before it (not a huge issue since execution is delegated to kernel, but I believe recording from before makes more sense)

### DIFF
--- a/usability/execute_time/ExecuteTime.js
+++ b/usability/execute_time/ExecuteTime.js
@@ -6,8 +6,12 @@
 //
 // Execute timings - display when a cell has been executed lastly and how long it took
 // A double click on the box makes it disappear
-
-(function (IPython) {
+define([
+    'base/js/namespace',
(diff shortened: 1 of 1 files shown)


juhasch:
Thanks for the PR. Please use the 3.x branch to make your PR, as the master branch is not kept up-to-date, at least as long the big split of IPython/Jupyter is ongoing.

Some of your changes should already be in the 3.x branch.

tristanreid:
Ah!  I see, I had mistakingly thought that master followed 3.x, I didn't see those changes.  Looks like my changes are roughly the same, I'll close this PR.

Thanks!

PR #230, opened 2015-07-12
Extension to skip displaying traceback, only show error message

juhasch:
Sometimes a traceback clutters the noebook output too much. Allow displaying only the error message without the complete traceback.


damianavila:
Looks nice...

juhasch:
Thanks. I often find it annoying looking at a large traceback when I exactly know what's wrong looking only at the error message (like a wrong filename).

damianavila:
>Thanks. I often find it annoying looking at a large traceback when I exactly know what's wrong looking only at the error message (like a wrong filename).

I agree :wink

Issue #236, opened 2015-07-24
nbextensions.py incorrectly terminates jQuery string


brunobeltran:
The symptom is a *completely* blank /nbextensions/ page. 

The reason appears to be that in config/main.js, the line `$('body').data('extension-list')` incorrectly returns a string instead of a list of JSON objects. (Leading to an error on line 35ish when the variable `extensions` is a character from a string instead of a JSON object with member 'Name').

The cause of this appears to be that the data-extension-list attribute of $('body') is incorrectly terminated. In my case it ends up being of the form (note the missing terminating `}]`):
```
[{"Link": "https://github.com/ipython-contrib/IPython-notebook-extensions/wiki/Codefolding", "Name": "Codefolding", "url": "nbextensions/usability/codefolding", "Icon": "icon.png", "Main": "main.js", "Type": "IPython Notebook Extension", "Compatibility": "3.x", "Description": "This extension enables the CodeMirror feature to llow codefolding in code cells"}, ... , {"Link": "https://github.com/jfbercher/IPython-notebook-extensions/tree/master/testing/latex_envs/", "Name": "(some) LaTeX environments for Jupyter", "url": "nbextensions/testing/latex_envs", "Icon": "icon.png", "Main": "latex_envs.js", "Type": "IPython Notebook Extension", "Compatibility": "3.x", "Description": "Enable to use laTeX environments in Jupyter"
```

Adding the terminating `}]` manually and rerunning main.js fixes the problem (as long as I flush jQuery's .data() cache). 

I understand that this data field is populated from inside of config/nbextension.py, but running said file by hand in an iPython shell line by line leaves `json_list` populated correctly (i.e. with the terminal `}]` included). This seems to suggest that the problem is in the function self.render_template or self.write

Unfortunately, I can't get these guys to print to any files by adding debug statements for some reason...

So I'm stuck with a very hack-y workaround of manually modifying the data string from inside main.js on page load. 

In a ?potentially related bug? all div's appear to have `display: none` set on page load. So without my workaround, not even the header with the Jupyter logo is visible. For some reason, I can fix the `display: none` problem by deleting everything in my nbextensions directory except for the files needed for the `/nbextensions/` page to load. I assume that this is because in this case the `extensions_list` is empty, so that main.js gets to the point where it loads the page's CSS or something like that.

juhasch:
Fixed by #237

brunobeltran:
[...]

PR #237, opened 2015-07-24
main.js now allows any HTML5-compliant nbextension name

brunobeltran:
In particular, "(some) LaTeX environments for Jupyter", one of the extensions shipped by default, was causing the `/nbextensions/` page to fail to load since it had parenthesis in the name. 

For a discussion on this problem, see http://stackoverflow.com/questions/17403636/jquery-x-does-not-work-but-document-getelementbyidx-works

It appears that since the only other option would be to manually escape all allowable characters, we should probably go with the solution given by Denys Séguret on that Stack Overflow page.

Tested only in my own branch which uses the workaround described in #236, since otherwise I can't get the page to load at all.

### DIFF
--- a/config/main.js
+++ b/config/main.js
@@ -27,7 +27,7 @@ require([
     var extension_list = $('body').data('extension-list');
 
     var html = "";
-    
+
     for(var i=0; i < extension_list.length; i++) {
         var extension = extension_list[i];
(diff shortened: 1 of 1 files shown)


juhasch:
Thank you. Looks good.

PR #242, opened 2015-07-25
NavPanel added to extensions

BjarkeDalslet:
NavPanel is an extension that:
Add navigation panel to side of Notebook
- Single click on a heading and
the notebook
  focuses on it.
- click and drag to move entire section

and subsections.
- ctrl + click to select entire section and
subsections
  for the chrome-clipboard

### DIFF
--- a/usability/NavPanel/NavPanel.yaml
+++ b/usability/NavPanel/NavPanel.yaml
@@ -1,0 +1,12 @@
+Type: IPython Notebook Extension
+Name: NavPanel
+Description: | 
+Add navigation panel to side of Notebook
+ - Single click on a heading and the notebook
+   focuses on it.
+ - click and drag to move entire section 
+   and subsections.
+ - ctrl + click to select entire section and subsections
+   for the chrome-clipboard
+Main: main.js
+Compatibility: 3.x
--- a/usability/NavPanel/main.css
+++ b/usability/NavPanel/main.css
@@ -1,0 +1,53 @@
+
+.NavPanelCell {
+    -webkit-touch-callout: none;
+    -webkit-user-select: none;
+    -khtml-user-select: none;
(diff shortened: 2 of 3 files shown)


juhasch:
Thanks, looks nice.
A few comments:
* The multiline description in your YAML file breaks the yaml parser in anaconda
* Navigating works well at the start. Using CTRL+click opens a context menu in Chrome here. 
* Selections (green border) don't work reliably for me
* Multiline markdown cells with headers look strange (this is the IPython "Inline Typesetting" example):
navpanel

BjarkeDalslet:
*   Navigating works well at the start. Using CTRL+click opens a context menu in Chrome here.
  * I cannot reproduce this (in Chrome on Windows). It is not the expected Chrome behavior (see e.g. https://support.google.com/chrome/answer/157179?hl=en) . Could it be some conflicting plug-in in your system? 
  
*   Selections (green border) don't work reliably for me
  *  ctrl-click actually "flips" the selection (i.e. selected cells get unselected and vice versa). This means that you for instance can select all cells in a section, except the cells in a single subsection by 
     * ctrl-click the section and then ctrl-click the subsection
     * ctrl-click the subsection and then ctrl-click the section.

  Does this explain the apparent unreliable behavior?
I will update the YAML to reflect this.

*  Multiline markdown cells with headers look strange (this is the IPython "Inline Typesetting" example):
   * I chose to crop the headings exceeding a certain css-definable length as I found that multiline entries into the Navpanel limited its use especially on low-res screens. Is it the cropping that looks strange to you?

jcb91:
OK, I'm closing this as outdated (no action for a year, during which the repository's been refactored a great deal). If & when someone comes back to it, please reopen or submit a new PR as appropriate

PR #245, opened 2015-08-01
Add parameters and improve single quote handling

juhasch:
Add parameters to the extension:
If the YAML description of the notebook contains these entries:
`Parameter:`
`ParameterDescription:`
An input element is added to the extension configuration.

Also, make sure all strings used to compose the json list are checked for single quotes.

Not sure why diff fails with `nbextensions.py`.


juhasch:
Oops. Something went wrong. Closing

PR #255, opened 2015-08-15
Preprocessors4x

juhasch:
- Import `nbconvert`
- Also fold `%` magic
- Make preprocessors less brittle using try/except


juhasch:
Better do it separately.

PR #261, opened 2015-08-18
Make those extensions work with Jupyter version 51fabeba530b1d50fa3255f48e97d24d72e605d7.

Konubinix:
Jupyter requires the modules to be correctly defined and to provide the
load_ipython_extension variable.

To be honest, I don't understand how it works. I just now that it works with a fresh install of Jupyter from the source code.

### DIFF
--- a/publishing/printview/main.js
+++ b/publishing/printview/main.js
@@ -6,34 +6,42 @@ define([
 ], function(IPython, $) {
     "use strict";
     if (IPython.version[0] < 3) {
-        console.log("This extension requires IPython 3.x")
+        console.log("This extension requires at least IPython 3.x")
         return
     }
-    
+
     /**
      * Call nbconvert using the current notebook server profile
      *
      */
 	var nbconvertPrintView = function () {
-		var kernel = IPython.notebook.kernel;
 		var name = IPython.notebook.notebook_name;
-		var command = 'ip=get_ipython(); import os; os.system(\"ipython nbconvert --profile=%s --to html '
+		var command = 'ip=get_ipython(); import os; os.system(\"jupyter nbconvert --profile=%s --to html '
             + name + '\" % ip.profile)';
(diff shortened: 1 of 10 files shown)


juhasch:
Thank you for your effort, this is welcome. Could you split the changes into smaller chunks (i.e. one for each extension).
I find it difficult to look at these changes en block, and I suspect a few of the extension need a little additional work to bring them back to life, too.

Konubinix:
Thank you for looking at it.

To be honest, I don't know what I did since I don't have now time to dig
into Jupyter source code. I just made sure the extensions I wanted
worked well by fixing the errors my Chromium inspector complained
about. To do so, I only mimicked the behaviour of working extensions.

Of course I can provide smaller patches. I will probably do that during
next week.

damianavila:
Sound like a good idea... I can probably provide some feedback when the thing is splitted... thanks @Konubinix

juhasch:
Needs rebasing after big reorganization PR #273

publishing/printview/main.js
```diff
@@ -6,34 +6,42 @@ define([
...
+
     /**
      * Call nbconvert using the current notebook server profile
      *
      */
 	var nbconvertPrintView = function () {
-		var kernel = IPython.notebook.kernel;
 		var name = IPython.notebook.notebook_name;
-		var command = 'ip=get_ipython(); import os; os.system(\"ipython nbconvert --profile=%s --to html '
+		var command = 'ip=get_ipython(); import os; os.system(\"jupyter nbconvert --profile=%s --to html '
```
JanSchulz: AFAIK, jupyter has no concept of "profiles"?!

juhasch: Yes, I already fixed this here but wanted to do the reorganization first.

juhasch:
Closing as outdated

PR #269, opened 2015-08-31
Update reference to js file: ExecuteTime.js

wikiped:
From previously non-existing main.js


juhasch:
Thank you. Included in PR #279

PR #278, opened 2015-09-02
gitignore potential mathjax files

jcb91:
mathjax can be stored in nbextensions for local serving when a network connections is not available to reach the CDN. We don't want to track anything that might be in the mathjax folder.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 custom.js
 custom.css
+mathjax/*


juhasch:
I think this is no longer true for 4.x/Jupyter. Mathjax is now included with the other Javascript components at installation. For 3.x this might be useful.

jcb91:
ah, true! Good to know. In that case, I should remove my local mathjax :)

PR #295, opened 2015-09-10
restructuring: place each extension into its own directory

jcb91:
* remove nbextensions/usability/hide_input.js, which is a duplicate of nbextensions/usability/hide_input/hide_input.js


juhasch:
Much needed. Thanks for taking care of this.

jcb91:
You're welcome :) it was niggling a little.

PR #302, opened 2015-09-15
Feature/tests

juhasch:
Update tests for execute_time, print preview and skip-traceback

### DIFF
--- a/tests/test_execute_time.py
+++ b/tests/test_execute_time.py
@@ -1,0 +1,60 @@
+# -*- coding: utf-8 -*-
+# test ExecuteTime extension
+
+import time
+from selenium import webdriver
+from selenium.webdriver.common.keys import Keys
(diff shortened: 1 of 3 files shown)


jcb91:
FYI, there are some classes which may be of interest for things like this (selenium-based tests which need a running notebook server) in [`tests/nbextensions_test_base.py`](https://github.com/Jupyter-contrib/jupyter_nbextensions_configurator/blob/5d103e839bd41fce3bd8154b72c0cba819260b06/tests/nbextensions_test_base.py) of  `jupyter_nbextensions_configurator`. See [`tests/test_nbextensions_configurator.py`](https://github.com/Jupyter-contrib/jupyter_nbextensions_configurator/blob/5d103e839bd41fce3bd8154b72c0cba819260b06/tests/test_nbextensions_configurator.py) for  example usage.

PR #306, opened 2015-09-22
Save notebook before calling nbconvert

juhasch:
A notebook should always be saved before it is converted using `nbconvert`.


juhasch:
Using something like `--to pdf --template mylatex.tplx` as parameter in printview now allows generating and viewing PDF files from within the browser. After some template massaging, this looks **very** nice.
For my custom template look here:
https://gist.github.com/38c60e098665761a392f.git

Needs https://github.com/jupyter/nbconvert/pull/117

PR #318, opened 2015-09-27
Add table of content notebook plugin

JanSchulz:
The ToC is mostly a direct copy from
https://github.com/minrk/ipython_extensions/ with minor edits
(filenames) and the addition of readme, icon and the yaml activations.

@minrk was fine with copying it here: https://github.com/minrk/ipython_extensions/issues/50

This script has a problem when two more occurences of the same headline
appear in the notebook (e.g. multiple "Plots" headlines in different analyis
steps). The resulting ToC will link to the same anchor and the browser will
always move to the first version of this headline. See
https://github.com/jupyter/notebook/issues/77 for more information on this
problem.


juhasch:
Thanks.

Is there already an issue on the multiple links issue in Jupyter/notebook ?

JanSchulz:
@juhasch Yes, the problem about multiple headlines with the same content resulting in the same ID/link is in jupyter/notebook#77

PR #323, opened 2015-10-01
Call nbconvert in notebook directory

juhasch:
When calling `jupyter nbconvert`, be in the current notebook directory


juhasch:
Not the right approach, doesn't really help.

PR #327, opened 2015-10-03
Update TOC on markdown render

juhasch:
When rendering a markdown cell, update TOC. Eliminates need to manually update TOC.


jfbercher:
Just tested this. The update works and is a nice idea. thx

juhasch:
Thanks for the feedback.

PR #335, opened 2015-10-04
Equation numbering extension

juhasch:
* Enables automatic equation numbering
* Resets equation numbers and rerenders math on pressing toolbar button


certik:
Thanks for making this extension. I think this is an important step towards making this work automatically in the notebook by default.

ketch:
Awesome!

lucasb-eyer:
Thank you for this!

maccesch:
Nice!

neeravbm:
@juhasch Equation numbering plugin is converting "x \\\\ y" to "x \ y" within the equation when downloaded as HTML file. Similarly it is converting f_x to f\<em\>x. I am not familiar with Jupyter but know HTML/CSS/JS. Can you let me know where should I fix the issue in code? I can then submit the changes as a pull request.

jfbercher:
I do not see how this could come from @juhasch equation-numbering. It may come from `latex_envs` if you use it (which also provides automatic equation numbering). In such a case, please try to isolate which plugin is involved. If it were `latex_envs`,  try to update to a recent version at [Pypi](https://pypi.python.org/pypi/jupyter_latex_envs) or to the most recent version on [github](https://github.com/jfbercher/jupyter_latex_envs). If the problem persists, please post an example notebook somewhere plus the commands you use to convert to html.

PR #339, opened 2015-10-06
add small extension to provide a button to insert current time & date

jcb91:


JanSchulz:
Whats the usecase here? There is a IPython extension which prints out date/time + some version information for specific packages and so on. Not sure if there is some overlap here...

jcb91:
Well, I wrote it because I was using notebooks to make a kind of lab logbook, and got tired of constantly typing time/date into markdown cells when I wanted to add a new note. In such cases, I wanted the date to be unrelated to the (re)-execution time of any code cell. Perhaps it would be worth mentioning in the description.

JanSchulz:
My problem with this is that it's a very small function, but a comparatively large space (icon).

This sounds like a good case for some kind of text expander (type "TS", get it expanded to "20151006-2200" or whatever you prefer... There are several text expander for all kind of OS, not sure how difficult it would be to implement this kind for the jupyter notebook.

jcb91:
well sure, if you feel it clutters your toolbar, you can always not use it, no?

juhasch:
I am fine with merging the extension. 

There is one thing I begin to worry about: We start to have a lot of extensions and the configuration page gets quite full. I think we should discuss how to make it easier to navigate.

jcb91:
Yes, I agree, it's getting quite long. Some kind of accordion-like structure might be appropriate, perhaps? I'll have a think.

PR #340, opened 2015-10-06
remove usability/help_panel…

jcb91:
…since it is now provided by the notebook's `help -> keyboard shortcuts` menu item.


JanSchulz:
Was this printable (I don't have this extension in /nbextension) ? If yes, that's one of the benefits over the internal one (see here: https://github.com/ipython/ipython/issues/8141)

jcb91:
ah, ok, good point!

I'll update it instead, then.

J

JanSchulz:
@jcb91 I've actually no idea if this extension *does* that... IMO this extension is not so nice to maintain, it would be better to get the normal keyboard shortcuts form the same source as the normal `help -> kb shortcuts`. Or extend that list with our own and make it printable :-)

jcb91:
oh yes, absolutely, use the builtin quickhelp to get the shortcuts, then present them in another way.
What exactly do you mean by printable? Would they need to be in a separate tab/window as suggested in ipython/ipython#8141?  This particular extension appears to have been a prototype for the jupyter quickhelp, but appearing at the right side of the screen rather than in a modal. As such, it might (?) be better for printing, but I'm not really sure...

jcb91:
getting all the shortcuts from jupyter's quick_help is not currently possible, simply because they aren't exported. I could try an intermediate hack by getting quickhelp to construct the modal dialog as usual, then hijacking it via jquery...

JanSchulz:
I like to have my keyboard shortcuts printed and taped over the screen :-) But trying to print the keyboard shortcuts will print a very ugly page with all the background noise...

jcb91:
ok, so, I've updated usability/help_panel, and will resubmit as a new PR...

PR #341, opened 2015-10-06
updates to config readme:

jcb91:
* switch back to a link to the repository's main readme rather than a relative link, in order to avoid breakage when viewed in rendermd
* clarify some of the manual installation instructions about where to find files
* remove references to editing json config files, in favour of editing the `jupyter_notebook_config.py` and `jupyter_nbconvert_config.py` files
* add github-flavour language specs for code blocks

### DIFF
--- a/nbextensions/config/readme.md
+++ b/nbextensions/config/readme.md
@@ -26,161 +26,164 @@ You can see a video of the config extension in action on youtube:
 
 <iframe width="420" height="315" src="https://www.youtube.com/embed/h9DEfxZSz2M" frameborder="0" allowfullscreen></iframe>
 
+
 Setup procedure
 ===============
 
 The most recent version
-can be found here:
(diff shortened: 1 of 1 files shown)


juhasch:
Thanks. I think everything is fine, except for the configuration files part.

`.py` files are better for human editing and keeping the configuration in one place is also good. The reason why I moved a part of the configuration to JSON is that IPython allows merging different configurations in JSON, and it is hard to parse Python configuration files programmatically. The reason why there is still a `py` configuration file is because it is needed to add the `extensions` directory to the Python path. The template settings should actually go to JSON, too. I am happy to discuss other options.

Also, Jupyter seems to move more into the direction of JSON configuration ?

jcb91:
Ah, yes, I missed some of the subtleties! I also hadn't thought of upgrading.
Ok, I think I'd agree it makes sense to move as much as possible into the JSON configuration files. But, I guess, the python path still needs modifying through the python config? I'll try to get time to make some edits this evening

PR #342, opened 2015-10-06
Config/js various updates to config

jcb91:
* add support for editable hotkeys as extension parameters
* make activation/deactivtion more robust (was breaking for extensions containing `-on` in the name, like `read-only`
* fix `nbextensions.html` template to use the new rendermd url
* improve rendering of markdown readmes to better match notebook markdown cells
* simplify config CSS by making more use of bootstrap classes
* split rendermd CSS overrides from main config CSS
* fix padding, add comments in rendermd CSS

### DIFF
--- a/nbextensions/config/hotkey_editor.js
+++ b/nbextensions/config/hotkey_editor.js
@@ -1,0 +1,236 @@
+
+define([
+    'underscore',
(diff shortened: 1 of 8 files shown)


juhasch:
Looks good.

PR #344, opened 2015-10-07
Created a yaml file for this extension.

idichekop:
This yaml file allows this extension to appear in the nbextension configuration webpage.  
Users can now activate/deactivate this extension easily.
Tested in Jupyter 4.0, and works well. No shortkey association yet, but will be neat in the future.


jcb91:
@idichekop I think the current implementation of the `hide_input_all` extension is incompatible with Jupyter 4.x, since it doesn't define a module object with a `load_ipython_extension` function. This is why it doesn't currently have a yaml file. I may be wrong though - did you try actually using the extension in a notebook, as well as the activation/deactivation?

I've actually updated this to 4.x already anyway, but haven't yet submitted a PR. I'll post back once I've done so.

jcb91:
See #356

juhasch:
Fixed in PR #356. Thanks for the input.

PR #347, opened 2015-10-08
update `usability/help_panel` to 4.x, adding features:

jcb91:
* get shortcuts from the `quickhelp` module, rather than creating them from scratch
* make side panel resizable
* make panel expandable to full screen
* add readme, similar to wiki page
* update screenshots, icon


JanSchulz:
Judging from the screenshot, this looks great.

JanSchulz:
Unfortunately, after activation, the button does not show up. Not sure whats wrong here :-(

jcb91:
Hmm, strange. Anything in the browser javascript console?

JanSchulz:
This is in the JS console:
```
Default extension for cell metadata editing loaded.
main.min.js:23979 Raw Cell Format toolbar preset loaded.
main.min.js:24027 Slideshow extension for metadata editing loaded.
main.min.js:12030 Loaded extension: usability/chrome-clipboard/main
main.min.js:12030 Loaded extension: usability/python-markdown/main
main.min.js:12030 Loaded extension: usability/toc/main
main.min.js:12030 Loaded extension: usability/limit_output/main
main.min.js:12030 Loaded extension: config/config_menu/main
main.min.js:23181 Session: kernel_created (9b9e21a9-a978-4f24-8643-4588fdfa0226)
main.min.js:22502 Starting WebSockets: ws://localhost:8888/api/kernels/f51e2e26-51de-4fdc-99b8-db39128ed085
http://localhost:8888/static/components/bootstrap/js/tooltip.js?v=20151008124651 Failed to load resource: the server responded with a status of 404 (Not Found)
main.min.js:12049 Failed to load extension: Array[1] Error: Script error for: components/bootstrap/js/tooltip(…)(anonymous function) @ main.min.js:12049
main.min.js:22160 Kernel: kernel_connected (f51e2e26-51de-4fdc-99b8-db39128ed085)
main.min.js:12030 Loaded extension: usability/ruler/main
main.min.js:22160 Kernel: kernel_ready (f51e2e26-51de-4fdc-99b8-db39128ed085)
:8888/nbextensions/usability/ruler/main.js?v=20151008124651:32 ruler_color: #808080
:8888/nbextensions/usability/ruler/main.js?v=20151008124651:39 ruler_column: 78
:8888/nbextensions/usability/ruler/main.js?v=20151008124651:44 ruler_linestyle: dashed
main.min.js:12030 Loaded extension: widgets/notebook/js/extension
:8888/nbextensions/widgets/notebook/js/extension.js?v=20151008124651:81 loaded widgets
```

jcb91:
Ah, OK, it looks like the bootstrap tooltip plugin may not always be
included, then. I'll add some error handling to fix this in a little
while...

jcb91:
@JanSchulz ok, I've pushed a commit which (touch wood) should fix this. Please let me know whether ti works for you...

JanSchulz:
Yep, it works, but I chromes print preview (ctrl+P) and afterwards printing does not work in "full text mode": I get "Druckvorschau fehlgeschlagen" (printpreview doesn't work) an dthe printing butotn is greyed out :-( 

In split screen mode, the keybord shortcuts are printed next to the code, each using up half on the page :-(

Is there a way to generate the page content and open it in a new window (without saving the content to disk)?

jcb91:
Right! Well, progress, I guess.

You should be able to drag the sidebar to take up 90% of the page width, although this isn't a perfect solution, obviously. I get the same problem with Chrome's print-preview failing, though I'm not sure why. I'll look into it if I get time. 

As for producing the content in another page... I think this *might* be possible in pure javascript, rather than a server-side extension, but I'm not sure. generating it from a server-side extension wouldn't be a great plan anyway, really, since it would make loading all the correct extensions and things complicated (since they can define keyboard shortcuts, it affects the help panel). However, since it's not clear what's causing the print-preview problem in the first place, I'd be reluctant to go rushing off trying to implement a javascript solution without knowing yet whether it'll actually help...

jcb91:
@JanSchulz fixed, I think. It turned out to be an issue with Chrome's print previewer being unable to deal with 100% height elements, so just hitting print might have worked anyway. I also fixed an issue where you'd only get a single column of keys when printed, rather than the two you get on screen (since the previewer considers the page only ~500px wide apparently). As before, please let me know whether or not this works for you...

juhasch:
This is much better than before!
Ready to merge, when you are.

JanSchulz:
Wow, great work. Printing works great! You might wanna upstream this...

keyboard_shortcuts

Thanks a lot!

@jcb91 I owe you, that was usually half an hour c&p and formatting work in word to get a new shortcut tipsheet... Ping me if you want something done in python...

jcb91:
:)

PR #351, opened 2015-10-08
update `publishing/print_preview_toolbar_button` to 4.x

jcb91:


juhasch:
Not sure if anybody uses this at all ?

jcb91:
Nor me!

juhasch:
There is a menu for exactly this, so let's remove this extension.

jcb91:
Fair enough, I'll edit.

jcb91:
Although, this seems to have been written when the menu item already existed, so I guess someone just wanted a button. It might be better though, to have an extension which would just provide toolbar buttons for arbitrary actions.

PR #354, opened 2015-10-08
update `usability/clean_start` to 4.x, also making it an action

jcb91:


JanSchulz:
At least in my version, the "restart" action (both menu and button) include a "clear all and restart" option. So IMO, this extensions can be removed...

jcb91:
True, mine too. I didn't think to check! I'll close this and open another PR to remove it.

PR #355, opened 2015-10-08
update `usability/hide_input` to 4.x

jcb91:
* remove usability/hide_input/hide_input.tpl, which now exists in `templates/`
* add icon & yam file


nbextensions/usability/hide_input/main.js
```diff
@@ -1,51 +1,44 @@
...
         // Toggle visibility of the input div
-        cell.element.find("div.input").toggle('slow')
-
-        if ( cell.metadata.input_collapsed ) {
-            cell.metadata.input_collapsed = false;
-        } else {
-            cell.metadata.input_collapsed = true;
-        }
+        cell.element.find("div.input").toggle('slow');
+        cell.metadata.input_collapsed = ! cell.metadata.input_collapsed;
```
juhasch: I think we should use the same metadata structure `cell.metadata.hide_input` as in the other extensions (hide_input_all, runtools)

jcb91: OK, sounds reasonable, I'll fix it when I get a chance later.

jcb91:
ok, done

PR #363, opened 2015-10-10
"Highlighter" extension

jfbercher:
## Abstract - 
* The proposed extension enables to highlight select text in a markdown cell, both in edit and command (rendered) mode. The highlighting scheme is chosen via several push buttons. In the case no text is selected, the whole cell is highlighted. Another button enables to remove all highlightings in the current cell. 
* A second commit add a small modification to the "latex_envs" extension in order to enable these highlightings in LaTeX environments. 

I hope it can be useful. 

---- 
Below is the readme.md :

## The highlighter extension:

- Firstable, the extension provides <span class="mark">several toolbar buttons</span> for highlighting a selected text _within a markdown cell_. Three different \`color schemes' are provided, which can be easily customized in the stylesheet `highlighter.css`. The last button enables to remove all highlightings in the current cell. 
- This works both <span class="burk">when the cell is _rendered_ and when the cell is in edit mode</span>; 
- In both modes, it is possible to highlight formatted portions of text (In rendered mode, since the selected text loose its formatting, an heuristic is applied to find the best alignment with the actual text)
- When no text is selected, the whole cell is highlighted; 
- The extension also provides two keyboard shortcuts (Alt-G and Alt-H) which fire the highlighting of the selected text. 

## Installation:

The extension can be installed with the nice UI available on IPython-notebook-extensions website, which also allows to enable/disable the extension. 

You may also install the extension from the original repo: issue
```bash
jupyter nbextension install https://rawgit.com/jfbercher/small_nbextensions/master/highlighter.zip  --user

```
at the command line.

### Testing: 

Use a code cell with
```javascript
%%javascript
require("base/js/utils").load_extensions("usability/highlighter/highlighter")
```

### Automatic load
You may also automatically load the extension for any notebook via
```bash
jupyter nbextension enable usability/highlighter/highlighter	
```


juhasch:
Very nice!

Have you thought about how highlighted text/cells can be exported ?

damianavila:
Really nice, I completely agree...

jfbercher:
Thank you very much! 

@juhasch Indeed, it will be useful to save highlihghtings on export. 
I haven't seriously thought to this because I was in a hurry for finishing this extension (my "use case" was to provide to my students a way to mark the text of notebooks during class). 
For the html export it is very easy since the highlight markup is preserved and it suffices to add a template that loads the css.  For other formats, it will be a bit more involved.

Issue #368, opened 2015-10-15
Incomplete nbextensions displayed for /nbextensions


achampion:
/nbextensions only displays a subset of the extensions in ~/Library/jupyter/nbextensions folder
The jupyter console shows discovery of many more extensions but only shows the up to Codefolding and then abruptly stops.
Tried different browsers (Safari, Chrome, Chrome Canary) all with the same result.

The console showed an error reading limit-output.yaml (couldn't read "default: **OUTPUT MUTED**" - removing the '*' fixed this error but still not displaying anything beyond Codefolding.

jcb91:
@achampion OK, sounds strange. Could you paste the full javascript console output? Whereabouts is/are the `*` you refer to? do you mean line 18 of `limit_output.yaml`? That should read
```yaml
  default: **OUTPUT MUTED**
```

jcb91:
Apologies, I think I got confused about which console you were referring to. When you said
> The console showed an error reading limit-output.yaml (couldn't read "default: OUTPUT MUTED"
I though you meant the browser javascript console, but I realise now you must mean the jupyter notebook shell console, since the client-side javascript doesn't know anything about the existence of yaml files in the first place. I suppose this might be something to do with the python  `yaml` module version, if asterisks seem like a problem? You could try quoting the offending text to see if that helps, so that it reads
```yaml
  default: "**OUTPUT MUTED**"
```
But, this seems like a separate problem to your partly-filled config page. Javascript console output would be helpful for that.

achampion:
Yes, sorry the jupyter shell console. Quoting does fix the problem reading that .yaml file. Here's the out put from the console (I've * the ones that show up in the /nbextensions page):

```
[I 09:13:35.201 NotebookApp] Found 4.x extension NbExtensions menu item
[I 09:13:35.205 NotebookApp] Found 4.x extension Printview
[I 09:13:35.208 NotebookApp] Found 4.x extension zenmode
*[I 09:13:35.213 NotebookApp] Found 3.x, 4.x extension AutoSaveTime
*[I 09:13:35.219 NotebookApp] Found 3.x, 4.x extension Autoscroll
*[I 09:13:35.221 NotebookApp] Found 4.x extension Chrome Clipboard
*[I 09:13:35.224 NotebookApp] Found 4.x extension Codefolding
[I 09:13:35.227 NotebookApp] Found 3.x, 4.x extension Comment/Uncomment Hotkey
[...]

PR #369, opened 2015-10-15
fix YAML file for `usability/limit-output`…

jcb91:
… since starting a YAML scalar with a special character ('*') can cause some versions of the python `yaml` module to baulk (although others seem ok with it, as long as there are two, so I didn't notice this on my version). See #368 for details.


juhasch:
LGTM

PR #370, opened 2015-10-15
fix bug in `config/main`…

jcb91:
…fixed a typo on line 520 which was causing a single failed extension-load to break the rest of the page load.


jcb91:
see #368 for details

PR #379, opened 2015-10-27
Config/js

jcb91:
Alter layout of config page to show the description and parameters for only a single extension at a time. This layout is hopefully a bit more mobile-friendly too.

image


juhasch:
I like that. I wonder if we should show the readme.md directly, if there is only one extension shown at a time now.

jfbercher:
Nice work! Showing the readme would be a plus.

jcb91:
Yeah, in-page readmes would be nice. It took me a while to figure out that you meant each extension's readme, and not the config page readme (doh!). Anyway, I've been thinking about it, but it's not quite as simple to implement as I'd thought it might be at first glance.

Rendering the markdown from the readme files and inserting it into the DOM doesn't work brilliantly, because a lot of the markdown files contain relative links to images, other readme files, etc. I feel like attempting to edit each link in the resulting markup to point to the correct place would be complicated, and prone to bugs. The simpler solution would be to use an iframe to load the existing rendermd page (which handles relative links just fine). However, the iframe gets its own scrollbar, and on my setup at least, ends up with a pretty tiny height:

image

I'll try to implement some signalling to get them to resize the iframe correctly.

juhasch:
I think we can merge it like it is now and you can create a new PR if there is a good way to render the reademe.
We probably need to read the readme file and render it in a container on the page, like render.js already does.

PR #384, opened 2015-10-29
Added move_selected_cells extension

jfbercher:
I found myself this afternoon facing the problem of moving many cells from a part of a notebook to another part of the same notebook. Since I am still in vacations :) and since I learned previously about multicells selection, I got the idea of this quick extension. Clearly, this kind of thing will be probably integrated shortly in the mainstream. But in the meantime, it could be useful to some people.

## Move selected cells

This is a quick (and dirty) extension - **move up or down several selected cells**

It is a bit dirty because it would be better to act on DOM elements and write a correct move_cells() function. In the meantime, moving cells or series of cells via simple keystrokes is super useful

Keyboard shortcuts: *Alt-up* and *Alt-down* (works also with single cells!)

**Cells selection**: Cells can be selected using the `rubberband` (which is a requirement here) or via Shift-J and Shift-K

### DIFF
--- a/nbextensions/usability/move_selected_cells/move_selected_cells.yaml
+++ b/nbextensions/usability/move_selected_cells/move_selected_cells.yaml
@@ -1,0 +1,7 @@
+Type: IPython Notebook Extension
+Name: Move selected cells
+Description: Move selected cell*s* using keybaord shortcuts Alt-up and Alt-down
+Link: https://github.com/ipython-contrib/IPython-notebook-extensions/wiki/
+Icon: 
+Main: main.js
+Compatibility: 4.x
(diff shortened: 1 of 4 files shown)


nbextensions/usability/move_selected_cells/move_selected_cells.yaml
```diff
@@ -0,0 +1,7 @@
+Type: IPython Notebook Extension
+Name: Move selected cells
+Description: Move selected cell*s* using keybaord shortcuts Alt-up and Alt-down
+Link: https://github.com/ipython-contrib/IPython-notebook-extensions/wiki/
```
juhasch: You can link to the README.md locally.

jfbercher: Done. 
Also added a cautionary note to say that such a functionality will probably available shortly directly in Jupyter.

PR #386, opened 2015-10-31
Extensions: code font size, qt console, SKILL language

benvarkey:
*Added an extension to increase/decrease code's font size
*Added an extension to launch the qt console as described in the example custom.js shipped with jupyter
*Added an extension to add syntax support for "SKILL", which is a language used in EDA (I don't think this extension is supposed to be here, but I am not sure where to dump it).


juhasch:
Hi, thanks. I know what Skill is (and I am happy not having use it ...).
I have to small comments, otherwise it looks goot to me.

nbextensions/codemirrormode/skill/skill.yaml
```diff
@@ -0,0 +1,5 @@
+Type: IPython Notebook Extension
+Name: SKILL Syntax
+Description: Enable SKILL syntax support for CodeMirror
+Main: skill.js
+Compatibility: 4.x
```
juhasch: Can you add a link to the readme.md ?
`Link: README.md`

nbextensions/codemirrormode/skill/README.md
```diff
@@ -0,0 +1,4 @@
+# CodeMirrorSkill
+
+This provides a *SKILL* mode for CodeMirror editor.
+The `skill` module adds a MIME type `x-skill` and a mode `skill` that can then be used with CodeMirror.
```
juhasch: Can you add a description/link of what skill is, so people don't have to google for it
(and if you search for skill alone you will never find the right answer...)

benvarkey:
I have updated the .yaml and the README.md

Once you use SKILL regularly, it isn't that bad; of course, a Python interface would be much easier :-). The default SKILL interpreter console doesn't hold a candle against IPython, so I ended up writing a Jupyter kernel for the SKILL interpreter as well (https://github.com/benvarkey/JuVi).

PR #387, opened 2015-10-31
Fix pyyaml requirement

jcb91:
The correct python package name is `pyyaml`. Anaconda also provides a `yaml` package, but that seems to provide Linux/Mac-only binaries, rather than python bindings.


juhasch:
Thanks for looking into it.

Issue #388, opened 2015-11-01
Error loading extensions with shortcuts that have default value


benvarkey:
I am on a Mac, using Safari. When using extensions that have keyboard shortcuts with default values (e.g. `toggle_all_line_numbers`), I get an error loading the extension. 
I chased it down to 
 **nbextensions/config/hotkey_editor.js**: **humanize_shortcut**: 
`_.map(shortcut.split('-'), this.humanize_key ).join(joinchar);`

FYI, replacing that line with `shortcut.split('-').join(joinchar);` fixes the problem, but I don't know what `_.map` was used for.

jcb91:
OK, thanks for the report. I think the `_.map` call is simply to apply
`humanize_shortcut` to each key in the combination, which should do some
replacements like '-' for 'minus', and on mac, things like 'alt' for the
appropriate unicode key symbol if I recall correctly. Presumably I've made
a boo-boo somewhere there, so I'll take a look today.

jcb91:
OK, so I've discovered/thought about a few things.

First, some of the extensions use the actions framework to register their keyboard shortcuts. The fact that the notebook recently changed from using a period to using a colon as the separator between the action prefix and action name mean that some of the extensions end up throwing errors while attempting to add button groups, if you're on a new enough version of the notebook (anything after   7df94eb0013f9990b85c380e453728b8c18a38ca). This is enough to stop the relevant extensions from loading correctly. The extensions in question are autoscroll, comment-uncomment, init_cell, read-only, and toggle_all_line_numbers. 

Second, I'm now a little confused by what exactly you're reporting, so I have a few question:

* What version of jupyter are you using?

* From the issue title

  > Error loading extensions with shortcuts that have default value

  I assume that you mean that if you set a specific value, the extension works correctly, but if you try to keep the default, it doesn't. Is that correct?

* When you say 

  > I get an error loading the extension

  I assume you mean you've seen an error message of some kind in the browser javascript console. Is that right? Could you post (all of) the console output from loading a regular notebook?

[...]

PR #389, opened 2015-11-02
bug fix in `nbextensions/config/hotkey_editor.js`…

jcb91:
… provide missing `humanize_key` method for use in quickhelp patch methods. This should have been included from the beginning, I just didn't do enough testing :(


jcb91:
addresses #388

jcb91:
there we go! That should work (touch wood!)

PR #390, opened 2015-11-02
future-proofing for extensions using the actions framework

jcb91:
The jupyter actions framework altered the character used to join the prefix to the action name from a period to a colon somewhere between 4.0.5 and 4.0.6, in commit 7df94eb0013f9990b85c380e453728b8c18a38ca.
This commit replaces code which was using hard-coded periods to construct full action names for things like buttons with code which uses the full action name which is returned by the action registration call.

### DIFF
--- a/nbextensions/usability/autoscroll/main.js
+++ b/nbextensions/usability/autoscroll/main.js
@@ -14,6 +14,7 @@ define([
     "use strict";
 
     var prev_threshold = 0;
+    var action_full_name; // will be set when registering the action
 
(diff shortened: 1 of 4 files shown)


juhasch:
Is this also needed in more places ?

jcb91:
I think I got all the ones I could find, I'll check again now.

jcb91:
yep, a quick grep for `\.actions` shows only the four already addressed. I think it was only ones I'd 'fixed' in the first place (embarrassed face)

PR #392, opened 2015-11-03
Add example for conda in README.md

juhasch:
Add example from nehalecky to README.md on how to create and use conda receipe

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,17 @@ Conda recipe
 -------------
 
 The conda recipe can be found in `meta.yaml`. It will build a conda package using the most recent master branch.
+
+### Example:
+```
+$ source activate root  #activate root conda env
+$ conda install anaconda-client conda-build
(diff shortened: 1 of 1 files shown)


JanSchulz:
BTW. should I build this Repo in my package builder? https://github.com/JanSchulz/package-builder -> conda packages end up here: https://anaconda.org/janschulz/ (win, linux, osx; 2.7, 3.4, 3.5)

juhasch:
I think this would be a good idea!
We should probably move this repo to jupyter-contrib first, to avoid even more confusion.

damianavila:
LGTM

JanSchulz:
@juhasch @damianavila Adding it to the package builder is no problem but it would be nice if the package would also get a tag for a version (aka a release :-))

damianavila:
>  the package would also get a tag for a version (aka a release :-))

Yep... I defer the decision about when to make a release to @juhasch because he knows better the status quo of the repo...

PR #394, opened 2015-11-05
Fix TOC for headlines with span

JanSchulz:
Closes: https://github.com/ipython-contrib/IPython-notebook-extensions/issues/380

### DIFF
--- a/nbextensions/usability/toc/main.js
+++ b/nbextensions/usability/toc/main.js
@@ -6,9 +6,9 @@ define(["require", "jquery", "base/js/namespace"], function (require, $, IPython
   var make_link = function (h) {
     var a = $("<a/>");
     a.attr("href", '#' + h.attr('id'));
-    // get the text *excluding* the link text, whatever it may be
     var hclone = h.clone();
-    hclone.children().remove();
+    // last() is the |P anchor which lights up when you hoover over the headline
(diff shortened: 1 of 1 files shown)


nbextensions/usability/toc/main.js
```diff
@@ -6,9 +6,9 @@ define(["require", "jquery", "base/js/namespace"], function (require, $, IPython
   var make_link = function (h) {
     var a = $("<a/>");
     a.attr("href", '#' + h.attr('id'));
-    // get the text *excluding* the link text, whatever it may be
     var hclone = h.clone();
-    hclone.children().remove();
+    // last() is the |P anchor which lights up when you hoover over the headline
+    hclone.children().last().remove();
     a.text(hclone.text());
```
jfbercher: Hi Jan, I think that there is perhaps a little better to do: try to replace the two .text() line 12 by .html(). This seems to enable to also render formatting in the link.

JanSchulz: Hm, at least `# Test $1=2$` is horrible with the current system... lets try your approach...

JanSchulz: Ok, text is much more readable...

juhasch:
Looks good for me.

PR #397, opened 2015-11-06
Write Python variables to metadata

bbirand:
I found that otherwise, I couldn't properly read these values in the Latex template file `printviewlatex.tplx`.

### DIFF
--- a/nbextensions/usability/hide_input_all/main.js
+++ b/nbextensions/usability/hide_input_all/main.js
@@ -10,8 +10,13 @@ define([
     "use strict";
 
     function set_input_visible(show) {
-        IPython.notebook.metadata.hide_input = show;
-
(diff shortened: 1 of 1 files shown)


nbextensions/usability/hide_input_all/main.js
```diff
@@ -10,8 +10,13 @@ define([
     "use strict";
 
     function set_input_visible(show) {
-        IPython.notebook.metadata.hide_input = show;
```
juhasch: I think simply ` IPython.notebook.metadata.hide_input = !show;`should be enough.

bbirand: Makes sense!

juhasch:
Should be fixed by #414

PR #407, opened 2015-11-11
Config/main

jcb91:
As first discussed in #379, This PR updates the config page to get extensions' markdown readme files to render on the config page directly, rather than on a separate page. This is accomplished by patching the marked.js renderer to create images and links with absolute URLs as their src/href attributes from relative URLs in the markdown (otherwise, the markdown URL is relative to the markdown file, but it's displayed on the config page, so the relative link points to the wrong place).

There are also a couple of bug fixes, and some attempts at UI improvements/fixes for annoyances.

### DIFF
--- a/nbextensions/config/main.css
+++ b/nbextensions/config/main.css
@@ -15,9 +15,10 @@
 }
 
 .nbext-showhide-incompat {
-    margin: 1em 0.5em 0.5em;
-    float: left;
-    clear: both;
+    padding-bottom: 0.5em;
+    /* the selector is hidden by default, revealed by js if any incompatible
+       extensions are found */
+    display: none;
 }
 
 /* styles for the extension-selector nav links */
--- a/nbextensions/config/main.css
+++ b/nbextensions/config/main.css
@@ -61,6 +61,35 @@
     content: "\f06a  (incompatible, enabled)";
 }
 
+/* styles for the readme */
+
+.nbext-readme > .nbext-readme-contents {
+    /* padding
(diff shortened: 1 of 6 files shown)


juhasch:
Nice!
Two things I noticed:
* http://127.0.0.1:8888/nbextensions/config/readme.md is displayed as raw text now
* Selecting extensions with a longer readme.md lead to scrolling down of a few lines

jcb91:
Ah, right.
So, if I remember correctly, `nbextensions/config/readme.md` always gave raw text, and in order to get the rendered version, you'd need the url `nbextensions/config/rendermd/nbextensions/config/readme.md`. This still works, as visible through the title-bar link (which I'd accidentally hidden, d'oh!). I've just fixed the hidden page title.

jcb91:
As for the scrolling, that was deliberate, but I see that it can be annoying. The idea was for use on mobile screens where the settings can be off the bottom of the page, to make it more obvious that they're there below the list of extensions. I'll try to make a fix so that the scrolling only happens if the extension list goes off the bottom of the page...

jcb91:
So, for some reason, I'd decided that it made sense to scroll so that at least a quarter of the screen was taken up by the extension settings, which looks weird if they're already on screen. I've altered the logic such that it now only scrolls if the settings are not already visible, and then only so that the selected extension title is visible. However, if you think it'd be preferable not to scroll at all, it's easy to just disable the scrolling completely.

juhasch:
I think it is fine this way. Looks good to me.

PR #409, opened 2015-11-12
The templates path is now set in the JSON config file

juhasch:
Remove setting the template path a second time from the PY config files


juhasch:
Closing as outdated

PR #415, opened 2015-11-14
Fix detecting a running notebook server

juhasch:
Seems detecting a running notebook server has changed.
This works for me, however not sure why things have changed.


jfbercher:
Does not work for me unfortunately [linux debian 9]. The process name seems to be "jupyter-noteboo"
image
Perhaps this could work?
image

juhasch:
Yes, this is better. Thanks.

juhasch:
Will be superceeded by changes in #417

PR #416, opened 2015-11-14
[Python-markdown] Fix error when the returned content includes newlines

HenriJ:
## How to reproduce:

In a Python cell:
```python
import numpy
```

And in the markdown cell:
```markdown
This will trigger an error: {{numpy.matrix([[0], [1]])}}
```
## Fix

In this case, python will return a string containing an unescaped newline
The current regex will not be able to match and the code is going to throw
In order to fix this bug, we simply replace the dot in the regex by [\s\S] that will also match new lines


juhasch:
Looks good. Thank you.

PR #417, opened 2015-11-15
New conda recipe

juhasch:
New recipe that will:
* Install the notebook extensions under `$PREFIX/share/jupyter`, allowing to install nbextensions in different environments
* configures the notebook extensions in a separate post-link phase, calling `configure-nbextensions.py`
* removes `setup.py` to allow a setuptools based approach.
* can be uninstalled by `conda remove nbextensions`

TODO:
* move configuration files to `$PREFIX/etc/jupyter` to make them environment-aware.
* change package source in `meta.yaml` from local to a GitHub hosted tagged version

To create and install a conda package:
```
cd IPython-notebook-extensions
conda build conda-recipe
conda install --use-local nbextensions
```


janschulz:
this repo might help: https://github.com/Anaconda-Platform/nb_conda

juhasch:
Did you try it ?
This seems to require another extension `nbsetuptools` which adds another set of command line tools to install extensions.

janschulz:
@juhasch Nope, only found these repos which seem to be intended to be installed by default in the anaconda install (https://github.com/jupyter/jupyter/issues/124)

damianavila:
We don't use nbsetuptools anymore... this was something we wrote to make the nbextension install/enable into the prefix... but now this is offered in notebook 4.2.

You can see an update nb_conda compatible with 4.2 here: https://github.com/Anaconda-Platform/nb_conda/tree/master/conda.recipe

RinaldoB:
I used a few nbextensions with an earlier ipython notebook version. I am now migrating to jupyter notebook and I would like to keep using or try new extension from this repo. however manually installimg the extension via setup.py (or pip) is to much of a hurdle for me. 

I would like to suggest that you not only provide a conda receipe but also a convenient way to install/uninstall notebook extensions within the conda ecosystem. [conda-forge](https://github.com/conda-forge) might be worth looking into.

jcb91:
@RinaldoB

> manually installimg the extension via setup.py (or pip) is to much of a hurdle for me.

what is your expected install process?

The current problem we have with making a conda-recipe (as far as I can tell) is that our current install script blindly copies files into the user's jupyter data/config directories, whereas conda requires files in sys.prefix to be written in order to be able to keep track of what's been changed. This would be fixed by e.g. #626 or #634, at which point we can add a conda recipe again.

RinaldoB:
@jcb91 

suggestion for an expected install process:

`conda install nbextensions`

> After installation, simply go to the /nbextensions/ page in the notebook to activate/deactivate your notebook extensions.

_Buildung_ and _Installing_ should be separated. Ideally the `nbextensions` package would be build by the maintainers of the package and submitted to a package repository. [conda-forge](https://github.com/conda-forge) might be an option for that. Or maybe it could be directly included  in the Anaconda distribution:

> Package creators may also contact Continuum Analytics to submit their package for consideration to be included in Anaconda.

https://docs.continuum.io/anaconda/faq, see also: https://github.com/jupyter/jupyter/issues/124

As a sidenote, the _building_ is really only meant to _build_ a conda package. The _building_ process  should not modify any files, paths, etc. I tried the outdated recipe from this repository and was a bit shocked to see that it was copying/modifying files in my local .`ipython` directory

jcb91:
@RinaldoB Ok, well that seems feasible once we've settled on a new package name 😜 

>  conda-forge might be an option for that

I've tried making a conda-forge PR for my `themysto` version of this repo (https://github.com/conda-forge/staged-recipes/pull/786), but haven't yet got it to build correctly.

> the building is really only meant to build a conda package

yep, agreed, this is a result of our deeply non-standard current install script. Changes are in the pipeline already.

> a bit shocked to see that it was copying/modifying files in my local .ipython directory

Although the current install script is likely to try to write to `~/.jupyter` (which is, I agree shocking in itself), anything being created/edited in `~/.ipython` is probably more likely to be a result of [`jupyter_core.migrate`](https://github.com/jupyter/jupyter_core/blob/62af662a4b21fef6223ef04ff6d2f2e1cc536181/jupyter_core/migrate.py)?

RinaldoB:
@juhasch why was this closed?

will conda installations not be supported?

jcb91:
@RinaldoB it's probably been closed because it didn't work. I'm in the process of getting this package neatly pip-installable, then will get a conda recipe working. I've got a functioning conda-forge one for the configurator: https://github.com/conda-forge/jupyter_nbextensions_configurator-feedstock so I'm confident we can get one working for this package, it's just a lower priority.

PR #423, opened 2015-11-25
Follow Symlinks

hiaselhans:
When extensions are symlinked (development) they should also be available...


jcb91:
You mean symlinks in the `nbextensions` directory should be followed so that extensions in them are available on the config page? That should probably be an easy fix...

jcb91:
whoops, sorry, got confused and didn't realize you'd already done it. Seems like a good plan to me. :+1:

hiaselhans:
cool, thanks!

PR #427, opened 2015-12-04
gist_it: type -> input_type in gist_it.yaml

cel4:
This seems to be a small bug in the gist_it config.


jcb91:
Yes, looks like I didn't test that, d'oh! Thanks for spotting it

juhasch:
Thanks.

PR #430, opened 2015-12-06
Toc2 - some new features for toc extension

jfbercher:
# Table of Contents (2)

This is an update of "Table of Contents" extension, with the following new features:
- The toc window can now be moved on screen. The position and states (that is 'collapsed' and 'hidden' states) are remembered (actually stored in the notebook's metadata) and restored on the next session.
- Numerotation - The different headers in the notebook can be automatically numbered (with automatic update)
- Toc cell - A "toc cell" can be created at top of the notebook and its contents automatically updated. This cell can be moved elsewhere in the notebook andd its position will be remembered and restored. 

Two links have been added to the toc window in order to toggle numerotation and toc cell. 
   - the "n" link toggles automatic numerotation of all header lines
   - the "t" link toggles a toc cell in the notebook, which contains the actual table of contents, possibly with the numerotation of the different sections. 

- Inital configuration (applied when creating a new notebook) can be specified in the /nbextensions tab. Thenafter, configuration will be stored in the notebook itself. 

# Testing 
- At loading of the notebook, loading of the configuration and initial rendering of the table of contents was fired on the event "notebook_loaded.Notebook". Curiously, it happened that this event was either not always fired or detected. Thus I rely here on a combination of  "notebook_loaded.Notebook" and "kernel_ready.Kernel" instead. Because of that, I propose to not overwrite toc and I put the extension in "testing" to avoid any possible/unknown side effect, though it works nicely for me. 
- The present extension also includes a small workaround for https://github.com/ipython-contrib/IPython-notebook-extensions/issues/429

Please make issues (or updates/PR) as needed


JanSchulz:
> Toc cell - A "toc cell" can be created at top of the notebook and its contents automatically updated. This cell can be moved elsewhere in the notebook andd its position will be remembered and restored.

Just a quick question before I start reading the code: Is that available in nbconvert?

JanSchulz:
IMO: just put it in as toc instead of toc2?

JanSchulz:
@jfbercher Not sure if you are aware of this: https://github.com/minrk/ipython_extensions/pull/53 I've no idea if that still applies to the toc here, but maybe you find it relevant?

jfbercher:
@JanSchulz 
1. Yes, the toc cell will persist in the notebook and will be converted 
2. I do not wanted to override the original toc before this is sufficiently tested. I had a small issue as indicated in the readme. Also there is probably a problem with restoring the toc window if resolution changes. Of course, toc and toc2 shouldn't be used simultaneously (It should work however)
3. No I was not aware of https://github.com/minrk/ipython_extensions/pull/53 ;  this is  interesting. I was thinking to implement something like that but turned back because I did not figured a good way to do it (also I found the previous behaviour satisfactory). I will have a look. If it works for toc it will work here.

JanSchulz:
Ok, trying it now: 

* in the config: "Display toc window at startup (block/none)" -> Could you make that a checklist?
* There is still the "two headlines with the same content will both get a ToC link to the first one"-problem but with the numbered headline everything should be in place to make the ID of  a headline unique?
* Toc cell: it would be nice if the links would be left aligned with the next cell


Other than that: wow, great overhaul. I like!

JanSchulz:
Uih: I just got a endless loop which killed the notebook tabs in chrome? Not sure which of the three open notebook tabs was responsible (chrome showed all three under item in the task explorer). The last thing I did was editing a markdown headline which was twice in the notebook (the "link to the second headline goes to the first one"-case). 

Unfortunately, I can't reproduce... :-/

jfbercher:
- Just updated the toc2 extension with nested numbering in the toc-window, as in https://github.com/minrk/ipython_extensions/pull/53 ;

- To align headings links in the toc-cell, just edit the style at the end pf  main.css 
- do not know how to make a checklist in config...

jcb91:
@jfbercher, I think what @JanSchulz meant by a checklist (correct me if I'm wrong!), was that rather than having the config key `toc-wrapper_display` be a string which must be 'block' or 'none', it should be a boolean (and thus represented on the form as a checkbox, similarly to `toc_number_sections`). Once display is a boolean, you could use, for convenience, [`jQuery.toggle(display)`](http://api.jquery.com/toggle/#toggle-display) to show/hide the cell, and [`jQuery.is`](http://api.jquery.com/is/#is-selector)`(`[`':visible'`](https://api.jquery.com/visible-selector/)`)` to check its status.

JanSchulz:
Yep, checkbox or selections... Just not something to write in...

jfbercher:
Thanks @jcb91. I did not realized that it could be so simple. I just created a new boolean variable toc_window_display, and updated the config using a ternary operation like `cfg['toc_window_display'] ? 'block' : 'none'`. Now we have a nice checkbox for specifying the initial visible state of the toc_window. 
image
Best for the new year!

juhasch:
Very nice, I like it!

nbextensions/usability/toc2/main.js
```diff
@@ -0,0 +1,422 @@
...
+    return d;
+  };
+  
+  var create_toc_div = function () {
+    var toc_wrapper = $('<div id="toc-wrapper"/>')
+    .append(
+      $("<div/>")
+      .addClass("header")
+      .text("Contents ")
+      .click( function(){
```
juhasch: One thing I noticed is, that dragging the TOC to move it also collapses it.
Maybe it would be better to move the click function down to line 115 so the TOC will only collapse when you click on the '-' ?

jfbercher:
Thanks. This is indeed a good observation and nice suggestion. Done!

I also enable the resizable functionnality for the toc window (with a max-width of 300px, can be changed in css).

PR #434, opened 2015-12-07
Add notify.js plugin into ipython-notebook-extensions

acamps:
I think it would be a great inclusion to have the notify plugin. I find it extremely useful.

It is hosted here https://github.com/sjpfenninger/ipython-extensions and I have no credit for this PR, since it's a simple copy paste.

I want to contribute, let me know how I can improve this request.

Thanks!

### DIFF
--- a/nbextensions/usability/notify/notify.js
+++ b/nbextensions/usability/notify/notify.js
@@ -1,0 +1,160 @@
+/*
+
+*************************
+Display Web Notifications
+*************************
+
+Add this file to $(ipython locate)/nbextensions/
+
+*/
+
+define(["require"], function (require) {
+  "use strict";
+
+  var current_time = function() {
+    return new Date().getTime() / 1000;
+  };
+
+  var start_time = current_time();
+  var min_time = 0;
+  var enabled = false;
+  var first_start = true;
+  var busy_kernel = true;
+
+  var add_permissions_list = function () {
+    var ipython_toolbar = $('#maintoolbar-container');
+    var label = $('<span/>').addClass("navbar-text permissions-list").text('Notify:');
+    var select = $('<select/>')
+                  .attr('id', 'permissions-select')
+                  .attr('class', 'permissions-list form-control select-xs')
+                  .append($('<option/>')
+                  .attr('value', 'Disabled')
+                  .text('Disabled'));
+    ipython_toolbar.append(label).append(select);
+    select.change(function() {
+      var val = $(this).val();
+      if (val == 'Disabled') {
+        enabled = false;
+      } else {
+        enabled = true;
+        min_time = val;
+      }
+      save_state();
+    });
+    // Add options in addition to the default, 'Disabled'
+    // Options give the minimum kernel busy time in seconds after which a notification is displayed
+    var presets = [0, 5, 10, 30];
+    for (var i=0; i<presets.length; i++) {
+      var name = presets[i];
+      select.append($('<option/>').attr('value', name).text(name));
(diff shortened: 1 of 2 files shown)


jcb91:
@sjpfenninger, would this be ok with you? Anything you'd care to alter? Apologies if you've already answered this elsewhere that I haven't found.

@acamps, I have a few suggestions:

1. I think it might be a bit clearer to rename the extension to something a little more specific, like `Notify on kernel idle`, or `Kernel-idle notifications`?
2. Add [notification.png](https://github.com/sjpfenninger/ipython-extensions/blob/master/notification.png) as an icon (specified by a relative path in the yaml as in [nbextensions/usability/ruler/ruler.yaml](https://github.com/ipython-contrib/IPython-notebook-extensions/blob/master/nbextensions/usability/ruler/ruler.yaml))
3. Add a `README.md` (and associated entry in the yaml) adapted from [sjpfenninger/ipython-extensions/blob/master/README.md#notebook-web-notifications](https://github.com/sjpfenninger/ipython-extensions/blob/master/README.md#notebook-web-notifications).
4. The readme should note that this extension will only work in browsers which support the [Notification API](https://developer.mozilla.org/en-US/docs/Web/API/notification). That means no IE, no Safari Mobile, no Opera Mobile. Alternatively, see below for a shim.
5. In the current implementation, it looks like if `Notification` is not defined, the extension will just throw an error, since `"use strict"` was invoked, and Notification hasn't been declared as a variable/argument. It could be worth looking at the shim at [nickdesaulniers/fxos-irc/blob/master/js/notification.js](https://github.com/nickdesaulniers/fxos-irc/blob/master/js/notification.js) to get around this and the previous issue of incomplete support, or just throwing a less cryptic warning on fail than `Uncaught ReferenceError: Notification is not defined`.
6. If you fancy extra work (this is by no means necessary), currently, it seems that the length of time for which the kernel must be busy before its idle message generates a notification is stored in the notebook metadata. This is sensible, but it might be nice to also have a default value for any notebooks which don't have the metadata value specified yet, which could be stored in the extension configuration. You could look at [nbextensions/usability/ruler/main.js](https://github.com/ipython-contrib/IPython-notebook-extensions/blob/master/nbextensions/usability/ruler/main.js) and [nbextensions/usability/ruler/ruler.yaml](https://github.com/ipython-contrib/IPython-notebook-extensions/blob/master/nbextensions/usability/ruler/ruler.yaml) for inspiration on how to do that. This isn't necessary for anything though, just might be nice if you fancy tinkering.

acamps:
Thanks @jcb91 for your quick reply!

About your comments:
1 - Sure! Makes a lot of sense. I was busy and wanted to start something.
2 - Actually it's not an Icon as I thought at the beginning. It's a picture from @sjpfenninger for the README.
3,4,5 - Of course. I'll appreciate a lot your comments and update the code as soon as possible!
6 - Depending on how long the previous steps take, I'll do that too.

I'll try to work on them in the following days. Thanks again!

sjpfenninger:
Fine with me! (just include the MIT license/attribution in the README or something)

juhasch:
The modified MIT license ist specified in COPYING.rst

sjpfenninger:
@juhasch the question isn't what license your project is under but whether you've included the copyright notice/attribution from my code in this PR (created by @acamps but containing my code)

juhasch:
OK, added a readme including copyright notice in PR #447.

sjpfenninger:
Thanks!

PR #435, opened 2015-12-15
Fixed the TypeError: 'str' object is not callable in line 24

damontallen:
There were unneeded parenthesise in line 24 of setup.py after p.name().


juhasch:
Should be fixed in PR #448

Issue #393, opened 2015-11-04
Consider making the package pip-installable


cel4:
The current way of installing is very unintuitive. The `setup.py`here is more a shell-script than a real setup.py. Moreover it has hidden dependencies on third party modules.

Is there any reason for not using `setuptools` and simply install a dummy package to fit everything in the python ecosystem?

jcb91:
You're absolutely right, I noticed this the other day when looking at the
`pyyaml` dependency (which, I think, is the only one which isn't also a
dependency of the notebook proper). I thought about changing it, but so far
haven't got round to it. If you fancy giving it a go, by all means do :)

jcb91:
Actually, also, one other thing to consider is a potential rename of the
repo to fit with jupyter (see #334, most extensions, after all, work with
any kernel type). If a have no idea how pip might handle such a renaming,
but it'd be with thinking about (maybe even start with the new name
directly?)

jfbercher:
I was also thinking that installing from pip could be useful to users; I have an attempt in my repo. 

Could somebody test this that installs from my git repo, branch test_pip:

`pip install git+https://github.com/jfbercher/IPython-notebook-extensions@test_pip#egg=IPython-contrib-nbextensions --user`
This works in my configuration (debian linux + pure python) but i do not know about other configs (anaconda/windows/macos)... Please report if it works. 

Some details: I have written a setup.py which uses setuptools, install specified requrements and then calls the original setup.py (renamed as install.py)

juhasch:
I did not make `setup.py` use setuptool, because it is not a real Python package.
Considering all the questions coming up, it looks like it would be a good idea to make it so.

jfbercher:
The pip-install I described above works for a clean Windows/Anaconda 2.3 install. I just had to upgrade psutil (?).

JanSchulz:
normal package and then put the current installer into something so that `python -c "import nbextensions; nbextensions.install()"` does the work?

jfbercher:
I think that this what I attempted above with pip: the pip install fires setup.py which itself fires install.py. That seems to work (but is not the same package manager as conda). 

Since pip/setuptools also enable to deploy data_files in the user directory (.local/..  under Linux) it should be possible to copy nbextension/ directly from setup.py. It would still remain to update all paths as does @juhasch in the original setup.py

JanSchulz:
```
Collecting PyYAML (from pyaml->IPython-contrib-nbextensions)
[...]

PR #439, opened 2015-12-26
Pip install attempt

jfbercher:
This is an attempt to get a kind of pip installable version of the repo, *directly from the repo*, as suggested many times eg #393. It is no more necessary to clone the repo since pip will do the download of the last version of the master branch for the user. Also pip automatically launches the install. 

- Please note that this may not be a "good practice" since the repo is actually not a python package, and cannot be uninstalled via pip, but only manually for now. 

This builds on the new conda installation recipe #417 and simply adds a setup.py that checks for the dependencies and then calls both install.py and configure_nbextensions (@juhasch, please note that I had to rename configure-nbextensions into configure_nbextensions because python's import do not like dashes in filenames). 

**Usage**: if the PR is accepted, installation will be possible via
`pip install  https://github.com/ipython-contrib/IPython-notebook-extensions/archive/master.zip --user`
verbose mode can be enabled with -v switch eg `pip -v install ...` , and upgrade with a --upgrade. A system install can be done by omitting the --user switch.  
If it it works, we will need to update the documentation. 

**Testing**: For now, install can be tested from my personal repo
`pip install  https://github.com/jfbercher/IPython-notebook-extensions/archive/pip-install.zip --user`

After a first install, the distribution can be updated either 
- by adding --upgrade at the end of the command-line
- by first "uninstalling" via `pip uninstall Python-contrib-nbextensions` (this is not a true uninstall but simply an erase of the install reference in the pip manager)

This PR was tested on debian 9, with a vanilla Python/Jupyter install, and under windows 8.1 + anaconda 2.3 distrib. 
Hope it will work for others and will be useful!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,7 @@ coverage.xml
 docs/_build/
 
 .idea
+
+#pip install
+Python_contrib_nbextensions.egg-info/
+dist/
--- a/configure_nbextensions.py
+++ b/configure_nbextensions.py
@@ -26,6 +26,12 @@ def remove_old_config(configdata):
     return '\n'.join(lines[0:start] + lines[end+1:])
 
 
+def make_backup(filename):
+    import shutil
+    backup = filename + ".bak"
+    shutil.copy(filename,backup)
+
+
 def update_config(config_file):
     """ Update .py configuration file with new path to extensions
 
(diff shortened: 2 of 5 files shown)


jfbercher:
@juhasch, please note that I had to rename configure-nbextensions into configure\_nbextensions. I also included a couple of modifications 

- open(x, a+) instead of open(x,r) [to create the file if it doesn't exist]
- corrected the path for saving json config files; I think that jupyter_config_dir() was missing
- added a "r" before the string in an os.path.join() to interpret the string as raw. Otherwise it causes a bug in Windows when reading nbextensions.html. Specifically "c:\users" was interpreted as c: "unicode escape" etc

Best

juhasch:
Thanks.I noticed a few things when trying this out:
- My old `jupyter_notebook_config.json` configuration was overwritten (I had `"open_browser" : false` in it)  and not appended. Not sure why. We should probably make a backup copy, anyway.
- We should add the local path `.` to the template path in nbconvert. Makes having local templates easier and is also the default in the current `jupyter_nbconvert_config.py` template in the repo. I don't understand where this happens.
- The `Python_contrib_nbextensions.egg-info`  and `dist` directories should be added to `.gitignore`

Looks good otherwise to me.

jfbercher:
As far as I am concerned, my json configs are not overwrited? I have corrected a small error that may still have caused this. Anyway, I have introduced a backup of these files as well as for the jupyter_*_config.py. I have also updated the paths to include '.' in templates path.

jfbercher:
## Another pip-install attempt
 I have worked out an alternative pip-install for the IPython-contrib collection. The idea is to make a small package that takes care of dependencies, and installs a script that in turn, can be used to download and install the current master. Installation of the distribution to the current user directories, configuration files and paths is then done by
```
Jupyter_nb_contribs.py install
```
and probably via 
```
Jupyter_nb_contribs.exe install
```
under windows. 

**Caution: for now, the installation still overwrites existing files..** If you have made modifications, then backup or simply rename your existing nbextensions directory. 

This package can be seen at https://github.com/jfbercher/Jupyter_contrib_nbextensions_install_utility and can be tried using
```
pip install https://rawgit.com/jfbercher/Jupyter_contrib_nbextensions_install_utility/master/dist/Jupyter_contrib_nbextensions_install_utility-0.0.1.tar.gz
```
Tests, comments, suggestions, PR are most welcome!

juhasch:
Looks good. Will merge if you can rebase to current master.

Issue #428, opened 2015-12-04
codefolding preprocessor


jcfschulz:
the addition of the  '↔\n' string in pre_codefolding.py causes problems on python2 systems. should be u'↔\n'.

juhasch:
Should be fixed by #445

PR #445, opened 2015-12-31
bugfix in `extensions/pre_codefolding.py` 

jcb91:
for #428 (can't use unicode literals in python2 strings)


juhasch:
Thanks for takeing care of this.

PR #451, opened 2016-01-11
in `nbextensions/usability/notify`:

jcb91:
* bugfix for the yaml descriptor, which wasn't loading correctly due to multi-line string not being indented. Switched also to extended, rather than multi-line string, to avoid unnecessary newline chars.
 * add image as an icon, and in readme
 * more verbose links in readme


juhasch:
Looks good, thanks.

PR #452, opened 2016-01-11
in `nbextensions/usability/keyboard_shortcut_editor`:

jcb91:
* remove code which made the keyboard shortcuts dialog open on notebook load (it was left in from testing)
 * add an option to not show the edit controls, while keeping the edits already configured, for brevity of the dialog


juhasch:
Thanks. I noticed the automatic loading but did not have time to open an issue.

PR #456, opened 2016-01-14
Don't create backup of nonexistent files.

kyrre:
Backup of configuration files will fail if they do not exist.


juhasch:
Thanks. Should be fixed by PR #458

Issue #461, opened 2016-01-25
setup.py install doesn't work on Mac OS X again


qobilidop:
@jonnybaik brought it up in #431 and resolved it in #432, but the issue seems to appear again. I was following the instructions in [README.md](https://github.com/ipython-contrib/IPython-notebook-extensions/blob/b124adca81c2c4e414b0487487eb91843d7d5b8a/README.md) to install the notebook extensions through conda. I checked out the master branch, then ran `conda build IPython-notebook-extensions`, but received the following error:

```
Traceback (most recent call last):
  File "/Users/qobilidop/miniconda3/envs/_build/lib/python3.5/site-packages/psutil/_psosx.py", line 199, in wrapper
    return fun(self, *args, **kwargs)
  File "/Users/qobilidop/miniconda3/envs/_build/lib/python3.5/site-packages/psutil/_psosx.py", line 238, in cmdline
    return cext.proc_cmdline(self.pid)
ProcessLookupError: [Errno 3] No such process

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "setup.py", line 52, in <module>
    import configure_nbextensions
  File "/Users/qobilidop/miniconda3/conda-bld/work/configure_nbextensions.py", line 62, in <module>
    if ("python" or "jupyter") in p.name():
  File "/Users/qobilidop/miniconda3/envs/_build/lib/python3.5/site-packages/psutil/__init__.py", line 546, in name
    cmdline = self.cmdline()
  File "/Users/qobilidop/miniconda3/envs/_build/lib/python3.5/site-packages/psutil/__init__.py", line 599, in cmdline
    return self._proc.cmdline()
  File "/Users/qobilidop/miniconda3/envs/_build/lib/python3.5/site-packages/psutil/_psosx.py", line 209, in wrapper
    raise ZombieProcess(self.pid, self._name, self._ppid)
psutil.ZombieProcess: psutil.ZombieProcess process still exists but it's a zombie (pid=54)
```

which is extremely similar to the one in the previous issue #431.

But since the [`setup.py`](https://github.com/ipython-contrib/IPython-notebook-extensions/blob/b124adca81c2c4e414b0487487eb91843d7d5b8a/setup.py) looks completely different now after @jfbercher's commits, I don't know how to apply the previous solution here. Could anyone help?

My setup:
Mac OS X 10.11.3 (El Capitan)
Python 3.5.1, installed using conda

jfbercher:
[...]

PR #432, opened 2015-12-06
Fixed issue when checking for jupyter-notebook process on OS X 

jonnybaik:
Resolves #431

Issue #431, opened 2015-12-06
setup.py install doesn't work on Mac OS X


jonnybaik:
I was following the instructions to install the notebook extensions by following the instructions in the [README](https://github.com/ipython-contrib/IPython-notebook-extensions/blob/master/README.md). I checked out the master branch, then ran `python setup.py install`, but received the following error:

```
$ python setup.py install
Traceback (most recent call last):
  File "/usr/local/var/pyenv/versions/3.4.3/lib/python3.4/site-packages/psutil/_psosx.py", line 199, in wrapper
    return fun(self, *args, **kwargs)
  File "/usr/local/var/pyenv/versions/3.4.3/lib/python3.4/site-packages/psutil/_psosx.py", line 238, in cmdline
    return cext.proc_cmdline(self.pid)
ProcessLookupError: [Errno 3] No such process

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "setup.py", line 24, in <module>
    if "jupyter-notebook" in p.name():
  File "/usr/local/var/pyenv/versions/3.4.3/lib/python3.4/site-packages/psutil/__init__.py", line 539, in name
    cmdline = self.cmdline()
  File "/usr/local/var/pyenv/versions/3.4.3/lib/python3.4/site-packages/psutil/__init__.py", line 592, in cmdline
    return self._proc.cmdline()
  File "/usr/local/var/pyenv/versions/3.4.3/lib/python3.4/site-packages/psutil/_psosx.py", line 209, in wrapper
    raise ZombieProcess(self.pid, self._name, self._ppid)
psutil.ZombieProcess: psutil.ZombieProcess process still exists but it's a zombie (pid=64)
```

It seems to crash when it checks to see if jupyter-notebook currently running.

My setup:

Mac OS X 10.11.1 (El Capitan)
Python 3.4.3, installed using pyenv

jonnybaik:
The lines of code in `setup.py` causing the crash:

```
 for p in psutil.process_iter():
     if "jupyter-notebook" in p.name():
         print("Cannot install while the Jupyter notebook server is running")
         exit()
```

When I google the problem, I find this archived thread discussing how `psutil` not properly handling zombie processes in OS X, throwing an exception when it comes across a zombie process.

I solved the issue by handling the zombie process exception by doing nothing:

```
for p in psutil.process_iter():
    try:
        if "jupyter-notebook" in p.name():
[...]

PR #462, opened 2016-01-25
When configuring, ignore Zombie processes on OS X

baldwint:
This patch was originally implemented by @jonnybaik, but got overwritten somewhere along the line. See Issue #431, Issue #432, and Issue #461


jfbercher:
Well done! Thx.

Issue #469, opened 2016-02-10
ExecutionTime - timing is not right, outputs to one cell whenn all cells are running


denfromufa:
This is what happens on second cell:

```
Lastly executed on Wed, Feb 10 2016 at 10:58 AM in 0 s in 1 s in 0 s in 1 s in 0 s in 0 s in 0 s in 0 s in 0 s in 1 s in 0 s in 39 s in 0 s in 0 s in 0 s in 0 s in 2 s in 0 s in 0 s in 0 s in 2 s in 2 s in 1 s in 1 s in 1 s in 0 s in 0 s in 0 s in 0 s in 0 s in 0 s in 2 s in 0 s in 0 s in 0 s in 1 s in 0 s in 0 s in 1 s in 0 s in 15 s in 2 s in 0 s in 0 s in 0 s in 0 s in 1 s in 1 s in 1 s in 2 s in 0 s in 1 s in 0 s in 0 s in 1 s in 1 s in 1 s in 1 s in 0 s in 0 s in 0 s in 1 s in 0 s in 1 s in 0 s in 0 s in 0 s in 1 s in 0 s in 0 s in 1 s in 0 s in 1 s in 1 s in 1 s in 1 s in 22 min 23 s
```

denfromufa:
@juhasch

denfromufa:
The version of the notebook server is 4.1.0 and is running on:
Python 2.7.9 |Continuum Analytics, Inc.| (default, Dec 12 2014, 14:56:54) [MSC v.1500 32 bit (Intel)]

jcb91:
This sounds like a duplicate of #436. I've recently done a rewrite of execute_timings (see #470) which would fix this, if you'd like to try that but if you'd like me to try to fix the existing implementation, @denfromufa, you could take a look at my reply to #436. I take your comments above to mean that:
1. You execute a bunch of cells (2? more than 2?) in one call (or queue several before the first execution is complete)
2. Whenever any cell completes, `'in 1 s'` (or similar) is added to the timing text on the _second_ (is this the last?) cell _instead of_ (as well as?) the appropriate cell.

lucasb-eyer:
While I'm not @denfromufa, I came here to report the same issue. I tried #470 and it seems to fix all of these issues, thanks a lot!

Just for clarification:

1. Any amount more than one cell. Can be queued by "Run all"-style commands, or Ctrl/Shift-enter while another one still runs. All ways trigger the problem.
[...]

Issue #436, opened 2015-12-18
Possible bug in ExecuteTime extension


enascimento:
When running a cell several times, I get output like:
"Lastly executed on Fri, Dec 18 2015 at 9:18 PM in 3 min 13 s in 3 min 13 s in 3 min 20 s in 3 min 21 s"

There is a repetition of the elapsed time in the end of the line.

jcb91:
Does this happen on every cell, every notebook? Even after reloading? Does aborting kernel executions affect things? Could you post the output of the [browser JavaScript console] (http://webmasters.stackexchange.com/questions/8525/how-to-open-the-javascript-console-in-different-browsers)?

I'm not sure exactly why this is happening without more information, but i can suggest a couple of possibilities.

The first would be that the extension is somehow getting loaded multiple times, which causes each cell to register its execution call multiple times. This is fixed in the latest master jupyter, so extensions should only be loaded once each, but I don't recall which version fixed it. A quick check of the javascript console should show whether you have multiple "loaded extension ExecuteTime" messages. 

The second: as far as I can make out, the extension maintains a list of cells which have been executed, then pops them out in fifo order when a kernel idle notifications are fired. I don't think the list gets cleared if the kernel is told to abort evaluation, which, if there are multiple cells in the queue, might mean that they don't all get popped from the list, so when a subsequent execution happens, they may get extra times...

jcb91:
presumably now closed by #470

PR #470, opened 2016-02-10
update `nbextensions/usability/execute_time`:

jcb91:
* update last-execution immediately on queueing the cell
 * alter extension to use kernel 'execute_reply' instead of 'idle' for increased accuracy
 * add start/end timings to cell metadata, restore messages on notebook load
 * use moment library for formatting times
 * improve messages and formatting

### DIFF
--- a/nbextensions/usability/execute_time/ExecuteTime.js
+++ b/nbextensions/usability/execute_time/ExecuteTime.js
@@ -46,7 +46,7 @@ define([
                     });
                     var timing_area = update_timing_area(cell);
                     if ($.ui !== undefined) {
-                        timing_area.stop(true, true).show(0).effect('highlight', {color: '#0F0'});
+                        timing_area.stop(true, true).show(0).effect('highlight', {color: '#0B0'});
                     }
                 }
                 else {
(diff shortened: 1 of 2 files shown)


lucasb-eyer:
I tested it. Works great and I believe it fixes both #469 and #436, albeit neither OP ever clarified.

nbextensions/usability/execute_time/ExecuteTime.js
```diff
@@ -1,182 +1,240 @@
...
+                if (msg.msg_type === 'execute_reply') {
+                    $.extend(true, cell.metadata, {
+                        ExecuteTime: {
+                            start_time: msg.metadata.started,
+                            end_time: msg.header.date
+                        }
+                    });
+                    var timing_area = update_timing_area(cell);
+                    if ($.ui !== undefined) {
+                        timing_area.stop(true, true).show(0).effect('highlight', {color: '#0F0'});
```
lucasb-eyer: A very minor nitpick, but would you mind changing that color to `#0B0`? The pure green is disturbingly bright, and I guess the original code used `#b00000` for that same reason.

jcb91:
good plan, I hadn't noticed I'd changed it, will alter it now.

PR #471, opened 2016-02-11
Made TOC draggable

kmbroga:


JanSchulz:
Nice!

Could you also make the change to toc2 version?

@jfbercher I think we should merge toc2 into toc and remove toc2... I'm using toc2 in my phd notebooks and it works :-)

kmbroga:
I was not aware of toc2 till now...  Have just checked it and it appears to be draggable already.

jfbercher:
Nice to hear. I still have some small improvements to submit (unfortunately I enabled too rapidly a resizable feature last time, which in turn introduced some blitches). I'll do it this week-end, and if ok we may merge.

cqcn1991:
Hi, @JanSchulz @jfbercher , I have a few idea want to experiment with `toc2`, but how can I get started? I know html, css and js, but didn't the workflow to build a `Ipython Notebook` extension (https://github.com/minrk/ipython_extensions/issues/54)

jcb91:
@cqcn1991 there's a little documentation at [jupyter-notebook/docs/source/extending/frontend_extensions.rst](https://github.com/jupyter/notebook/blob/620fb2919229d1c0b5e79087edcaaf95b770473b/docs/source/extending/frontend_extensions.rst).

Essentially though, if you want to create a copy of toc, you need to:

1. Find the original toc extension. You can do this from python with the following snippet:
  ```python
import os.path
from jupyter_core.paths import jupyter_path

def find_extension(ext_path):
    for base in jupyter_path():
        curr_path = os.path.join(base, 'nbextensions', ext_path)
        if os.path.exists(curr_path):
            print('found extension at {!r}'.format(curr_path))

find_extension('usability/toc')
  ```
(for me, the above outputs
`found extension at '/Users/josh/Library/Jupyter/nbextensions/usability/toc'`)
2. copy the `toc` directory to a new name, e.g. 'mytoc'
3. To get your new extension to show up correctly on the config page, edit its yaml file to give it a different name
4. customize away any file in your new extension directory

There isn't a build process for frontend-only extensions, as they're dynamically loaded as javascript, without any minification.

janschulz:
IMO this can be closed as toc1 is removed and toc2 has this functionality.

jcb91:
:+1:

PR #472, opened 2016-02-13
nbextensions/usability/toc2/: Fixed problems with skipped heading levels,  nested TOC lists

slonik-az:
* Table of Contents was empty when the first heading is `<h2>` or higher.
* TOC had incorrectly numbered labels when some heading levels are skipped over.
* Make calculations of numbered labels more robust.
* Minor code cleanup

### DIFF
--- a/nbextensions/usability/toc2/main.js
+++ b/nbextensions/usability/toc2/main.js
@@ -10,7 +10,7 @@ define(["require", "jquery", "base/js/namespace",  'services/config',
 
 // ...........Parameters configuration......................
  // define default values for config parameters if they were not present in general settings (notebook.json)
-    var cfg={'toc_threshold':4, 
+    var cfg={'toc_threshold':6, 
              'toc_number_sections':true, 
              'toc_cell':false,
              'toc_window_display':false}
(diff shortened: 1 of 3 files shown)


jfbercher:
Hi @slonik-az. 
I have checked the code and tested your modifications. This is a nice work that largely improves over what I did. This is very useful -- I was myself annoyed to be forced to begin the notebook with a level-1 heading. Thanks. 
@juhasch, I think that we can merge as is.

nbextensions/usability/toc2/main.js
```diff
@@ -274,77 +273,60 @@ var table_of_contents = function () {
...
+
+    //loop over all headers
+    all_headers.each(function (i, h) {
+      var level = parseInt(h.tagName.slice(1), 10) - min_lvl + 1;
       // skip below threshold
-      if (level > threshold) return;
+      if (level > threshold){ return; }
       // skip headings with no ID to link to
-      if (!h.id) return;
+      if (!h.id){ return; }
```
JanSchulz: BTW: IMO it would be nice to a) add IDs so that you wouldn't need to add them on your own if you display H1 from a  kernel and b) to check that a ID wasn't already included in the list if if it was, do some magic that the links work anyways...

[The last is a problem when including something like this: `"analysis 1": "plots", "stats"; analysis 2: "plots", "stats"` -> the links for both "stats" all go to the first one because the notebook does not render unique IDs -> https://github.com/jupyter/notebook/issues/77]

slonik-az:
@jfbercher, @juhasch:
Thanks for merging my patch into master!  toc2 is a great extension which makes notebook navigation so much more fun. Making it more robust benefits everybody.
--Leo

nbextensions/usability/toc2/main.js
jfbercher: I finally understood what is the actual problem... I did an essay to address it (since we number headings, we can try to use this number as a kind of unique marker). Of course external links may fail when toc changes but I suggest a way to address this too -- see #488

PR #475, opened 2016-02-16
updates to `nbextensions/usability/collapsible_headings`:

jcb91:
* Make the heading toggle buttons optional. This also necessitates registering existing cells only after config has loaded
* add `collapsible_headings_collapsed` css class to the cell, rather than the toggle button
* remove `cell.metadata.heading_collapsed` value from non-heading cells
* update readme

### DIFF
--- a/nbextensions/usability/collapsible_headings/collapsible_headings.yaml
+++ b/nbextensions/usability/collapsible_headings/collapsible_headings.yaml
@@ -32,6 +32,14 @@ Parameters:
   description: Adjust the size of the toggle controls to match their heading levels
   input_type: checkbox
   default: false
+- name: collapsible_headings_show_section_brackets
+  description: show Mathematica-style brackets around each collapsible section
+  input_type: checkbox
+  default: false
+- name: collapsible_headings_show_ellipsis
+  description: show a gray bracketed ellipsis at the end of collapsed heading cells
+  input_type: checkbox
+  default: true
 - name: collapsible_headings_use_shortcuts
   description: Add command-mode keyboard shortcuts to collapse/uncollapse the selected heading cell
   input_type: checkbox
--- a/nbextensions/usability/collapsible_headings/collapsible_headings.yaml
+++ b/nbextensions/usability/collapsible_headings/collapsible_headings.yaml
@@ -11,9 +11,27 @@ Parameters:
   input_type: checkbox
   default: false
 - name: collapsible_headings_use_toggle_controls
-  description: Add a button in each heading cell's input prompt to collapse/uncollapse it
+  description: Add a control in each heading cell's input prompt to collapse/uncollapse it
   input_type: checkbox
   default: true
+- name: collapsible_headings_toggle_color
+  description: Color for the toggle control icon
+  input_type: color
+  default: '#aaa'
+- name: collapsible_headings_toggle_closed_icon
+  description: font-awesome class for the toggle control icon on collapsed headings
+  default: fa-caret-right
+- name: collapsible_headings_toggle_open_icon
+  description: font-awesome class for the toggle control icon on uncollapsed (expanded) headings
+  default: fa-caret-down
+- name: collapsible_headings_make_toggle_controls_buttons
+  description: "Make the toggle control into a button (if false, it's just an icon)"
+  input_type: checkbox
+  default: false
+- name: collapsible_headings_size_toggle_controls_by_level
+  description: Adjust the size of the toggle controls to match their heading levels
+  input_type: checkbox
+  default: false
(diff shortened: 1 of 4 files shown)


Kevin-McIsaac:
Josh
Really nice extension, thanks. I use collapsed headings daily now and its one of the most useful additions to a notebook.

With regards to the toggle buttons, I too found them overly large. As an alternative to removing them, perhaps use a small, grey triangular toggle. This is pretty much the UI standard for collapsible headings.

Here is an example of it being used
http://codepen.io/echozone/details/LwJCD 

and  some idea in CSS
http://stackoverflow.com/questions/19913491/toggle-triangle-on-dropdown-jquery

If you wanted to be fancier, look at how it done in a Mathematica notebook (which was the original notebook UI) look at these two.
https://reference.wolfram.com/language/tutorial/UsingANotebookInterface.html
https://reference.wolfram.com/language/guide/NotebookBasics.html

Note the group brackets to the right shows the scope of what will be collapsed. You double click anywhere on the bracket to open/close.

jcb91:
Thanks Kevin, glad you like it :)

Small grey carets (like the old hierarchical collapse, but preferably using fa icons) would certainly be less obtrusive, I'll add an option.

I did originally try the mathematica brackets, but couldn't think of a neat way to get them done purely in CSS, so abandoned the idea as too complicated. I didn't want to use an overlay div, as it gets complicated trying to keep track of resizing windows and such. I'd thought of using a border for the cell element, but that doesn't really work for nested brackets. The only way I can think of at the moment to add nested brackets would be to add extra elements to the right of each cell, but that'd need one in each of the input area, widget area and output area, would require fiddly handling for the tops and bottoms of the brackets depending on which elements were visible, and would break given any padding between cells sub-areas or between cells. If you can think of a neat way to do it, by all means let me know and I can give it a shot, though!

I also thought of adding a bottom border to collapsed heading cells, but didn't really want to mess with the stock css too much. Anyway, now that the `collapsible_heading_collapsed` css class is applied to the cell rather than the toggle control, a user could add a rule in their custom css to add a bottom border, if they wanted.

jcb91:
Ok, I've added a bunch more options, but no mathematica-style brackets yet, I'm afraid.

OndrejMarsalek:
Thanks for the updates, the extension is already very useful and this makes it nicer. Would perhaps a small (gray?) bracketed ellipsis at the end of each collapsed heading be a good way to indicate that it is collapsed? Like this:

### Heading

And some text.

### Heading 2 [...]

jcb91:
@OndrejMarsalek mhm, I see what you mean about the bracketed ellipsis, it's a nice idea.

I managed to get a quick-and-dirty mockup done in pure css by tagging it onto the end of the heading as a pseudo-element, but it's far from ideal as it fights for space with the anchor link. In addition, it sort of looks like it should be clickable, but as a pseudo-element, can't be.

```css
.collapsible_headings_collapsed .rendered_html h1:after,
.collapsible_headings_collapsed .rendered_html h2:after,
.collapsible_headings_collapsed .rendered_html h3:after,
.collapsible_headings_collapsed .rendered_html h4:after,
.collapsible_headings_collapsed .rendered_html h5:after,
.collapsible_headings_collapsed .rendered_html h6:after {
  content: "[\002026]";
  color: #aaa;
}
```

The real way to go about it would be to insert the ellipsis after rendering, but I think that might be slightly complicated to implement correctly because of the way the markdown is rendered, which is asynchronously, so that a function bound to the textcell execute event may (will? certainly is at least sometimes) be executed before all the markdown is in place. And that still doesn't solve the problem of conflicting with the anchor link...

jcb91:
So, with a little relative/absolute positioning tinkering, I've added ellipses that I think are passable, and even Mathematica-style brackets.

I think there's now plenty of options for discreetly showing (un)collapsed heading cell status, with the mini caret icons (turn icon size-matching option off), ellipses (optional), and right-hand brackets (again, optional).

FYI, Since the Mathematica-style brackets can be affected by headings below them opening/closing, it means that if they're enabled, the brackets for the whole notebook are regenerated every time a heading is toggled. This isn't the smartest scheme ever, and in really long notebooks might turn out to be a bit sluggish, but seemed ok in my testing, and, after all, it's optional anyway. Trying to make the updates smarter (skipping unaffected cells) is of questionable value in shorter notebooks, and is partly what got hierarchical collapse into buggy territory in the past, so I've avoided it for now.

OndrejMarsalek:
Thank you for the update, @jcb91, this is great. Personally I like the ellipses the best (especially aligned to the right), but I think it is best to have these various configuration options for different people. :+1:

PR #476, opened 2016-02-17
Configure to load all server extensions apart from `nbextensions`.

haobibo:
Find all server extensions in folder jupyter_data_dir()/extensions, and
append them to newconfig.NotebookApp.server_extensions, so that server
extensions apart from `nbextensions` can be loaded after install.

### DIFF
--- a/configure_nbextensions.py
+++ b/configure_nbextensions.py
@@ -151,9 +151,19 @@ update_config(py_config)
 json_file = 'jupyter_notebook_config.json'
 config = load_json_config(json_file)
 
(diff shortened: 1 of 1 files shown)


jcb91:
Nice idea, but I think we ought to avoid enabling any extensions that we didn't install - so maybe better to list the directory we install _from_ rather than the one we install _to_?

haobibo:
@jcb91 yeah, I have already changed the code, now list dir we install from.

jcb91:
:+1:

PR #478, opened 2016-02-18
bugfix in `nbextensions/usability/codefolding`…

### LINKED ISSUE #422: Fix codefolding so that cells in already created notebooks work
I can get codefolding to work on newly created notebooks after I activate it. 
I can't get it to even show the gutter with code-folding arrow on existing notebooks . 

I get a Javascript error in foldcode.js on existing notebooks that doesn't occur on newly created ones. 
TypeError: funcs[i] is not a function
```var found = funcs[i](cm, start);```
on line 112 of  foldcode.js
```
  CodeMirror.registerHelper("fold", "combine", function() {
    var funcs = Array.prototype.slice.call(arguments, 0);
    return function(cm, start) {
      for (var i = 0; i < funcs.length; ++i) {
        var found = funcs[i](cm, start);
        if (found) return found;
      }
    };
  });
```
Missing metadata on already created cells/notebooks?

jcb91:
Fixes #422

Addresses a dependency race condition with a require call.

### DIFF
--- a/nbextensions/usability/codefolding/firstline-fold.js
+++ b/nbextensions/usability/codefolding/firstline-fold.js
@@ -5,10 +5,10 @@ CodeMirror.registerHelper("fold", "firstline", function(cm, start) {
       var lineText = cm.getLine(start.line);
       var found = lineText.lastIndexOf(Token,0);
       if (found == 0) {
-        end =  cm.lastLine(); 
+        end =  cm.lastLine();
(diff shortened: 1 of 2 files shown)


juhasch:
This seem only to change a few spaces ?

jcb91:
apologies, I hadn't staged what I thought I had. Updated....

juhasch:
Let's hope it works better now.

PR #483, opened 2016-02-21
Fixed toc window resizing issues

jfbercher:
This commit fixes some issues when resizing the toc window: 
- the text did not resized as the window and overflowed
- there was no possible scrolling
Now the text is reflowed according to the dimensions of the new window, overflow is avoided and a scroll bar appears when needed. 

This commit includes the modifications by @slonik-az in #472. Thus should be merged after this commit.

PR #488, opened 2016-02-22
toc resizing issues / unique anchors in headings

jfbercher:
I have closed #483 and redo/compete it:

- **toc resizing issues**: the commit fixes some issues when resizing the toc window:
    - the text did not resized as the window and overflowed
    - there was no possible scrolling Now the text is reflowed according to the dimensions of the new window, overflow is avoided and a scroll bar appears when needed.

- **create unique anchors in headings and update links in tocs**. This tries to address the longstanding issue https://github.com/jupyter/notebook/issues/77 pointed out by @JanSchulz. Hope it will help...
     - The commit adds current toc number to headings anchors. This enables to get unique anchors for recurring headings with the same text. This is not perfect since external links may fail if toc changes (eg headings are added/removed). Anyway, one can always add an anchor `<a name="myanchor"></a>` in the heading and refer to that anchor, eg `[link](#myanchor)`.  This anchor is automatically removed when building toc links.


jfbercher:
Just edited the last sentence in the presentation, where the html was interpreted, and thus the axample was difficult to understatnt. Please read it as 
"This is not perfect since external links may fail if toc changes (eg headings are added/removed). Anyway, one can always add an anchor `<a name="myanchor"></a>` in the heading and refer to that anchor, eg `[link](#myanchor)`.  This anchor is automatically removed when building toc links.  "

nbextensions/usability/toc2/README.md
```diff
@@ -24,15 +24,17 @@ first occurrence.
...
+- This extension also includes a quick workaround as described in https://github.com/ipython-contrib/IPython-notebook-extensions/issues/429
 
 ## History
 
 - This extension was adapted by minrk https://github.com/minrk/ipython_extensions
 from https://gist.github.com/magican/5574556
 - Added to the ipython-contrib/IPython-notebook-extensions repo by @JanSchulz
-- Modified by @junasch (automatic update on markdown rendering), 
+- @junasch, automatic update on markdown rendering, 
 - @JanSchulz, (enable maths in headers links)
```
JanSchulz: With or without `(...)`?

nbextensions/usability/toc2/README.md
```diff
@@ -24,15 +24,17 @@ first occurrence.
...
-- @jfbercher december 06, 2015 -- (automatic numbering, toc cell, window dragging, configuration parameters)
+- @jfbercher december 06, 2015 -- Big update: automatic numbering, toc cell, window dragging, configuration parameters
 - @jfbercher december 24, 2015 -- nested numbering in toc-window, following the fix by [@paulovn](https://github.com/minrk/ipython_extensions/pull/53) in @minrk's repo. December 30-31, updated config in toc2.yaml to enable choosing the initial visible state of toc_window via a checkbox ; and now resizable. 
-
+- @slonik-az february 13, 2016. rewritten toc numberings (more robust version), fiwed problems with skipped heading levels, some code cleanup
```
JanSchulz: s/fiwed/fixed/

nbextensions/usability/toc2/README.md
```diff
@@ -24,15 +24,17 @@ first occurrence.
...
+- @jfbercher february 21, 2016. Fixed some issues when resizing the toc window. Now avoid overflows, clip the text and add a scrollbar. 
+- @jfbercher february 22, 2016. Add current toc number to headings anchors. This enable to get unique anchors for recurring headings with the same text. This is not perfect since external links may fail if toc changes (eg headings are added/removed). It is still possible to directly an html anchor within the heading text. This is taken into account when building toc links (see comments in code).
```
JanSchulz: ... o directly **+add+** an html anchor ...

nbextensions/usability/toc2/main.js
```diff
@@ -307,7 +333,14 @@ var table_of_contents = function () {
           ul= ul.parent();
           while(!ul.is('ul')){ ul= ul.parent(); }
       }
-
+      // Change link id -- append current num_str so as to get a kind of unique anchor
```
JanSchulz: If this means (just came from the dentist and JS is not my strongest language...) that all headline IDs get changed then this means that all links get bad. A not so drastic solution would be to use additional IDs for the Toc links *if these links are not unique*.

JanSchulz:
@jfbercher Wow, thank you! This is really something because it annoyed me like hell during my analyses... Do you have an amazon wishlist? :-)

nbextensions/usability/toc2/main.js
jfbercher: Yes, all ids are changed, but all links are updated accordingly; so this essentially works. [Detection of non unique toc links would add overhead.]

JanSchulz: m, ok, so the only thing which goes really bad is manually inserted links in the same notebook (e.g. a manual TOC or to jump around)... For links to certain headlines in nbviewer, I would probably also get these in nbviewer, so the URLs there are the old ones and have the same problems as the old solution but still work for the other cases...

Ok, I rest my case :-)

jfbercher:
Commit "	Add anchors to original IDs" :    

The original ID is also preserved and an anchor is created to refer to it and can be used (but toc uses the new IDS!). It is thus possible to link to the heading line using both the old style (but ids may not be unique) or the new style. 
Finally it is possible to link to a heading line by:  `[link](#initialID)`, or `[link](#initialID-tocnumber)` or `[link](#myanchor)` [if one defines his own anchor]

nbextensions/usability/toc2/main.js
jfbercher: On Mon, Feb 22, 2016 at 10:18 PM, Jan Schulz <notifications@github.com>
wrote:

> In nbextensions/usability/toc2/main.js
> <https://github.com/ipython-contrib/IPython-notebook-extensions/pull/488#discussion_r53694442>
> :
>
> > @@ -307,7 +333,14 @@ var table_of_contents = function () {
> >            ul= ul.parent();
> >            while(!ul.is('ul')){ ul= ul.parent(); }
> >        }
> > -
> > +      // Change link id -- append current num_str so as to get a kind of unique anchor
>
> m, ok, so the only thing which goes really bad is manually inserted links
> in the same notebook (e.g. a manual TOC or to jump around)... For links to
> certain headlines in nbviewer, I would probably also get these in nbviewer,
> so the URLs there are the old ones and have the same problems as the old
> solution but still work for the other cases...
>
> Ok, I rest my case :-)
>
I have added an anchor to the original ID. So in principle you can  link to
the heading line ever using the old-style or the new one.
Now I must go back to work...

juhasch:
Not sure if this is related to the new ID generation:
`# Heading Name` creates a TOC entry
`# Heading/Name` not
`# Heading:Name` not

jfbercher:
Thanks. Good tests. 
Indeed, this was related.... 
These special characters still caused a problem when checking for the existence of new ids (before creating them). I have now escaped all special characters in my keyboard (and "normal" extended characters work), so it should work now
image

juhasch:
This fixes it. Looks good now!
I wonder, should we remove the old toc extension to avoid confusion. This one is now much better.

JanSchulz:
> I wonder, should we remove the old toc extension to avoid confusion. This one is now much better.

yes please :-)

juhasch:
OK. Let's merge the changes for toc2 and remove old toc.

jfbercher:
OK, nice :-) 
Perhaps should we edit the docs in toc2 and rename it as toc?

juhasch:
Will there be confusion with minrk's original extension ?

jfbercher:
Yes, you are right.
Let's keep it as toc2.

JanSchulz:
Do you think the name on the disc matters?

JanSchulz:
No sure if this is something this started, but I notice that the execution of an md cell now has a noticeable delay. Is there any way to do the regeneration of the toc in the background?

jcb91:
@JanSchulz perhaps try a setTimeout in the event `rendered.MarkdownCell` callbacks registered in [nbextensions/usability/toc2/main.js#L434-L444](https://github.com/ipython-contrib/IPython-notebook-extensions/blob/991ae98250d68c70a473478502c5467c98e63eb4/nbextensions/usability/toc2/main.js#L434-L444)?

JanSchulz:
@jcb91 I change it to 

```
    $([IPython.events]).on("notebook_loaded.Notebook", function(){ // curiously, the event is not always fired or detected
                                                       // thus I rely on kernel_ready.Kernel to read the initial config 
                                                       // and render the first  table of contents
            read_config(); 
        table_of_contents(); 
        // render toc for each markdown cell modification
        //$([IPython.events]).on("rendered.MarkdownCell", table_of_contents);
        $([IPython.events]).on("rendered.MarkdownCell", function(){setTimeout(table_of_contents,100)});
            console.log("toc2 initialized (via notebook_loaded)")
        extension_initialized=true  ; // flag to indicate that initialization was done
})

    $([IPython.events]).on("kernel_ready.Kernel", function(){
        if (!extension_initialized){
            read_config(); 
            table_of_contents(); 
            // render toc for each markdown cell modification
            $([IPython.events]).on("rendered.MarkdownCell", function(){setTimeout(table_of_contents, 100)});
            console.log("toc2 initialized (via kernel_ready)")
        }
```

and it seems to be faster. Is that the way you wanted it?

jfbercher:
As far as I am concern, I do not noticed any delay introduced by the last modifications and indeed these add a very small amount of new calculations. 
I have some big notebooks with big tocs and I have no problems with them. 
There are much more consuming oerations that happens when you render a md cell, eg the rendering itself, mathjax, what happens in latex_envs, in highlighter, etc. Furthermore, javascript is asynchronous and is able to process several things together. 
*I can certainly be wrong*, but I don't think that the processing delay you notice is due to the last modifications in toc, or to toc itself. These kind of delays may happen when there are lot of things that are written in cell outputs (and that may behidden), eg javascript code or html via display. You should inspect your notebook, the extensions you use, perhaps copy cells to another nb. 
Anyway, if delaying toc computation by 0.1s does something noticeable, we could implement it! I have absolutely no opposition to that. But it would be nice to better understand from where comes the problem.

JanSchulz:
@jfbercher How would you debug that? Up to now I haven't done any profiling in the browser :-(

jfbercher:
@JanSchulz Can you create a repo, eg "toc_test",  in your github account and upload an example of notebook with this slow rendering problem? I would like to check with one of your files since I do not have  this feeling myself.

JanSchulz:
@jfbercher Ok, I tried to recreate it by removing the timepout but now I can't: it's fast even when I use the old code (as seen in code inspect in chrome). I've no idea why this happend or why it went away when I used the timeout.

Sorry for the noise :-(

PR #491, opened 2016-02-22
Change print statements to print functions in aspell server

JanSchulz:


jcb91:
This should probably include a 

```python
from __future__ import print_function
```

For best compatibility. In addition, there are currently two copies of `ipy-aspell-server.py` in the repo, and the one that's causing the problems is
[extensions/ipy-aspell-server.py](https://github.com/ipython-contrib/IPython-notebook-extensions/blob/b508b0ffbda62ac935088f57b579bc9122743e88/extensions/ipy-aspell-server.py), which is the one auto-enabled by #476, rather than [nbextensions/usability/aspell/ipy-aspell-server.py](https://github.com/ipython-contrib/IPython-notebook-extensions/blob/b508b0ffbda62ac935088f57b579bc9122743e88/nbextensions/usability/aspell/ipy-aspell-server.py). I think the intention was in the reorganization to move the extension to the `extensions/` folder, but somewhere in the move, the original was left in place in addition to the new copy. I think we should probably remove the version at `nbextensions/usability/aspell/ipy-aspell-server.py`?

juhasch:
Python 2.7.11 at least includes the print function.
Should we revert #476 ?

jcb91:
Oh, really? My 2.7.11 still seems to function like a print statement, so I figured adding the `__future__` is simple and shouldn't break anything for anyone with a print function already, anyway. We should probably revert #476 at least until we've fixed things so that everything that's left in `extensions/` can be loaded without error. I think it might be more reasonable to just add to an explicit list of enabled extensions, rather than blithely enbling any `.py` file in the directory?

juhasch:
Sounds sensible. The print statement should be fixed by #498.

juhasch:
PR #499 reverts #476 for now

JanSchulz:
Re print statement vs. function: in this case it work without the future import, as `print("string")` is interpreted (by py27) as `print  ("string")` which is `print "string"` as the `(...)` is unrelevant. would be differnet for `print("s", "s")` as there the `(...)` results in a tuple...

-> Losing as this seems to be fixed in #498, which is merged :-)

PR #499, opened 2016-02-25
Revert "Configure to load all server extensions apart from `nbextensions`."

juhasch:
Reverts ipython-contrib/IPython-notebook-extensions#476 as it tries to load all kind of Python scripts, not only server extensions.


jfbercher:
I agree with that.

Issue #485, opened 2016-02-22
Selecting heading cell and all children


Kevin-McIsaac:
Now that notebooks support copy and past of a set of cells, it would be very useful to be able to select the heading cell and all its children with a single command. This would then the heading and all its children to be easily move. Indeed, I suspect this is more common that wanting to select just the heading. 

A few options:
1. Change the default so that selecting a heading select all the heading and all children.
2. Make shift-select of a heading select all children.
3. Single click of a heading bracket, anywhere, selects the heading and all children.

jcb91:
Yep, seems like a good plan. As for options:

1. I think making it default might be a bit confusing/frustrating. But it would, I think, make sense to do this for _collapsed_ headings, as it seems a little more intuitive. This can be done probably quite easily, but it will mean that having a heading as the current cell will also highlight it as part of a multi-selection. I think I'd prefer new bindings
2. Shift-select is already bound, extending selection, so I don't think we ought to repurpose it. Maybe shift-left and shift right could be added as extra key bindings, to select/unselect all cells in the section?
3. Yep, brackets are an easy, intuitive option, so I can add that in addition to anything else

Thoughts?

kkumer:
Also, note that in Mathematica if you select all cells via bracket, shift-Enter will execute all those cells. This is quite useful - I always have some sort of initialization section in my notebooks.

jcb91:
@kkumer yep, once all the cells are selected, the usual ctrl-enter shortcut will queue all of them for execution.

jcb91:
So, a preliminary solution is available at my branch https://github.com/jcb91/IPython-notebook-extensions/compare/master...usability/collapsible-headings. So far, this includes work to:

1. Add an extra keyboard shortcut (default shift-right) to select the heading's whole section.
2. Make the Mathematica-style brackets single-clickable to select the whole section
3. Decide whether selecting a whole section also reveals it (which was the default, and a bit annoying, so I made it configurable)
4. fixes an off-by-1 bug in `reveal_cell_by_index`
5. fixes the right-alignment of the Mathematica brackets, which I'd accidentally broken somewhere along the line
6. fixes toggle icon margins inside buttons, which I'd broken (insufficient testing, again, d'oh!)

As always, any thoughts welcome.

jcb91:
So, a few more commits pushed to my branch, which should (hopefully):

 * Fix persistence and icon-showing issues seen by @kkumer in #482
[...]

Issue #482, opened 2016-02-20
Persistence of collapsed state with collapsible_headings


kkumer:
(This follows the discussion in #473 - it doesn't make sense to hijack another extension's issue.)

Sorry  for the delayed answer but I was quite busy. And during that I actually used collapsible_headings a lot and it's working great for me. So, what we are discussing here are just minor things. It turns out that my two problems: (1) Icons left of headings not shown (but I can collapse by clicking the right place) and (2) collapsed state not persistent over reloading the notebook (BTW, I was careful to save first) are happening only on my home computer (Chrome on Windows 10)  accessing remote jupyter server. When working localy at jupyter server machine (Chromium on Linux), everything is perfect. Icons are there and things are persistent.

Your new updates didn't help with persistence (when working from home), and helped partially with icons. Now only some are missing, some of the time. Note that  I can actually always bring icons into existence by executing the heading cell.
(BTW, I like the new triangle icons much better than those huge plusses.)

Here are javascript console errors:

`Font from origin 'null' has been blocked from loading by Cross-Origin Resource Sharing policy: Invalid response. Origin 'https://[MY SERVER HOSTNAME]:8888' is therefore not allowed access.`

`chrome-extension://laankejkbhbdhmipfmgcngdelahlfoji/core/vendor/jquery.min.js:2 Uncaught SecurityError: Failed to read the 'contentDocument' property from 'HTMLIFrameElement': Blocked a frame with origin "https://[MY SERVER HOSTNAME]:8888" from accessing a frame with origin "chrome-extension://oknpjjbmpnndlpmnhmekjpocelpnlfdi".  The frame requesting access has a protocol of "https", the frame being accessed has a protocol of "chrome-extension". Protocols must match.`

jcb91:
OK, that's... confusing. Glad to hear it's sort of working, but I'm so far a bit stumped by the remote/local difference. So, just to ensure I've understood correctly, when you're working from the home machine, you're opening a browser on the home machine, and just entering the server hostname instead of the more common `localhost` right? So if that's the case, the server-side of the local and remote setups is exactly the same. 

The `Font from origin 'null' has been blocked from loading by Cross-Origin Resource Sharing` error is something I see in my console all the time, and it seems to be a normal part of MathJax loading (i.e. not actually an error, see https://github.com/mathjax/MathJax/issues/926#issuecomment-127452031).

[...]

PR #506, opened 2016-02-29
Updates to `usability/collapsible_headings`:

jcb91:
New features:
 * Add an extra keyboard shortcut to select all cells in the selected heading's section
 * Add support for single-clicking Mathematica-style brackets to select the whole of the bracketed section
 * Add a configurable flag so that selecting a whole heading section doesn't also expand it
 * Make the collapse/uncollapse shortcuts also select the next/previous heading if one isn't currently selected
 * make bracket width configurable
 * add a hover state for brackets, as suggested by @Kevin-McIsaac

improvements:
 * alter css to give brackets mitred corners
 * reduce use of patches, relying instead on the `delete.Cell` and `rendered.MarkdownCell` events.
 * Add comments to explain why we still have to patch undelete and select
 * use `Jupyter.notebook.find_cell_index` to find cell indices, seems more future-proof
 * update readme

fixes:
 * bugfix in yaml
 * fix margins for button-style toggles, broken in 70473e6
 * restore accidentally-broken right-hand alignment of Mathematica-style brackets
 * fix persistence and icon-showing issues in github issue #482, which seem to be a result of variability in when callbacks are executed
 * bugfix reveal_cell_by_index, which wasn't opening level 6 headings

### DIFF
--- a/nbextensions/usability/collapsible_headings/collapsible_headings.yaml
+++ b/nbextensions/usability/collapsible_headings/collapsible_headings.yaml
@@ -52,3 +52,7 @@ Parameters:
   description: Command-mode shortcut to uncollapse (expand) the selected heading cell
   input_type: hotkey
   default: right
+- name: collapsible_headings_shortcut_select
+  description: Command-mode shortcut to select all cells in the selected heading cell's section
+  input_type: hotkey
+  default: shift-right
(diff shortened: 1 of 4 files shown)


juhasch:
Looks very good!
I'll merge it and make a PR to remove the old `testing/hierarchical_collapse` extension.

jcb91:
fixes #485, #482

Issue #481, opened 2016-02-20
Replace testing/hierarchical_collapse with usability/collapsible_headings/


Kevin-McIsaac:
I've been using  usability/collapsible_headings/ in my work for the last couple of week and I find it more stable that testing/hierarchical_collapse.  With the new features merged today, this now looks as good. Finally, the dev is very responsive.

I propose that the old  testing/hierarchical_collapse be deprecated and replaced by  usability/collapsible_headings/

PR #507, opened 2016-03-01
Remove old extension, replaced by collapsible_headings

juhasch:
The templates and nbconvert preprocessor need to be converted to the new collapsible_headings extensions.


jcb91:
Yeah, I haven't addressed those so far, mainly because I don't currently use nbconvert, so don't have much idea how it works. They also seem a little prototypical, so I guess they could do with an update. I'll have a look at some examples & give it a try.

jcb91:
Addresses #481

juhasch:
For a nbconvert template, see #510

PR #511, opened 2016-03-02
Updates to usability/collapsible headings:

### LINKED ISSUE #509: Collapsible headings conflicts with display of docstring (Shift + Tab)
The notebook extensions and the GUI are great! What I noticed today is that the extension "collapsible headings" leads to weird behaviour with Shift+Tab (dropdown with docstring of function): it does not display, or only displays at the head of the notebook.

jcb91:
* add preprocessor and template for nbconvert use, adapted from the old `hierarchical_collapse` ones, as in #510, #507
* Fixes #509 via a patch to the `Tooltip` class


juhasch:
Thanks. Let's merge and work out how well nbconvert handles this in reality.

PR #513, opened 2016-03-06
Tree filter extension. 

juhasch:
Allows filtering what is displayed in the notebook tree. Based on https://github.com/jdfreder/jupyter-tree-filter.git

As this extension is only active in the tree view, it is enabled in the `tree` and not the `notebook` config. This needs to be addressed for it to work in the nbextensions configuration page.
Should we create a new page for tree extensions or merge notebook and tree extensions ?


jcb91:
Nifty! I think a whole extra page would be overkill for the moment, given we've so far only one non-notebook extension. perhaps we could add a new (optional) key to the `yaml` files, something like `Views`, which would be a list of the views (and therefore config files) for which the extension can be enabled?

juhasch:
Yes, I had a similar idea. Unfortunately the config is used in many places in `config/main.js` and for parameters there is no direct link between the extension and it's parameters, as far as I could see.
Any ideas ?

jcb91:
Well, there are a few usages in `config/main.js`, but I don't think too many to update. There is currently no direct link between an extension and where its parameters might be stored (config just writes to the key names defined in the `yaml` file), so I was thinking of just writing the config keys to every config that the extension specifies (defaulting to just the regular `notebook` section). The other thing I've been thinking about for a while is grouping all of an extension's config keys under a single dict per extension, so that instead of

```json
{
  "collapsible_headings_make_toggle_controls_buttons": false,
  "collapsible_headings_select_reveals": false, 
  "collapsible_headings_show_ellipsis": false, 
  "collapsible_headings_show_section_brackets": true, 
  "collapsible_headings_size_toggle_controls_by_level": true, 
  "collapsible_headings_toggle_closed_icon": "fa-caret-right", 
  "collapsible_headings_toggle_color": "#aaaaaa", 
  "collapsible_headings_toggle_open_icon": "fa-caret-down", 
  "collapsible_headings_use_toggle_controls": true,
  "load_extensions": {
    "usability/chrome-clipboard/main": true, 
    "usability/collapsible_headings/main": true
  }
}
```

we could have something more like

```json
{
  "collapsible_headings": {
    "make_toggle_controls_buttons": false,
    "select_reveals": false, 
    "show_ellipsis": false, 
    "show_section_brackets": true, 
    "size_toggle_controls_by_level": true, 
    "toggle_closed_icon": "fa-caret-right", 
    "toggle_color": "#aaaaaa", 
    "toggle_open_icon": "fa-caret-down", 
    "use_toggle_controls": true
  },
  "load_extensions": {
    "usability/chrome-clipboard/main": true, 
    "usability/collapsible_headings/main": true
  }
}
```

I've wanted to use this structure in a couple of extensions, but haven't been able to due to the current shortcomings of the config extension, so perhaps it's worth fixing it...

jcb91:
A thought:
It might make sense to support dot-notation in parameter name definitionsthen we could define parameters in `yaml` as
```yaml
Parameters:
- name: collapsible_headings.add_button
  ...
- name: collapsible_headings.use_toggle_controls
  ...
- name: collapsible_headings.toggle_color
  ...
...
```

In fact, this could also be used to specify which sections to write config keys to, as in

```yaml
Parameters:
- name: notebook.collapsible_headings.add_button
  ...
- name: notebook.collapsible_headings.use_toggle_controls
  ...
- name: notebook.collapsible_headings.toggle_color
  ...
...
```

Any thoughts?

PR #517, opened 2016-03-06
Only toggle collapse when clicking on icon directly

juhasch:
Original click area:
heading_1
New click area:
heading_2

### DIFF
--- a/nbextensions/usability/collapsible_headings/main.js
+++ b/nbextensions/usability/collapsible_headings/main.js
@@ -155,8 +155,9 @@ define([
 						.addClass('collapsible_headings_toggle')
 						.css('color', params.collapsible_headings_toggle_color)
 						.append('<div><i class="fa fa-fw"></i></div>')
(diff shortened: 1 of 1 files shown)


jcb91:
Although I see what you're getting at here, and it is a good plan, it only makes sense if the option to make the icons into buttons _isn't_ selected (as the whole button should be clickable).

juhasch:
Sure. If I understand your code right, the `on('click')` function is only added when the icon is not a button.

jcb91:
No, I think it's slightly different: the `on('click')` function is registered for all toggle controls (buttons or otherwise), and the `collapsible_headings_make_toggle_controls_buttons` config value just controls whether the `btn` and `btn-default` css classes are added...

jcb91:
Something like jcb91@81973f55cce078718eb26fd7cabb40d93e9fb434 ought to do it...

juhasch:
Thanks. I'll take a look.

PR #520, opened 2016-03-09
Updates to config extension

jcb91:
Including:

* support use of `tree`, `common` and `edit` config sections, as well as the default `notebook`, used up to now. This was prompted by (needed in) #513
* support dot-notation in parameter names for nested objects, e.g. `ruler_column` and `ruler_color` keys can be changed to `ruler.color` && `ruler.column`, giving a json structure like
```json
{
  "ruler": {
    "color": "#aaaaaa",
    "column": 79
  }
}
```
* various source cleanup things, including:
  * use named functions rather than assigning anonymous functions to variables, where possible
  * avoid declaring functions inside other functions
  * rename IPython -> Jupyter
  * make greater use of utils url split/join functions
  * use object notation where possible
  * move columns from js into template
  * make the extension fit the module pattern better, by not loading things automatically
  * reduce use of global variables
  * treat extensions as more object-like, making less use of their id, adding `full_url` and `ui` properties

### DIFF
--- a/extensions/nbextensions.py
+++ b/extensions/nbextensions.py
@@ -57,8 +57,7 @@ class NBExtensionHandler(IPythonHandler):
 
             if any(key not in extension for key in required_keys):
                 continue
-            if not extension['Type'].strip().startswith(
-                    'IPython Notebook Extension'):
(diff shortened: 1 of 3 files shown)


juhasch:
Looks good to me.

PR #524, opened 2016-03-10
Add a table beautifier extension

JanSchulz:
Based on the comments in
https://github.com/jupyter/notebook/issues/1182#issuecomment-194754977

Closes: https://github.com/ipython-contrib/IPython-notebook-extensions/issues/519


JanSchulz:
Unfortunately, it doesn't work: the `load_jupyter_extension` function is never called?

PR #525, opened 2016-03-10
fix conflicting config following install…

### LINKED ISSUE #523: template path set twice
I get the following error in the log after a reinstall of the extensions

```
[W 11:20:01.819 NotebookApp] Collisions detected in jupyter_notebook_config.py and jupyter_notebook_config.json config files. jupyter_notebook_config.json has higher priority: {
      "Exporter": {
        "template_path": "['.', 'C:\\\\Users\\\\jschulz\\\\AppData\\\\Roaming\\\\jupyter\\\\templates'] ignored, using ['C:\\\\Users\\\\jschulz\\\\AppData\\\\Roaming\\\\jupyter\\\\templates']"
      }
    }
```

I removed one (in the json) before the install and it was there again after the install, so it seem that the extension installer is to be blamed...

jcb91:
…
fixes #523
don't configure `Exporter.template_path` in python config, as it's already done in json, and doing both causes conflicts

### DIFF
--- a/configure_nbextensions.py
+++ b/configure_nbextensions.py
@@ -43,7 +43,7 @@ def update_config(config_file):
         print("Configuring %s" % config_file)
     make_backup(config_file)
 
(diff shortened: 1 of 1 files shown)


juhasch:
Thanks. Don't know why this survived.

PR #526, opened 2016-03-10
Fix typo

juhasch:


jcb91:
good spot, I forgot to restart my server to test this, d'oh!

jcb91:
yep, tests would be a nice plan. If I ever get round to looking at travis, docker, or similar, I'll set some up haha :laughing:

JanSchulz:
travis is IMO easy, i have no clue how to write tests for javascript UI Code :-(

juhasch:
I looked into using chromedriver some time ago to make Javascript tests:
https://github.com/juhasch/IPython-notebook-extensions/tree/feature/tests/tests

It would be some work to do this all nb extensions. But maybe it would be a good start to see if they load at all. Also, testing the Python nbconvert filters should not be so difficult.I believe I had something some time ago.

Travis sounds good, once there are at least a few tests.

juhasch:
Btw. @jcb91: I think it would be a good idea to have you in the ipython-contrib team.
@JanSchulz: What do you think ?

JanSchulz:
@juhasch +1

jcb91:
:smile: thanks! @juhasch yeah, doing everything through chromedriver might be a bit labour-intensive! I think the jupyter-notebook repo uses PhantomJS (see https://github.com/jupyter/notebook/tree/master/notebook/tests), which seems like it might be worth piggybacking on - it looks like their testing script is pretty long, but maybe we might do something similar largely by copy-and-paste (!) :laughing:

PR #530, opened 2016-03-11
add spellchecker extension

jcb91:


JanSchulz:
Yay!

nbextensions/usability/spellchecker/README.md
```diff
@@ -0,0 +1,39 @@
+This extension provides a CodeMirror overlay mode to highlight incorrectly
+spelled words in Markdown and Raw cells:
+
+![screenshot.png](./screenshot.png)
+
+It was inspired in part by
+[NextStepWebs/codemirror-spell-checker](https://github.com/NextStepWebs/codemirror-spell-checker/blob/78773ebdd6c8cf8acd043342023636ae345ca0f3/src/js/spell-checker.js)
+at the
+[suggestion](https://github.com/ipython-contrib/IPython-notebook-extensions/issues/521)
+of [@JanSchulz](https://github.com/JanSchulz).
```
JanSchulz: Thanks but IMO there is no need for that (-> `s/at the .*//`). :-)

nbextensions/usability/spellchecker/README.md
```diff
@@ -0,0 +1,39 @@
...
+javascript-based spellchecker.
+
+
+Dictionaries
+------------
+
+The dictionaries used by the extension are fetched according to the parameters.
+The urls can be relative (for files on your jupyter server) or absolute (for
+files hosted elsewhere, e.g. on a cdn like the defaults).
+I ([@jcb91](https://github.com/jcb91)) think that typo.js _can_ also load
```
JanSchulz: Is there any way to load them from the local filesystem? E.g. "Download two files from ... and put them into ... as `de_DE.....`and then specify the language?

JanSchulz:
Tried it, it works: unfortunately no replacement suggestions :-(

@jcb91 Have you tried the other way with codemirror type `"contenteditable"` and then add a `spellcheck=true` to each span?

-> https://github.com/codemirror/CodeMirror/issues/1017#issuecomment-52249542

nbextensions/usability/spellchecker/README.md
jcb91: Yes, absolutely. You can:

1. Put them in some directory in `/nbextensions`, and then use the configuration page to specify their urls as relative ones (relative to the spellcheck extension script),
2. Alternatively, `Typo.js` already supports multiple dictionaries, I think, so probably all you have to do is to put them into `typo/dictionaries/de_DE/`, then on the configuration, set the `dic`/`aff` URLs to blank, to get typo to try to use its own. I haven't tried doing that yet though...

nbextensions/usability/spellchecker/README.md
jcb91: Ah, you're welcome. Besides, I like to keep a history of where ideas come from, sometimes it's handy to remind yourself when you want to come back to something after not looking at it for some time!

jcb91:
Yeah, no replacement suggestions as yet, although since Typo.js provides suggestions, I expect they could be added as a popover or something fairly easily. I didn't bother trying the `contenteditable` method, since as huangp notes in https://github.com/codemirror/CodeMirror/issues/1017#issuecomment-52438444, it doesn't trigger correct events for CodeMirror to update the text when replacements are used, which could lead to a whole bunch of weird inconsistent-state problems. Maybe I'll give it a try and see how horrible it is, though :stuck_out_tongue:

nbextensions/usability/spellchecker/README.md
JanSchulz: Can you add this to the help? Also how to name the path (i actually have no idea how the URL is for the things in my nbextension dir)

JanSchulz:
re contenedible: from the above issue, I got the feeling that in 2014 they didn't want to include a contentedible part (["Okay, yeah, that uses contentEditable, which is not going to happen in CodeMirror."](https://github.com/codemirror/CodeMirror/issues/1017#issuecomment-52287597)) but the documentation on this was [included in 2015](https://github.com/codemirror/CodeMirror/commit/c1e30b8cc3ab741912f33b981ce00fbee15130cd) and includes the wording 

> the <code>"contenteditable"</code> model. The intention is to make it the default on modern  desktop browsers in the future.</dd>

jcb91:
True, they do seem to use `contentEditable` at least on mobile, so it may be worth a shot. The fact that they have explicitly disabled spellcheck makes me suspicious that the events won't cooperate, but whatever, we can give it a try, at least...

nbextensions/usability/spellchecker/README.md
jcb91: will do...

JanSchulz:
Re disable spellcheck: this is also used on the "other version" and there it is used to disable the spellcheck on the (hidden?) textarea... So another explanation could be that they disabled it in the "regular" version and just used the same thing in the to-be-new version...

jcb91:
OK, I've added a more complete explanation to the readme, including some python sample code, and actually tried switching to other dictionaries. It seems to work ok for me.

You've lost me a bit with the "other version" - I'm not quite sure what you mean?

jcb91:
Unfortunately, it seems that the "contenteditable" model referred to in the docs doesn't actually set the contentEditable attribute at all...

JanSchulz:
Sorry: "other version" = whatever is not the "contenteditable" mode.

Actually I really thought, that setting the "contenteditable" mode in codemirror would use something in the backend which sets "contenteditable to the used html element. So it seems that I misled you :-( Sorry again!

jcb91:
A few fixes, the most prominent being that I hadn't got raw downloads of the dictionaries - I managed to get round this by using the base64-encoded versions which are provided.

janschulz:
Just to let you know: I'm using this now for a few days and I'm loving it. It would be perfect with suggestions, but this is great as it is :-)

jcb91:
okey doke, I'll merge as-is then, and maybe add suggestions later on...

PR #533, opened 2016-03-14
Unbreak the table beautifier

janschulz:
The beautifier worked, but it seems that the commands after the beautifier were not executed anymore, as the ToC didn't update afterwards anymore...

### DIFF
--- a/nbextensions/styling/table_beautifier/main.js
+++ b/nbextensions/styling/table_beautifier/main.js
@@ -21,7 +21,6 @@ define([
 ) {
 
   function bootstrapify_all (){
-    debugger;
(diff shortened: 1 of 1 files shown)


JanSchulz:
The notebook problem is in https://github.com/jupyter/notebook/issues/1207

janschulz:
Can someon please test and then merge? This is slightly embarrassing that this slipped in :-(

jcb91:
Looks ok to me. Could we remove the `debugger;` calls, as they're a bit annoying if you're trying to debug a different extension while this one is enabled :stuck_out_tongue:

janschulz:
@jcb91 removed...

PR #542, opened 2016-03-17
remove unused wiki-images

jcb91:
There are no longer any wiki pages which refer to the repo for their images, so we can safely remove these.


juhasch:
:+1:

PR #543, opened 2016-03-17
remove old, unused configuration scripts

jcb91:
None of these scripts are actually used in any of our recommended installation methods anymore, so I think we should get rid of them, in favour of promoting pip/setup.py


juhasch:
:+1:

Issue #505, opened 2016-02-29
How can I uninstall all extensions?


cqcn1991:
So I was experimenting with nbextension, and somehow, I don't remember what I've done, I just have an error in my browsers console

`Failed to load extension: ["nbextensions/usability/my_toc/toc_test"] `(One of my test files)

The problem is that I already delete that dir and diabled the extention. I tried to delete the `nbextension` dir and re-installed the `anaconda`. But the error is still there.

So I'm thinking, what's the correct way to erasing all nbextention files?

jcb91:
So, the problem here is that the server doesn't load extensions based on the existence of files, but has a config file that tells it what extensions to load. In your current version of that config file, it's asked to load `usability/my_toc/toc_test`, but, since you've deleted the file, it fails to load it. Unfortunately, the `nbextensions` server extension only knows about extensions that have a `yaml` descriptor file. This means that if you have deleted the yaml descriptor, or altered it so that it refers to a different extension path , you can no longer use the `nbextensions` page to disable it. You can fix this quite easily from a python prompt, I'll get back to you in a minute with a solution...

cqcn1991:
@jcb91 exactly, I think I've altered somewhere, maybe in a yaml file. But just couldn't find it. (I don't know the files' location, so I don't know where to perform a full-text search).

jcb91:
So, this has a happened to me on a number of occasions, where I've had an extension enabled, then renamed or moved it, such that its path changes, without disabling it at the old path first. This means the server still tries to load the old extension path, which, of course, no longer exists. You can alter the config so the server won't load it anymore with

```python
from notebook.services.config import ConfigManager
path_to_disable = 'usability/my_toc/toc_test'
cm = ConfigManager()
cm.update('notebook', {'load_extensions': {path_to_disable: True}})
```

cqcn1991:
@jcb91 just tried it. It works. That's really impressive.

jcb91:
@cqcn1991 glad to hear it worked for you. If you want to find the config file itself (it's also the one used for lots of nbextensions' config settings), you can use

```python
from notebook.services.config import ConfigManager
cm = ConfigManager()
print(cm.file_name('notebook'))
```

which for me (on OSX) gives `/Users/josh/.jupyter/nbconfig/notebook.json`

juhasch:
Can we close this ?

cqcn1991:
@juhasch done.

ruxi:
[...]

PR #544, opened 2016-03-17
Updates to config extension

jcb91:
Function:
* add entries for nbextensions which are enabled, but not associated with a yaml file. This should make fixing problems like #505 a little easier
* relax requirements for nbextension yaml descriptor files, removing absolute requirement for `Compatibility` and `Description` keys
* Use `window.history` instead of hashes to record which extension is selected.

UI
* Only have two columns when there is an icon to put into the right one. Otherwise, it'd just be empty anyway
* Make parameters full-width, rather than sitting in the left column

Code cleanup:
* treat extensions as more object-like
* remove redundant css classes, exception-catching
* some style-guide fixes
* Don't export functions which aren't used outside this module.

### DIFF
--- a/extensions/nbextensions.py
+++ b/extensions/nbextensions.py
@@ -36,15 +36,12 @@ class NBExtensionHandler(IPythonHandler):
 
         # Build a list of extensions from YAML file description
         # containing at least the following entries:
-        #   Type          - identifier
-        #   Compatibility - compatible notebook version, e.g. '4.x'
-        #   Name          - unique name of the extension
-        #   Description   - short explanation of the extension
-        #   Main          - main file that is loaded, typically 'main.js'
+        #   Type - identifier, must be either 'IPython Notebook Extension' or 'Jupyter Notebook Extension'
+        #   Name - unique name of the extension
+        #   Main - relative path to main file that is loaded, typically 'main.js'
         #
         extension_list = []
-        required_keys = (
-            'Type', 'Compatibility', 'Name', 'Main', 'Description')
+        required_keys = ('Type', 'Name', 'Main')
 
         for ext_dir, yaml_filename in sorted(yaml_list):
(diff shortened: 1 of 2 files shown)


nbextensions/config/main.js
```diff
@@ -126,10 +115,10 @@ define([
      */
     function set_config_active (extension, state) {
         state = state === undefined ? true : state;
-        console.log('nbext', state ? ' enable:' : 'disable:' , extension.Name );
+        console.log('nbext', state ? ' enable:' : 'disable:', extension.Name);
```
juhasch: Could you make the message a little more self-explaining for the user ?

extensions/nbextensions.py
```diff
@@ -36,15 +36,12 @@ def get(self):
...
-        #   Description   - short explanation of the extension
-        #   Main          - main file that is loaded, typically 'main.js'
+        #   Type - identifier, must be either 'IPython Notebook Extension' or 'Jupyter Notebook Extension'
+        #   Name - unique name of the extension
+        #   Main - relative path to main file that is loaded, typically 'main.js'
         #
         extension_list = []
-        required_keys = (
-            'Type', 'Compatibility', 'Name', 'Main', 'Description')
+        required_keys = ('Type', 'Name', 'Main')
```
juhasch: This is a good idea, I think.

jcb91: Actually, to be honest, we could even dispense with requiring `Name`, using the URL of the main JavaScript file as the default

nbextensions/config/main.js
jcb91: Certainly, did you have anything in mind? Something like `"nbextension {name} set {not} to load for every notebook "`?

juhasch: I think just something like `Notebook extension <name> enabled/disabled` would be enough.

PR #545, opened 2016-03-17
bugfix non-matching default parameters in `collapsible_headings`…

jcb91:
… js and yaml files disagreed about defaults, leading to confusing behaviour on config page. Also, it seems that config colors can't be specified as 3-char hex codes, but need to be 6 in order to set html5 input correctly.


jcb91:
whoops, didn't mean to make a branch in the main repo! Will re-open form my fork...

PR #548, opened 2016-03-17
[Freeze] Initial commit

kukanya:
I wrote an extension which allows to make cells read-only (can be executed, cannot be altered) and frozen (cannot be either altered or executed). I'd like to make it accessible and easy to install for everyone interested. Thanks


jcb91:
Looks nice, plays with runtools well, thanks for the submission!

nbextensions/usability/freeze/main.js
```diff
@@ -0,0 +1,126 @@
...
+				cell.metadata.run_control = {};
+			if (cell.metadata.run_control.frozen == undefined)
+				cell.metadata.run_control.frozen = false;
+			if (cell.metadata.run_control.read_only === undefined)
+				cell.metadata.run_control.read_only = false;
+			if (state == undefined)
+				state = 'normal';			
+			switch(state) {
+				case 'normal':
+					cell.metadata.run_control.frozen = false;
```
jcb91: rather than setting values to `false`, it can be nicer to just remove them

jcb91:
The main issue with this so far is is lack of a return value and `load_ipython_extension` method. The nbextension mechanism works by the notebook require-ing a javascript file which should contain a `define` call, which returns an object that has a key `load_ipython_extension`, which is a (parameter-less) function that the notebook will call in order to perform setup. If it can't find it, it'll show up in the console as an error...

nbextensions/usability/freeze/main.js
```diff
@@ -0,0 +1,126 @@
...
+	events,
+	codecell
+){
+	'use strict';
+	
+	var CodeCell = codecell.CodeCell;
+	
+	function patch_CodeCell_execute () {
+		console.log('[Freeze] patching CodeCell.prototype.execute')	
+		CodeCell.prototype.old_execute = CodeCell.prototype.execute;
```
jcb91: when patching objects, it's best not to attach your old values back onto the object - it can cause problems if another extension tries to patch the same method, and overwrites the `old_execute` value. Instead, it's safer use a variable local to either the `patch_CodeCell_execute` method, or the extension definition method.

nbextensions/usability/freeze/main.js
```diff
@@ -0,0 +1,126 @@
...
+		
+		CodeCell.prototype.execute = function () {
+			if (!this.metadata.run_control.frozen) {
+				this.old_execute(arguments);
```
jcb91: This doesn't do what you (presumably) intended. `this.old_execute(arguments);` will pass the arguments as a list. You want `this.old_execute.apply(arguments)` - see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/apply

kukanya:
Thanks for your input, @jcb91 
Hopefully everything is fixed now.

jcb91:
Thanks! Looks good for everything I mentioned, unfortunately, I missed a couple of tings (my fault, apologies). To whit:

1. it's probably nicer not to alter the background of 'normal' cells - you can remove the jquery-set css by just doing `css('background', '')`. This avoids trying to chase changing defaults, or mess up others' custom themes - my `div.input`s seem to have a bg color of '#f7f7f7', for instance.
2. For long notebooks, or connections to remote/laggy servers, the extension can get loaded _before_ the notebook finishes loading, in which case your call to your initialisation function (now called `initialize_states`) won't work correctly. You can get round this by checking whether the notebook has been defined, and if not, registering an event callback, as in the recent PR #550.

jcb91:
Further apologies, a bug introduced by my suggestion of not setting metadata where unnecessary, now the patched execute also needs to check

```javascript
if ((cell.metadata.runtools === undefined) || cell.metadata.runtools.frozen)
```

otherwise any attempted execute will throw an exception if runtools is undefined, and execution won't happen...

kukanya:
Thanks, I'll look into it.
I want to ask for your advice too.
When I first started developing this extension, I made it so it shared read-only cell state with already existing Read-Only extension (kind of backwards compatibility, I guess?) but it now introduces some nuances. In particular, patched `execute` actually needs to check 3 conditions
``` javascript
if (this.metadata.run_control === undefined ||
	this.metadata.run_control.frozen === undefined ||
	!this.metadata.run_control.frozen) 
```
as `run_control.frozen` property may still be undefined while `run_control` is defined.
Do you think I should stick with it or should I put my properties into different sub-structure?

jcb91:
Well actually, if `run_control.frozen` is undefined, it's equivalent to being false (since in javascript `Boolean(undefined) === false`). I think it's nice to keep them all in the same structure. You might find a [`jQuery.extend`](http://api.jquery.com/jquery.extend/) call to be an easier way to update the metadata status though - you can even set keys to `null` to remove them :smile: so to set a cell as readonly but not frozen you could use something like:

```javascript
var $ = require('jquery');
var new_runtools_values = {
    readonly: true,
    frozen: null, // null gets $.extend to remove the key from the object being updated
};
$.extend(cell.metadata.runtools, new_runtools_values);
```

kukanya:
Thanks a lot for explaining! It's actually the first time I use javascript (I just wanted that 'freeze' button really badly :laughing:) so I'm kind of learning its features on the run. Your comments are incredibly informative and helpful.
I will now dive into fixing the things you mentioned, and will reach out once I'm done. Thank you once again.

kukanya:
Setting keys to `null` doesn't remove them for me, so I'm still setting them to `false` instead.
Otherwise, seems like I've fixed everything you suggested.

jcb91:
sorry, I got confused with the notebook's `ConfigSection` class, which uses `null` as a shorthand for deletion, as you point out, `jquery.extend` doesn't. Looks good to me...

kukanya:
It's fine. Thanks a lot for your help!

PR #550, opened 2016-03-20
hide_input: Hide cells after notebook is fully loaded

amueller:
Otherwise cells might not be hidden properly if the notebook takes a long time to load. That might lead to very confusing inconsistencies between the metadata and the actual display status (clicking hide again hides the cell in the browser but toggles the metadata to ``hide_input=False``)

### DIFF
--- a/nbextensions/usability/hide_input/main.js
+++ b/nbextensions/usability/hide_input/main.js
@@ -5,22 +5,30 @@ define([
     'base/js/namespace'
 ], function(
     $,
-    IPython
+    Jupyter
 ) {
     "use strict";
 
     var toggle_selected_input = function () {
         // Find the selected cell
-        var cell = IPython.notebook.get_selected_cell();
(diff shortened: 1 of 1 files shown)


amueller:
hm it still doesn't seem to work everytime :-/

amueller:
works now after following advice from @minrk

juhasch:
Looks good to me. Thanks for looking into it.
Btw., have you tested the `check_input_prompt` filter to remove the prompts ?

amueller:
yeah but I wanted to remove the prompt without changing the template. Using a template means that if the user writes a custom template, they need to be aware of the plugin template to inherit from it. Using multiple plugins, this looks impossible.

juhasch:
Yes, I think this is an unfortunate restriction of the template system.
Should be easy to do in a few lines using a nbconvert preprocessor, too.

amueller:
yeah but that doesn't work for stripping the ``Out[X]`` part unfortunately

PR #553, opened 2016-03-22
Make hide_input extension work with nbconvert HTML export

juhasch:


juhasch:
I removed the Python file as it is only an (outdated) explanation on how to configure nbconvert an added the relevant part to the readme file.

Issue #551, opened 2016-03-21
TOC2 table of content links & numberings don't work once pages are nbconverted to HTML


hhayama:
TOC2 is great at generating the table and auto updating the numbering but once I convert the notebook to HTML, the links all become broken & the numbering from each individual section disappears from the nbconverted file.  Is this the intended behavior?  It unfortunately makes this extension not as useful for HTML converted files which forces me to go back to (more or less) manually updating and creating the links.

jfbercher:
Yes. I realized that too. This has been introduced recently when adding unique ids to cope with repeated headings. For the toc cell it works in the live notebook but not after conversion (where headings numeration is lost). I have made a small patch which returns to the previous behavior  for the toc cell, and I will upload it shortly.

hhayama:
Awesome news, Thanks so much!

jcb91:
@hhayama has #554 fixed this for you?

brunobeltran:
Would a cleaner solution be to do what the Python-Markdown extension does and implement a preprocessor in python?

Then we could get rid of the duplicate code and confusion that the `saveid`/`id` stuff causes?

jfbercher:
The original issue is solved by #554. 
I think that the last comment by @brunobeltran is solved, see the discussion in https://github.com/jcb91/jupyter_contrib_nbextensions/issues/3. Otherwise please reopen an issue.

PR #554, opened 2016-03-22
Change links in toc_cell to original ids

jfbercher:
This is a small modification to address part of #551 

The links in toc_cell, the cell with toc at the bottom of the notebook were incorrect since they link to anchors constructed as the original id+number of section, in order to get a unique id. But since these anchors only exist in the live notebook, they do not survive to nbconvert conversion. At this time, the best I can see is to link to the original ids instead for the toc_cell -- but with the problem that we may have same anchors for sections with the same heading. The unique anchors are preserved within the toc window itself. 

----
I will investigate possible solutions for preserving/restoring section numberings + unique ids in the toc_cell. This shall involve writing a preprocessor (I think that invoking a js file through nodejs could be a way to do it).. Will investigate, but later...

### DIFF
--- a/nbextensions/usability/toc2/main.js
+++ b/nbextensions/usability/toc2/main.js
@@ -66,6 +66,8 @@ define(["require", "jquery", "base/js/namespace",  'services/config',
       return ary.slice(0, h_idx+1);
   }
 
+
+
   var make_link = function (h, num_lbl) {
     var a = $("<a/>");
     a.attr("href", '#' + h.attr('id'));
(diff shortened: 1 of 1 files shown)


janschulz:
> through nodejs 

Jikes: please no additional dependencies...

jcb91:
I was thinking, perhaps the links in the ToC cell could take advantage of the unique-ified ids through a toc-added jquery click callback using a data attribute? That way, you can keep the original id as the href, but not lose the benefit of the unique ids in notebook usage...

jcb91:
well, this seems simple enough for now, so I guess we should merge, and think about future solutions somewhere else...

Issue #563, opened 2016-04-02
Automatic code hinting, without hitting tab


rkpatel33:
Is it possible to turn on autocomplete so that the suggestion list is displayed on every letter key press in edit mode?

This would bring the typing experience closer to that of Atom or Sublime code complete and would make switching back and forth just a bit easier.

Would love to see this if it does not exist, or please let me know if someone knows an easy way to hack it in `custom.js`.

jcb91:
I'm not aware of an existing implementation, but this does sound like a nifty idea, and should certainly be possible. I'll have a quick think.

rkpatel33:
Cool. I had suggested in the main Jupyter repo issues to just connect the even attached to <tab> in edit mode to and edit mode keypress of [A-Zaz], but the fear was it was gum up the front end and introduce latency.

I'n not familiar enough with the code to have a feel for it.

jcb91:
Indeed, it could end up making things a bit unresponsive, particularly where rounds trips to the kernel are slow (think kernels on remote machines, etc), so I can understand not wanting to implement it in core. This is in many ways a good thing for an extension, as people are free to enable/disable it as they choose...

jcb91:
ok, I've made a basic version (in #564). It turned out to be a little more complicated than I'd anticipated - the code completer in notebook javascript is doing some interruption of the codemirror events, and has a start/modify/stop model, so it takes a little more handling. I think it's passable now, but it relies a little on a timeout, currently set to 200ms, which may need adjusting. If you could let me know if/how well it suits you, that'd be great...

rkpatel33:
Awesome, I will give it shot. I installed Jupyter with the pip command. Is there a way I can switch to this branch, or do I need to pull the repo and run from there?

jcb91:
So, since this a notebook extension, installing it should be relatively independent of your jupyter installation. The easiest route is to install the (`IPython-notebook-extensions`) repo using its `setup.py` script, (which isn't yet a full setuptools installer, but will modify the necessary config files, and copy things into the correct directories for you). This should work via pip, with

```bash
pip install https://github.com/jcb91/IPython-notebook-extensions/archive/hint_constantly.zip
```

Some people have had trouble with using pip directly, if it doesn't work for you, you can clone the repo, switch to the correct branch, and run the setup script directly:

```bash
git clone https://github.com/jcb91/IPython-notebook-extensions.git jcb91
cd jcb91
git checkout hint_constantly
[...]

PR #564, opened 2016-04-03
add `hint_constantly` nbextension

jcb91:
addresses #563, enabling a code-completion dropdown all the time.


jfbercher:
Tried it: it works and is very nice. Well done @jcb91 !

It would be super nice if we could detect the opening parenthesis ( and show the function hint (signature) as it shows with shift-tab... 

[have a small issue that the extension does not show up in my /nbextension/ tab; installed manually instead - but it could be my own config which is a bit messed up]

janschulz:
> It would be super nice if we could detect the opening parenthesis ( and show the function hint (signature) as it shows with shift-tab...

this is https://github.com/jupyter/notebook/issues/1296 -> basically the hints are sorted "suboptimal", the parameter are already in there (look for "para=").

rkpatel33:
I just tried it and works great! So helpful.

One more small suggestion is to make the match filtering a fuzzy match instead of exact character sequence, so you don't have to type underscores etc.

(FYI -- I had to install manually as well as the addition did not show up in my /nbextension/ tab)

jcb91:
Thanks @jfbercher :smiley: 

So, it seems that something strange is happening (at least for two of you!) with it not appearing on the config page, but I'm at a loss so far to try to diagnose what that might be - it shows up fine for me. Are there any odd (server) log messages when loading `/nbextensions`? Things like failing to load the yaml file?

The shift-tab bit might be possible until core fixes the sorting, will investigate.

@rkpatel33, that fuzzy matching is, I think, something that'd have to be provided not by the individual kernels, which ultimately provide the suggestions lists to the notebook, so that's something I can't really fix here, although it might be something you could suggest to jupyter core to add a fuzzy-mode completion request to the existing regular one, so that kernels can implement it.

jfbercher:
@jcb91 No specific message in logs. For my part, replacing "Jupyter Notebook" by "IPython Notebook" in the Type keyword of the yaml file solves the nbextension tab issue. Does it means something for you?

jcb91:
Ah, yes, @jfbercher, that presumably means that the config extension is ignoring the yaml file as its `Type` key doesn't match the expected. #544 altered the config extension to allow `Jupyter Notebook Extension` in addition to `IPython Notebook Extension`, but it's possible that if you haven't restarted your jupyter server after installing the newer repo version, that you'll still be running with the older one, and hence having the config ignored...

jfbercher:
>  you'll still be running with the older one

Indeed @jcb91 . Had not refresh recently.

jcb91:
Added a few updates, including:

* rename (apologies for those for whom this causes complications, I just liked the new name :stuck_out_tongue_closed_eyes:
* a help-menu item to toggle the extra hinting off (in case it makes UI slow on laggy connections etc)
* an attempt to integrate shift-tab on opening parentheses. This isn't ideal, as it seems to supply the signature of the most recently referenced function, even if the current parenthesis is actually not a function call (e.g. it might be just a bracket, or part of a tuple).
* ignore more special keys - I found starting completion on pressing 'delete' was particularly strange

jfbercher:
Initial test. Works for me and is really nice! 
Thanks for implementing Shift-tab! Even if it is not ideal, it is a nice and useful functionality.

nbextensions/usability/hinterland/hinterland.js
```diff
@@ -69,8 +70,13 @@ define(function (require, exports, module) {
 						}, cur);
 
 						if (pre_cursor !== '' && !ignore_re.test(pre_cursor)) {
-							completer.startCompletion();
-							completer.autopick = false;
+							if (pre_cursor === '(') {
+								cell.tooltip.request(cell);
```
janschulz: [This is a comment without trying the new functionality...]
The right hints are in the completer, they are just buried under the global stuff. So if it would be possible to simple sort the completion tipps so that `xxxx=` (equal sign == parameter) are at the top, this would be even better (because you could then select the param)... IMO this could even be done in all cases: if the cursor is not in the context of a function call, then there will simple no such items...

jcb91: ah, gotcha! I hadn't really appreciated what you meant. Mm, although certainly possible, this would involve digging into and patching the jupyter internals of how the tooltips are displayed, so it may take a little longer...

janschulz: If you find it, please also fix upstream :-)  -> https://github.com/jupyter/notebook/issues/1296

PR #567, opened 2016-04-04
Fixed imported get_nbext_dir function name

ecogodi:
Updated to match the newest code in jupyter notebook master


jcb91:
Good plan, but we ought to wrap this in a try/except to work with both cases

PR #570, opened 2016-04-06
add sorting for tables...

jcb91:
a la #540. This PR uses js from cdnjs. I guess we could include a whole lib, but for now, to reduce bloat, I haven't bothered.

Issue #540, opened 2016-03-17
Table sorter option for table-beautifier


janschulz:
A sorter and a little button which does filter cells in the table could be interesting... E.g. like the stats table here: https://jiffyclub.github.io/snakeviz/#interpreting-results

janschulz:
Or this, which looks even better: https://mottie.github.io/tablesorter/docs/example-widget-bootstrap-theme.html

jcb91:
fixed by #571

PR #571, opened 2016-04-07
Feature/table beautifier sortable

jcb91:
replacement for #570 

@janschulz sorry about the errors, I ought to have gone with the route I've used previously for require-ing things which may or may not be available, I've made it more robust now I think...

### DIFF
--- a/src/jupyter_contrib_nbextensions/nbextensions/table_beautifier/main.css
+++ b/src/jupyter_contrib_nbextensions/nbextensions/table_beautifier/main.css
@@ -8,3 +8,27 @@ table.table {
     margin-left: 0;
     margin-right: 0;
 }
+/* stuff for tablesorter plugin */
+.tablesorter-default .header,
+.tablesorter-default .tablesorter-header {
+	background-image: url(data:image/gif;base64,R0lGODlhFQAJAIAAACMtMP///yH5BAEAAAEALAAAAAAVAAkAAAIXjI+AywnaYnhUMoqt3gZXPmVg94yJVQAAOw==);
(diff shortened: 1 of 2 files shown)


jcb91:
(to provide some of #540)

jcb91:
I'm not sure what the 'check' was about though, as it doesn't seem to be in the source.

janschulz:
Ok, this "works", but has teh side effect to remove the last column and replace all valueswithin that colum with `-`:


R Code (needs a master version of all libs):

```R
library("dplyr")
options(jupyter.log_level = 3)
options(jupyter.rich_display = TRUE)
data.frame(a=1:200, b=201:400, c=401:600)
```

janschulz:
When I reload the notebook without the extension activated, all values are there. When I reload the notebook with the extension activated, all values are there. Only when I run the cell will the last column be replaced with '-'

jcb91:
yeah, I think the bootstrap-table extension I tried to use relies on the presence of thead and tbody. That said, it doesn't seem to work correctly in all cases for me either, for different (presumably?!) reasons. I'll give a few other libraries a try instead...

jcb91:
OK, so Mottie's tablesorter seems to work ok on the R-generated table, despite the missing heading/th column.

The reloading being different to running was down to the load time of the external library, which meant that the notebook.loaded callback was firing before the library loaded. I've added an extra `bootstrapify_all` call on library load to fix that (hopefully!)

jcb91:
@juhasch any joy with the latest version?

jcb91:
(rebased)

PR #573, opened 2016-04-12
This extension adds the ability to split a cell vertically in a notebook.

cdknorow:
This shows the idea, in this extension you can currently only toggle the split cell using shift-s in command mode. I haven't added a button yet. The information is stored in the cell metadata. 

3f97874c-f93d-11e5-9f3c-7efe97d7dac3

### DIFF
--- a/nbextensions/usability/splitcell/readme.md
+++ b/nbextensions/usability/splitcell/readme.md
@@ -1,0 +1,3 @@
+## Split Cells for Jupyter/Ipython Notebooks
+
+Enter command mode (esc), use shift-s to toggle the current cell to either a split cell or full width.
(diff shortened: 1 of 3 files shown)


jcb91:
Looks nice! Could you use Jupyter rather than IPython for the namespace please?

For adding a toolbar button, it's probably worth making the switching an action so that it can also be used more generally (see, e.g. [nbextensions/usability/init_cell/main.js#L40-L50](https://github.com/ipython-contrib/IPython-notebook-extensions/blob/22dd78c42ec8dd1724120345b7472d595bfdb264/nbextensions/usability/init_cell/main.js#L40-L50) for an example).

cdknorow:
@jcb91 Thanks for the advice. I updated the name space and added a nice button. I messed up the commit update, ended up just creating a new pull requests instead.

Issue #585, opened 2016-04-19
Can't get Markdown to export rendered values


rfenner:
I installed the extensions using the setup.py file but When exporting some python markdown it doesn't render the markdown output.
I'm using jupyter 4.0.6 with python 2.7
I had to fix the install.py error where FileNotFoundError is not available in python 2.7. I fixed it by adding at the after imports
```
try:
  FileNotFoundError
except NameError:
  FileNotFoundError = IOError.
```

Beyond that it installed fine

My jupyter_nbconvert_config.json looks like
`{
  "Exporter": {
    "preprocessors": [
      "pre_codefolding.CodeFoldingPreprocessor",
      "pre_pymarkdown.PyMarkdownPreprocessor"
    ],
    "template_path": [
      ".",
      "/Users/rfenner/Library/Jupyter/templates"
    ]
  },
  "NbConvertApp": {
    "postprocessor_class": "post_embedhtml.EmbedPostProcessor"
  },
  "version": 1
}`

If I followed the wiki and added the line for the preprocessor .py file I'd see a message about a conflict when running jupyter nbconvert.

For some reason it just doesn't seem to be running the preprocessor.
I did add the path to preprocessor to the PYTHONPATH as well

juhasch:
Can you run nbconvert using the `--debug` option, i.e.
```
jupyter nbconvert test-markdown.ipynb --debug
```
and post the output ?

rfenner:
[NbConvertApp] Config changed:
[NbConvertApp] {'NbConvertApp': {'log_level': 10}}
[NbConvertApp] Searching [u'/Users/rfenner/Downloads/IPython-notebook-extensions-master', '/Users/rfenner/.jupyter', '/Users/rfenner/Library/Enthought/Canopy_64bit/User/etc/jupyter', '/usr/local/etc/jupyter', '/etc/jupyter'] for config files
[NbConvertApp] Attempting to load config file jupyter_config.py in path /etc/jupyter
[NbConvertApp] Attempting to load config file jupyter_config.json in path /etc/jupyter
[NbConvertApp] Attempting to load config file jupyter_config.py in path /usr/local/etc/jupyter
[NbConvertApp] Attempting to load config file jupyter_config.json in path /usr/local/etc/jupyter
[NbConvertApp] Attempting to load config file jupyter_config.py in path /Users/rfenner/Library/Enthought/Canopy_64bit/User/etc/jupyter
[...]

PR #586, opened 2016-04-20
FileNotFoundError is not available in Python 2

juhasch:
As noted in #585

### DIFF
--- a/install.py
+++ b/install.py
@@ -80,7 +80,7 @@ def uninstall_files():
                     try:
                         os.rmdir(src)
                     except OSError as ex:
-                        if ex.errno not in (errno.ENOTDIR, errno.ENOTEMPTY):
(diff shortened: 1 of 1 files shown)


jcb91:
Rather than remove it's handling completely, perhaps we could check its errno, as with the oserror, and as discussed in http://stackoverflow.com/questions/21367320/searching-for-equivalent-of-filenotfounderror-in-python-2 ?

juhasch:
Hmm, I guess `FileNotFoundError` is a subclass of OSError and corresponds to `errno ENOENT`.
https://docs.python.org/3/library/exceptions.html#FileNotFoundError

Maybe add this:
```
if ex.errno not in (errno.ENOTDIR, errno.ENOTEMPTY, errno.ENOENT):
```
 ?

jcb91:
Yep, sounds about right

juhasch:
Done. Thanks for the hint.

PR #587, opened 2016-04-20
Feature/svg2pdf

juhasch:
Add SVG to PDF preprocessor for nbconvert, so notebooks with svg images in markdown can be converted to PDF


juhasch:
As mentioned in https://github.com/jupyter/nbconvert/issues/244

PR #588, opened 2016-04-21
Add tests for preprocessors

juhasch:
Test
* pre_pymarkdown
* pre_codefolding
* pre_svg2pdf

Requires Inkscape installed for svg to pdf conversion.


jcb91:
Nice. I'll patch these into the nosetests I've started working on. It would probably make sense to rename to a python-compatible name (i.e. replace hyphen `test-postprocessors.py` -> `test_postprocessors.py`)

juhasch:
Sure, extract what you think you can use. For the `pre_svg2pdf` test I have to check if Inkscape is available, or otherwise skip the test.

PR #593, opened 2016-04-29
toc2: sidebar + html export

jfbercher:
##  toc2: new features: 
1. sidebar
2. html export
3. Navigation menu

### Sidebar

Triggered by @cqcn1991's discussion and work, cf [discussion here](https://github.com/ipython-contrib/IPython-notebook-extensions/issues/532),  a sidebar option is added. The floating toc window can be dragged and docked as a left sidebar. The sidebar can be dragged out as a floating window. These different states are stored and restored when reloading the notebook. 

demo

### html export
Add html export capability via templates toc.tpl and toc2.tpl. 
It is possible to export (most of) table of contents functionalities to html. The idea is to link a relevant part of the javascript
extension and the css, and add a small script in the html file. This is done using a template by
```
jupyter nbconvert FILE.ipynb --template toc
```
or 
```
jupyter nbconvert FILE.ipynb --template toc2
```
For the first template (toc), the files toc2.js and main.css (originally located in JUPYTER_DATA_DIR/nbextensions/usability/toc2) must reside in the same directory as intended for the html file. In the second template, these files are linked to the IPython-notebook-extensions github website. Export configuration (parameters) shall be edited directly in the template files (in templates directory JUPYTER_DATA_DIR/templates). 

A  **"Save as HTML (with toc)"** option is also available in the File menu and enable to directly convert the actual notebook. This option requires the IPython kernel and is not present with other kernels. I was not able to find a way to do it for other kernels (even with a customized exporter called from js).

###  Navigation menu

since I realized that it is not difficult to add menus elements, I had the idea of adding a navigation menu". This menu can live together with the floating window or sidebar. It can be enabled/disabled using the nbextension configuration utility (enabled by default).

demo2


janschulz:
This looks great!

jcb91:
Boss! This sounds fantastic :)

janschulz:
Unfortunately, here the sidebar feature doesn't work as advertised: it clings to the side, but is not displayed as a sidebar but still as a floating window:

sidebar

The feature is enabled (IMO that could be enabled as a default: I had to manually click it...)

jfbercher:
Stange. It seems that a style is missing. Have you downloaded/updated main.css? Can you have a look at the console output?

janschulz:
@jfbercher There isn't an error in the console. I use the table beautifier. I reinstalled the package via python setup.py install and I pressed a few times F5 in the opend notebook. Also, chrom tells me that the toc2/main.css file has `.float-wrapper {` in line 33, so it seems to be updated.

jfbercher:
It works for me in linux debian. I just ckecked with a fresh install on windows 8, anaconda, chome and it also works without any problem.
When dragging on the left side, toc should take the class ".sidebar-wrapper". Can you look at $('#toc-wrapper') in the console? Also can you start with an empty notebook?

juhasch:
Very nice, indeed!
I have a notebook where it works (i.e. the Jupyter example notebooks seem to work) and one that behaves like @janschulz describes.

jfbercher:
Can you @janschulz or @juhasch send me an example notebbok where the extension doesn't work? eg make a pull request at https://github.com/jfbercher/test and I will investigate. thxs

juhasch:
@jfbercher : See the link in my comment above

jfbercher:
I think I got it. 
It is a configuration issue (the toc config is copied from the config stored in notebooks, if it exists - in such a case, the key sidebar will not exist and yield this problem). I will update shortly.

jfbercher:
Changed reading of initial config. Hope this will fix. Thanks for the file @juhasch

juhasch:
Looks good, this fixes the docking issue. 
Unfortunately, Chrome still crashes after changing a markdown cell when toc2 is loaded.

jfbercher:
Modified extension loading -- in some conditions, there was a recursive call of toc generation. Hope this fixes the hanging issue.

cqcn1991:
This is fantastic. Great Work!
There is another thing I didn't mention, but it may simply the js code a little bit, making the original js file easier to maintain.

You can make the `nestesd ordered list` by pure css. No need to make the lable by js, see the second answer here. 
http://stackoverflow.com/questions/2729927/number-nested-ordered-lists-in-html

jfbercher:
@cqcn1991 thanks. As for the css nested ordered lists tip, this deserve attention. But using such approach will affect our current unique ids generation. that relies on the items numbers. 

**Export:** If anyone wants to test the html export feature, the current 
```
jupyter nbconvert FILE.ipynb --template toc2
```
will only work (hopefully) after the PR is merged, because it uses links to the css and js files in the gitgub website. In order to test the functionality, I have added a toc2 template that links to my git account (I will remove it after).  Testing can be done via
```
jupyter nbconvert FILE.ipynb --template toc3
```

janschulz:
> work (hopefully) after the PR is merged, because it uses links to the css and js files in the gitgub website

Most of the rest of the css and js in a notebook are included, so I would prefer if this is included directly as well.,.. More or less that it would be easier to send html files to my advisor (but I think this isn't possible because require.js, mathjax and jquery are also loaded from a CDN)...

cqcn1991:
@janschulz That's how my original toy version works:
https://github.com/cqcn1991/sidepyter
You can just use the `save as HTML` in Jupyter Notebook, and the toc would be there. Just like this:
https://cdn.rawgit.com/cqcn1991/sidepyter/master/sidepyter_demo.html

If you really want it, you can try it. 

I also made a `auto ouput to HTML` after running the Jupyter Notebook. If you want to know that, I can write about it.

janschulz:
JFYI: it works now :-) Thanks, this is great!

janschulz:
BTW: would it be possible to output the TOC cell as a proper markdown code instead of html? The links in that cell are currently not converted to pdf, so the toc is just for show, but not to be used to access the individual sections... 

[ok, md links also don't work in PDF: https://github.com/jupyter/nbconvert/issues/11 Damn!]

janschulz:
Ok, one issue still: 



-> The scrollbar is at the bottom (=can't scroll down any down; some parts of the scrollbar are below the start menu!), but there is still some more headlines (12, 12.1 and 12.2). Could it be that it takes 100% of the page height, but as the page header already takes a few pixels, some of it is below the visible view port?

cqcn1991:
@janschulz I'm not sure if you are talking about my toy version, or @jfbercher 's official one.

My toy version have this problem. The root reason for this is the `css` layout setting at the `notebook` (https://github.com/jupyter/notebook/issues/1202), so I haven't figure out a good way to deal with this. 

I think your guess is correct. And that's why I set the height to be 95% insteand of 100%. But it seems still not working under all circunstances. A possible solution may use `jQuery` to dynamically set the height, instead of using css to define the height.

janschulz:
@cqcn1991 Sorry, meant this PR -> updated my comment above...

nbextensions/usability/toc2/main.js
```diff
@@ -146,26 +147,14 @@ define(["require", "jquery", "base/js/namespace",  'services/config',
     $([IPython.events]).on("notebook_loaded.Notebook", function(){ // curiously, the event is not always fired or detected
```
jcb91: FYI, the event _is_ always fired, the issue is that depending on the notebook's complexity, the connection speed & lag time, it's possible for the notebook to load _before_ the extension is loaded, in which case you miss the event.

jfbercher: Yes, thanks. I have realized that in the meantime. 
Similarly, the kernel.ready event can also happen before the extension is fully loaded. In the last version, therefore, I load the stuff directly and only update things if these events occur. 
@janschulz I will upload an update tomorrow for the hidden scrollbar issue.

jfbercher:
- some fixes to sidebar (fixed the hidden scrollbar, here at least)
- changed some styles for links in toc
-  added a "Save as HTML (with toc)
file1
available in the File menu and enable to directly convert the actual notebook. This option requires the IPython kernel and is not present with other kernels. I was not able to find a way to do it for other kernels (even with a customized exporter called from js). 

**Navigate menu**

- since I realized that it is not difficult to add menus elements, I had the idea of adding a navigation menu". This menu can live together with the floating window or sidebar. It can be enabled/disabled using the nbextension configuration utility (enabled by default). 
demo2

I hope that there are not to much errors. Please report. 
But for now I really must go back to work...

There is a small issue at least with my latex_envs where all cells are refreshed at the startup. Since the building of the toc is fired each time a markdown cell renders, this adds some overload at startup (when latex_envs is loaded before toc2).

cqcn1991:
This is really good. Great work again, @jfbercher

jcb91:
so are we ok to merge this?

janschulz:
I just tested it and apart from a different highlighter for the TOC entry (looks in my case like brown...), I really like it!

cqcn1991:
Just a simple thouhts. As builders, some times we can build great features like this. But I'm afraid maybe we need to do more of this? There are a lot of people out there do not know the existence of this thing and other things.

jfbercher:
@janschulz You can change the background color in `main.css` 
``` #toc-level0 li > a:hover {  display: block; background-color: #DAA520}``` 
eg something like `{  display: block; color: white; background-color: blue}`
I am tired of the standard blue for links.
@cqcn1991 With the new mechanisms for nbextensions in 4.2 it would not be too difficult to transform it into a separate pip installable package (incorporating  the dedicated toc2.py exporter I put in extensions). Unfortunately I will not be able to work on this before, say the end of june [and at this point I would like to rewrite the export part in latex_envs]. Furthermore, the toc2 source really needs a big cleanup, following all @jcb91 's goodies in his toc2_cleanup branch (do not remove it @jcb91 !). Any help/PR/contribution would be gladly received and appreciated.

**Comment by janschulz** [comment]
One bad thing: if I have the toc activated (no matter if as sidebar or floating) and use [RISE](https://github.com/damianavila/RISE) to display it as a slideshow, the sidebar is shown :-( Not sure what can be done here, but probably either we or RISE needs to set some css to hide the toc if teh slideshow is active?

CC: @damianavila

**Comment by janschulz** [comment]
There seems to be another problem with RISE: if I have it as sidebar and disable it and then use RISE; the code area afterwards is moved to the left:


-> Could it be that the "disable" button doesnot remove all css?

**Comment by janschulz** [comment]
In nbextension page, the last item has two dots:

File: nbextensions/usability/toc2/README.md [hunk 2]
```diff
@@ -16,11 +20,30 @@ The state of these two toggles is memorized and restored on reload.
...
+It is possible to export (most of) table of contents functionalities to html. The idea is to link a relevant part of the javascript
+extension and the css, and add a small script in the html file. This is done using a template by
+```
+jupyter nbconvert FILE.ipynb --template toc
+```
+or 
+```
+jupyter nbconvert FILE.ipynb --template toc2
+```
+For the first template (toc), the files toc2.js and main.css (originally located in JUPYTER_DATA_DIR/nbextensions/usability/toc2) must reside in the same directory as intended for the html file. In the second template, these files are linked to the IPython-notebook-extensions github website. Export configuration (parameters) shall be edited directly in the template files (in templates directory JUPYTER_DATA_DIR/templates). An option "Save as HTML (with toc)" is also provided in the File menu and enable to directly convert the actual notebook. This option requires the IPython kernel and is not present with other kernels.
```
janschulz: Would there be a way to simply include these files in the template/final output? static one page files are best for sending this stuff to my adviser :-)

nbextensions/usability/toc2/README.md
```diff
@@ -33,6 +56,8 @@ from https://gist.github.com/magican/5574556
...
 - @jfbercher february 21, 2016. Fixed some issues when resizing the toc window. Now avoid overflows, clip the text and add a scrollbar. 
 - @jfbercher february 22, 2016. Add current toc number to headings anchors. This enable to get unique anchors for recurring headings with the same text. An anchor with the original ID is still created and can be used (but toc uses the new ones!). It is also possible to directly add an html anchor within the heading text. This is taken into account when building toc links (see comments in code). 
+- @jfbercher april 29, 2016. Triggered by @cqcn1991, cf [discussion here](https://github.com/ipython-contrib/IPython-notebook-extensions/issues/532),  add a sidebar option. The floating toc window can be dragged and docked as a left sidebar. The sidebar can be dragged out as a floating window. These different states are stored and restored when reloading the notebook. Add html export capability via templates toc.tpl and toc2.tpl (see above).
+- - @jfbercher may 04, 2016. Added a "Save as HTML with toc" menu. Added a new "Navigate" menu with presents the contents of the toc. Changed default styling for links in tocs.
```
janschulz: This results in two dots...

damianavila:
Yep, RISE makes dirty things all over the place so it does not surprise me that there is a collision with other extensions like this one... one of my ideas for further RISE development is trying to isolate all the RISE-specific changes to avoid this kind of collisions... but not there yet...

nbextensions/usability/toc2/README.md
jfbercher: Will correct this at next iteration.

nbextensions/usability/toc2/README.md
jfbercher: @janschulz You just have to replace the link in the template by 
```
{% include "FILENAME" ignore missing%}
``` 
This assumes that the file is in the same directory as the notebook (full paths do not work in jinja tempates). 
I do not do it because anyway you still need an internet connection to get access to jquery.  Please make a PR with an extra template if you want

jfbercher:
Perhaps would it be time to merge and let other users test and make issues/suggestions?

nbextensions/usability/toc2/README.md
janschulz: IMO the *default* should be to include stuff which is not on a CDN, so that sending the html file will work as long as a user has access to the internet... IMO the "toc must be on my drive" is useless because IMO html exports are to be exchanged with peers and so not on my drive. Or do I misunderstand anything here?

janschulz:
+1 on merging and iterating afterwards...

nbextensions/usability/toc2/README.md
jfbercher: Perhaps. You do not need to have toc or the notebook installed. Everything will work as long as the user has an internet access. 
If you do not have access to internet, you will not have jquery and the toc javascript will not work. This is not true in the notebook since it has a copy of jquery inside. Otherwise you need to have an access to jquery, and most of the time this suppose an internet access.

cqcn1991:
Would it be good if we can make a video or a gif demonstrate this and post it on sites like hackernews and some jupyter network?

nbextensions/usability/toc2/README.md
janschulz: > You do not need to have toc or the notebook installed. Everything will work as long as the user has an internet access. 

+1

jcb91:
as noted, this seems mature enough to merge, and iterate later.

PR #594, opened 2016-04-29
fix of month in datestamper

hagne:
Month is zero based in javascript, so the datestamp came out one month short.


jcb91:
Haha! How did I not notice that for so long :flushed:

Issue #595, opened 2016-04-30
install requirements


janschulz:
Currently, when installing, I get lots of stuff installed via pip (e.g. qtconsole, ipython, setuptools) because I didn't have jupyter installed via conda but only "notebook". Is `jupyter` really a hard requirement or could that be adjusted?

jcb91:
I think we can probably get away with requiring only `jupyter_core` for its
paths module, (which is required by notebook anyway) I don't think we
currently use anything else from jupyter, so should be safe to alter...

jcb91:
so, on further investigation, the current complete list of imports seems to be:
* `jupyter_core`
* `nbconvert`
* `notebook`
* `psutil`
* `six` (used in `extensions/pre_codefolding.py` but could be dispensed with)
* `tornado`
* `traitlets`
* `yaml`

plus a few which are used only by server extensions:

* `aspell` (used only by `extensions/ipy-aspell-server.py`)
* `numpy` and `zmq` (only used in `nbextensions/testing/history/ipy_nb_history_websocket.py`)
* `requests` (used only in `extensions/post_embedhtml.py`)

In addition, `nbextensions/usability/latex_envs/conversion/thmInNb_tolatex.py` uses `IPython.version_info` to decide where to import nbconvert from, in case it's used with ipython 3. Since we require notebook, which requires nbconvert, we could simplify this in the 4.x branch.

Ultimately, most of that list are also requirements for notebook, so I think we could get away with:

* `notebook`
* `psutil`
* `yaml` (`pyyaml` pip package)

with little effort.

janschulz:
I wouldn't mind these dependencies (first group), just declare them and hopefully PIP and conda will do the right thing. Six is usually installed, so it's mostly just a problem in cases like mine where i started form scratch and installed a clean env just for the nbserver.

IMO aspell could be removed now that we have a browserside way to spellcheck.

jcb91:
true, we should probably drop aspell, especially since the js side hasn't been updated to use 4.x requirejs syntax anyway...

PR #597, opened 2016-05-02
Remove aspell

jcb91:
As discussed in #595,

Remove the old aspell server extension and its associated js, since it hasn't been updated to 4.x, and is redundant now that we have client-side aspell checking.


juhasch:
👍 from my side.

Issue #613, opened 2016-05-28
Codefolding of cell magics no longer enabled


m7thon:
The codefolding code for cell magics (starting with `%%`) is currently not enabled (it is still there, though). It used to be, but this somehow got lost in the reorganization commit [e3f2afa]. It seems this was partially re-enabled by the latest commit [3a23c4e] by adding back the relevant code as a requirement [here](https://github.com/ipython-contrib/IPython-notebook-extensions/blob/master/nbextensions/usability/codefolding/main.js#L198) but without enabling it by adding `CodeMirror.fold.magic` to the lines [here](https://github.com/ipython-contrib/IPython-notebook-extensions/blob/master/nbextensions/usability/codefolding/main.js#L113-L116).

Was there a reason for deactivating this feature, or was that just an oversight?

juhasch:
I think it simply was an oversight. Care to make a pull request to enable this feature again ?

PR #614, opened 2016-05-28
Fix codefolding of cell magics

m7thon:
Simple fix to re-enable codefolding for cell magics (cells starting with %%), see issue #613. This used to work, but somehow got messed up during reorganization.


juhasch:
Fixes #613

juhasch:
Thanks!

PR #615, opened 2016-05-31
Add support for curses-style printing

jbn:
Lots of console-based programs print backspaces or solitary carriage-returns to mock progress meters. The most popular example in Python package: the `tqdm` library. Currently, `limit_output` doesn't account for control characters. This patch adds such support. 

To be explicit, this patch trades (potentially) increased memory usage for faster processing-time. The backspaces and CRs still print and take up space. But, given the usage of curses-style printing, I don't think the risk of memory exhaustion (or, tab memory-exhaustion yielding a tab-freeze) is lower than risk imposed by rewriting the output. And, given that unadorned Jupyter supports this style printing, this patch makes output muting "less surprising."

### DIFF
--- a/nbextensions/usability/limit_output/main.js
+++ b/nbextensions/usability/limit_output/main.js
@@ -36,6 +36,44 @@ define([
         return !isNaN(parseFloat(n)) && isFinite(n);
     }
 
+    function makePrintCounter() {
+        var count = 0,
+            currentCount = 0,
+            lastWasCR = false;
+
+        // Libraries like TQDM don't nessessarily send messages on clean
+        // boundaries (i.e. line breaks). This makes counting stateful!
+        var printCounter = function(str) {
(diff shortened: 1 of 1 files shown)


jcb91:
Sounds good to me. I'd agree that we're unlikely to run into memory-exhaustion problems, and even if we do, we could always make this new curses-style behaviour optional if _really_ necessary (although, if you run into the problem, I think it's probably likely you had a bigger problem that caused it in the first place!).

jbn:
I literally just found a weird regression that I don't get / can't fix at this moment. Perhaps it's best to revert for a day until I can fix it. 

Sorry, @jcb91!

jcb91:
ahaha ok!

maximsch2:
Any progress on the regression hunting? What was it by the way? I've just hit the progress-printing with limiting issue in Jupyter and was going to file a separate issue here.

Issue #609, opened 2016-05-23
toc2 links don't work when collapsible-headings is enabled


brunobeltran:
Title says it all, haven't had time to dig into the cause, but figured I'd go ahead and report it.

jcb91:
Thanks for the report. In what way do thy not work, just nothing happens when you click them? I suppose this may be a result of both extensions attempting to patch markdowncell render functions, and somehow interfering with each other, though nothing obvious springs to mind right now...

brunobeltran:
Yeah they act just like they would if the links didn't exist. Since this
only appears to happen if the target section is hidden (i.e. collapsed), my
hypothesis is that this might require actually looking at what happens to
the link targets when they are in a collapsed section, and maybe some code
to open the section before jumping to the link if it's collapsed.

jcb91:
Right, this is probably because cells in collapsed sections are `display:none`, so presumably won't show up as anchors for the links to point to...?

brunobeltran:
So I guess it's a feature request for collapsible headings then. It should
register all anchors in the collapsed sections to some kind of JavaScript
hook that opens the subsection in question and scrolls to it when one is
referenced?

jcb91:
Yep, sounds about right. Should hopefully be relatively simple...

PR #620, opened 2016-06-06
new collapsible_headings features

### LINKED ISSUE #616: [Feature Request] [Collapsible Headings] Shortcut to add heading above/below
A really slick UI feature that would save people a lot of annoyance would be a shortcut to add a heading at the current level above/below the current section. 

The most obvious use case is when you have two really long sections next to each other and want to add a section between. If you add a cell above the lower section title, the collapsible headings section (rightfully) expands the upper section. This then requires that you scroll to the bottom of the section to first add the desired section title, then usually back up again to collapse the accidentally expanded section. 

This seems silly, but in truly huge notebooks like the ones I use to organize my Ph.D. work, I end up temporarily reactivating Vimperator to navigate at that point, when collapsible-headings could in principle be enough to keep everything organized.

jcb91:
Fixes #616 and #609

### DIFF
--- a/nbextensions/usability/collapsible_headings/main.js
+++ b/nbextensions/usability/collapsible_headings/main.js
@@ -585,14 +585,13 @@ define([
 		var cached_select_reveals = select_reveals;
 		select_reveals = false;
 		var new_cell = Jupyter.notebook.insert_cell_above('markdown', index);
-		// restore cached setting
-		select_reveals = cached_select_reveals;
 		var new_text = 'New heading';
 		new_cell.set_text(new_text);
 		new_cell.set_heading_level(level);
 		new_cell.code_mirror.setSelection({line:0, ch: level + 1}, {line:0, ch: level + 1 + new_text.length});
 		Jupyter.notebook.select(index, true);
-		Jupyter.notebook.execute_cell(); // to make sure the new heading is "registered" by the plugin
+		// restore cached setting
+		select_reveals = cached_select_reveals;
 		Jupyter.notebook.focus_cell();
 		Jupyter.notebook.edit_mode();
 	}
--- a/nbextensions/usability/collapsible_headings/main.js
+++ b/nbextensions/usability/collapsible_headings/main.js
@@ -535,6 +535,30 @@ define([
 		);
 	}
 
+	function imitate_hash_click ($element) {
+		var site = $('#site');
+        var adjust = $element.offset().top - site.offset().top;
+        site.animate({scrollTop: site.scrollTop() + adjust});
+    }
+
+    function toc2_callback (evt) {
+		// evt.target is what was clicked, not what the handler was attached to
+		var toc_link = $(evt.target);
+		var href = toc_link.attr('href');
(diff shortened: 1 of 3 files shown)


jcb91:
> Was going to submit a fix actually, or you can add it. You need a
> Jupyter.notebook.execute_cell() at line 595ish of main.js. That way, if you
> add a cell below, Collapsible Headings registers it as a title before you
> put the user into edit mode, preventing it from being "bury"'d inside of an
> existing, closed section.
> 
> Not sure if this makes sense?

Ah, does your new cell get hidden then? I thought I'd taken care of that by setting `select_reveals` false while inserting the new cell, but perhaps not?

nbextensions/usability/collapsible_headings/main.js
```diff
@@ -533,6 +533,85 @@ define([
...
+		var cached_select_reveals = select_reveals;
+		select_reveals = false;
+		var new_cell = Jupyter.notebook.insert_cell_above('markdown', index);
+		// restore cached setting
+		select_reveals = cached_select_reveals;
+		var new_text = 'New heading';
+		new_cell.set_text(new_text);
+		new_cell.set_heading_level(level);
+		new_cell.code_mirror.setSelection({line:0, ch: level + 1}, {line:0, ch: level + 1 + new_text.length});
+		Jupyter.notebook.select(index, true);
```
brunobeltran: adding a
`Jupyter.notebook.execute_cell()`
here after the `Jupyter.notebook.select(...)` would prevent the cell from being hidden when below, when done in combination with the current `select_reveals` code.

jcb91: Ah, or, I should have put the select_reveals reset *after* the `Jupyter.notebook.select(index, true);` line - clue was in the name, really!

brunobeltran:
Ah, left you a line comment then. The bug specifically occurs when inserting a new cell below a closed header cell. The previously existing header cell is reclosed by your `select_reveals` trick, but since you haven't executed the new cell yet, the plugin does not know that it is a title cell, just that it's a markdown cell, so it gets hidden at the bottom of the collapsed cell that you were trying to insert it beneath.

nbextensions/usability/collapsible_headings/main.js
brunobeltran: Yes, that works!

jcb91:
Ok, I'm getting too tired to think about this clearly right now, so I think it's best to leave it for you to fix 😝 I guess the simplest way would be for you to submit a PR to the `feature/collapsible_headings` branch of my fork? I'm off to sleep, but will have a look at merging tomorrow

jcb91:
as a side-effect, this also makes toc-cell links select the correct heading when collapsible headings is enabled 😝

jcb91:
ok, I think I've successfully merged in #593 - let me know if/how well this works :)

jcb91:
Ok, I think this works, and doesn't mangle toc2, so I'm going to merge it. Let me know if it breaks something!

PR #621, opened 2016-06-06
fix requirejs usage in latex_envs

jcb91:
to fix issues with custom base_url e.g. https://github.com/ipython-contrib/IPython-notebook-extensions/issues/619


jcb91:
worth tagging @jfbercher here for latex_envs

jfbercher:
We had already solved a similar issue, I think, cf https://github.com/ipython-contrib/IPython-notebook-extensions/pull/402, and https://github.com/ipython-contrib/IPython-notebook-extensions/pull/404, but no problem, this seems a good and robust solution (in addition, it is useful to remove these hard encoded paths for @jcb91 themysto forthcoming PR).
+1

Issue #617, opened 2016-06-04
can't install on jupyter 4.1 system python debian sid


epifanio:
No matter what I try the nbextensions folder is never found. (404) 
I installed everytihing using PIP.
I also tried to remove all the ipytoh/jupyter related files, anf then a clean installation of just the jupter notebook ... and in the js log I was still having thew (404) error for a missing ```nbextensions``` directory.
Can yuo help to debug this?
https://gist.github.com/4fac50855c12ada0a6983ad2d8ff4ec1

epifanio:
maybe useful to debug, when i try to run the single user jupyter notebook:

the link: ```http://myhost:9000/nbextensions``` in the js console gives:

```
require.js?v=6da8be3…:900 TypeError: $ is not a function(…)Module.check @ require.js?v=6da8be3…:900
require.js?v=6da8be3…:900 TypeError: Cannot read property 'extend' of undefined(…)Module.check @ require.js?v=6da8be3…:900
require.js?v=6da8be3…:900 TypeError: $ is not a function(…)Module.check @ require.js?v=6da8be3…:900
require.js?v=6da8be3…:900 TypeError: Cannot read property 'build_page' of undefined(…)
```

and in the bash shell from where the process is running, i have: 

```
404 GET /static/components/underscore/underscore-min.map (50.187.182.220) 28.79ms referer=None
```

In the browser a blank page with entitled *Notebook Extension Config* is displayed.


The code below:

```
import os
import glob

from jupyter_core.paths import jupyter_config_dir, jupyter_data_dir

pats = {
    'config': os.path.join(jupyter_config_dir(), '*'),
    'extensions': os.path.join(jupyter_data_dir(), 'extensions', '*'),
    'nbextensions': os.path.join(jupyter_data_dir(), 'nbextensions', '*'),
}
for name, pat in pats.items():
    print('{}:\n  {}'.format(name, '\n  '.join(glob.glob(pat))))
```


gives me: 

```
config:
  /home/epinux/.jupyter/jupyter_nbconvert_config.py
  /home/epinux/.jupyter/migrated
  /home/epinux/.jupyter/jupyter_nbconvert_config.json
  /home/epinux/.jupyter/jupyter_notebook_config.py
  /home/epinux/.jupyter/nbconfig
  /home/epinux/.jupyter/jupyter_notebook_config.json
nbextensions:
  /home/epinux/.local/share/jupyter/nbextensions/codemirrormode
[...]

Issue #607, opened 2016-05-18
install to custom location


flying-sheep:
i want a way to install the extensions to a custom location (i.e. a location where jupyter can find them globally)

when running `setup.py install`, the files install themselves to a fixed location in my $HOME, unmanaged by pip.

i want to be able to track them with pip or my system’s package manager.

1. please create a way to do this or respect `--prefix` or so
2. how can the extensions be installed systemwide so that jupyter can still find them?

jcb91:
You could try my trav branch, which supports many of the same options as `jupyter nbextension install`:

```
--user
    Perform the operation for the current user
--system
    Perform the operation system-wide
--sys-prefix
    Use sys.prefix as the prefix for configuration files
--config-dir=<Unicode> (ThemystoConfigModifyApp.config_dir)
    Default: ''
    Full path to config directory
```

I'm hoping to merge that branch in as  a PR here eventually...

jcb91:
note that my themysto version requires an additional install step, so it'd require something like:

```
setup.py install
themysto install --sys-prefix
```

The setup.py installs a regular python package, then the `themysto install` command edits a config file to ensure the location of the python package's nbextensions is added to the jupyter notebook nbextensions search path.

jcb91:
(see also #577)

flying-sheep:
cool! so your way basically doesn’t install them to and of the `jupyter_path()`s, but to the pip path and `themysto` activates it?

or, if something still ends up in a `jupyter_path()`, which step does that?

jcb91:
yep, first one:

none of the funcitonal files are copied into any of the `jupyter_path()`s, they all stays in the pip path.
The `themysto` call only creates/modifies a config file (which will be in one of jupyter_config_path() folders, depending on which arguments are passed) to activate everything.

jcb91:
also, if you run `themysto` with the `--debug` flag, it should tell you which files it's editing as it runs.

flying-sheep:
[done](https://aur.archlinux.org/packages/jupyter-nbextensions-git/)

jcb91:
ace 😄

jcb91:
@flying-sheep FYI, I've merged trav into my master branch now.

flying-sheep:
updated the package. hmm, either i stay on git or i have to rename it to “themysto”, since your fork’s tags don’t match upstream’s versions, right?

jcb91:
[...]

Issue #606, opened 2016-05-16
Config option 'nbserver_extensions' not recognized by 'NotebookApp'.


waio1990:
Had a nice install with conda 4.0 until I ran a notebook and got this in the initialisation:

`[W 17:00:15.380 NotebookApp] Config option `nbserver_extensions` not recognized by `NotebookApp`.  Did you mean one of: `reraise_server_extension_failures, server_extensions`?`

Not sure what this means, and rather noob with jupyer.

Any help is appreciated

jcb91:
Well, this seems to be related to the change from Jupyter 4.2, when the config key used to determine which server extensions to load was changed from a list called `server_extensions`, to a dict called `nbserver_extensions`.

My guess is that you're using jupyter < 4.2, and either you're using an IPython 3.x setup (seems unlikely?), or somehow our version detection isn't working correctly.

Could you please post the lines you use to launch the jupyter server, along with the output of 

```bash
jupyter notebook --version
```

The current `configure_nbextensions.py` script attempts (in [configure_nbextensions.py#L158-L161](https://github.com/ipython-contrib/IPython-notebook-extensions/blob/a81e759c75ee9f6510009e792b1f7f6553749d7f/configure_nbextensions.py#L158-L161)) to detect support based on the value of `notebook.version_info`, which is perhaps not the right way to go about it. It would be more robust to check for the existence or otherwise of the `nbserver_extensions` attribute on the `NotebookApp` class, I suppose.

waio1990:
I have Jupyter 4.1.0, and I simply go with 
`jupyter notebook`

to start, will try after an update

jcb91:
hi, sorry, only just noticed you edit (I guess github doesn't notify for edits?)

Ok, well, the install _should_ have worked for jupyter 4.1 but it seems that for some reason the `configure_nbextensions.py` script got the version wrong. I'm still not sure why that happened.

Now, with your jupyter 4.2 install, it seems that the config file hasn't been written correctly: the python path is extended by a python config file written in [configure_nbextensions.py#L38-L59](https://github.com/ipython-contrib/IPython-notebook-extensions/blob/50ae35e4b1c5bd730a5eb470326e3f0db2c62149/configure_nbextensions.py#L38-L59) to include the jupyter data directory, into which the server extension is copied by the `install.py` script. It seems that for some reason that hasn't happened correctly for you.

What do you get from (python):

```python
import os
import glob

[...]

Issue #596, opened 2016-05-02
missing depend on six


janschulz:
It seems that the nbconvert stuff uses six, but when I install an env with nbconvert and notebook, six is not installed and therefore nbconvert to pdf fails: https://github.com/conda-forge/nbconvert-feedstock/issues/3

jcb91:
yes, apologies, `six` is used in `extensions/pre_codefolding.py`, although to be honest, it could do without it fairly easily. I made a slightly more extensive investigation of dependencies in #595, but couldn't post away from wifi. Will update there.

jcb91:
fixed in #598

PR #626, opened 2016-06-08
themysto

jcb91:
So, this is a fairly extensive set of changes, but in summary:

- overhaul installation mechanism to be a fully pip-idiomatic package, along with to a new command-line utility called `themysto` which does the configuration.
- Make nbextensions available to jupyter notebook through a server extension, rather than writing a hardcoded path to a config file. This is helpful in cases where the pip installation is performed to a virtual environment.
- Add a 'retirer' script to remove config and files from the current repo installation mechanism
- Add tests for configuration page, along with installation & retire mechanisms
- Add continuous integration testing on Travis for Linux & OS X, and on Appveyor for Windows
- Add code-coverage (for python only so far, though I guess js may also be possible through something like [gotwarlost/istanbul-middleware](https://github.com/gotwarlost/istanbul-middleware))
- Updates to configuration page:
  - reinstate selector link hrefs to give correct cursor (6bbeb9a63f7c5c33da0d03fdcbad0100020af497)
  - fix bugs with url encoding (4470c2b5c9adfdb38079772e98dce2e6c5358b2d)
  - don't list duplicated nbextensions twice. Instead, warn in server logs, identifying the yaml files in question. (c21a1986645655d7a4d2b1194227b8636ac304b9)

This should address, among others, #617, #607, #606, #579, #577, #576, #529, #527, #505, #457
and hopefully prevent us from hitting problems like #596, #558, #222.

This PR will need some extra changes before merge, since it currently refers to my fork of the repo in readme files and suchlike. The easiest way to test it may be from pip directly (see [README.rst#installation](https://github.com/jcb91/IPython-notebook-extensions/tree/0.0.4#installation) for details).

Any & all feedback welcome 😄


janschulz:
> Make nbextensions available to jupyter notebook through a server extension, rather than writing a hardcoded path to a config file. This is helpful in cases where the pip installation is performed to a virtual environment.

How does this compare to the nb extension by the anaconda guys?

For the rest: too big to review :-/

jcb91:
@janschulz 

> How does this compare to the nb extension by the anaconda guys?

Not quite sure which one you mean?

> For the rest: too big to review :-/

True. I can try to split it up into smaller chunks for separate PRs, hopefully...

janschulz:
@jcb91 I think I meant this Repo: https://github.com/Anaconda-Platform/nb_config_manager

jcb91:
> @jcb91 I think I meant this Repo: https://github.com/Anaconda-Platform/nb_config_manager

@janschulz well, that one just copies info about which nbextensions are enabled from user-specific config dir to environment config dir. Where the files for those nbextensions are stored is a separate issue, and that's what I'mm attempting to fix by providing them as part of a pip package, then patching the server's `nbextensions_path` using a server extension.

jcb91:
for a smaller illustration of the technique in question, see #631, which uses the same technique to move all of the javascript for the config serverextension into a python package

janschulz:
Ok, I found this:

```
+extension configuration
 +-----------------------
 +
 +Once you have installed the python package, you can use its ``themysto``
 +command-line script to write config files which will configure notebook and
 +nbconvert to use the jupyter server extensions, nbextensions, templates and
 +processors:
 +
 +::
 +
 +    themysto install
 +
 +The install command takes some options which can be used to set where the
 +config files are written: use
 +
 +* ``--user`` to write config files in the user's jupyter directories (default location)
 +* ``--system`` to configure in the system-wide jupyter directories
 +* ``--sys-prefix`` to respect system environment variables, used for example to
 +  configure in the jupyter directories of a virtual environment
 +
 +After the package's configuration has been installed, and the jupyter server
 +has been (re)started, simply go to the ``/nbextensions`` page in the notebook,
 +as noted above, to activate/deactivate and configure your notebook extensions.
 ```

and I have a question: What's the releationship to `jupyter nbextension install`? Is that just a wrapper?

jcb91:
> What's the releationship to jupyter nbextension install? Is that just a wrapper?

the flags (`--user`, `--system`, `--sys-prefix`) are equivalent, since they just relate to where to find config files. However, rather than a wrapper, essentially I just skipped `jupyter nbextension install` by directly editing the jupyter server's `nbextensions_path` using a jupyter server extension (`themysto.nbextenions_injector`).

So, the `themysto install` command is pretty much an analogue of `jupyter serverextension install` for both `themysto.nbextenions_injector` and `themysto.nbextenions_configurator`.

Does (any of) that make sense?

janschulz:
I'm a fan of using what upstream provides instead of inventing our own way and maybe get incompatible on the next version.

IMO the goal should be that this repo basically just hosts extensions and all the infrastructure should be upstream (including the `/nbextension` page for administrating the individual extensions).

jcb91:
> I'm a fan of using what upstream provides instead of inventing our own way and maybe get incompatible on the next version.

Unfortunately, the upstream provision becomes incompatible from one version to the next. 
We could use the inbuilt `jupyter nbextension install --py` (I did in fact do this, before coming up with my current solution). The main problem I found with that approach was the inconsistent API of `jupyter.nbextensions` across jupyter versions (4.0 vs 4.1 vs 4.2 don't provide the same functions/class methods, see #618 for an illustration of this). Additionally, attempting to uninstall things with `jupyter.nbextensions` leaves us open to breaking things that people have deliberately changed.

> IMO the goal should be that this repo basically just hosts extensions and all the infrastructure should be upstream (including the /nbextension page for administrating the individual extensions).

I guess we could (quite easily) split the config page into a separate repo, and make that a dependency of this one.

I had the impression that part of the point of this repo was to make installation of (numerous) extensions simpler for end users, which is pretty much the point of my install/retire apps, attempting to polyfilla the huge cracks in the upstream installation process. But, perhaps it'd be better to limit the scope of things provided by this repo, as you suggest. I'd be interested to know what everyone else thinks too... @juhasch, @jfbercher, any thoughts on the installation/configuration front?

juhasch:
Sorry for being slow to answer. I agree that changes are required to the repo and the way extensions are installed.

What do most people seem to want ?
This is my interpretation of mailing list and github issues:

Usually they like to add some functionality the base notebook does not give them. So most important is 
- finding an extensions
- providing this functionality
- having at least some basic documentation
- easy installation, activation and deactivation
- extensions should not mess around too much with their configuration / should just work
- instructions do not conflict with the 'official' documentation

This could mean:
* installing a single extension should be possible without copying the full repo and the configuration page server extension  ('I just want this single extension...'). 
* They should also be installable using the 'official' method in some way
* The config server extension should be split into a separate repo

As the official notebook extension installation method is in some regards inferior to what we do, I believe we still need our own installation methods. So having a 'super-installer/configurator' like 'themysto' is quite useful. My suggestion would be to split this off as a separate repo, and adding it as a `jupyter` subcommand, i.e. calling it through the `jupyter <subcommand>` machinery.

jcb91:
thanks @juhasch, some good ideas there. Some are easier than others to implement, but I think you're broadly right, and all are possible.

> * installing a single extension should be possible without copying the full repo and the configuration page server extension ('I just want this single extension...').

this is difficult to do straight from the repo, because github doesn't provide urls for single-directory archive downloads. One could install the pip package (which would include all extensions as data without installing them to jupyter), then install only the desired extension to jupyter, which has the desired effect, although it still requires the whole package to be downloaded. 

> * The config server extension should be split into a separate repo

easily done, and I think a good idea. I've already done the bulk of the work of making it into a self-contained pip package anyway. Also, as I mentioned before, we can always add/remove it as a dependency of this repo later on. I'll get that underway as `jupyter_nbextensions_configurator` unless anyone has another name they'd prefer?

> * They should also be installable using the 'official' method in some way

for most things, this is pretty simple using the `jupyter nbextension install --py` method (for those with notebook >= 4.2). Where it gets a bit more complicated is for those with earlier notebook versions, and for extensions which provide pre/post processors and templates which are for use with nbconvert (which doesn't provide similar extension installation machinery), and need to be added to configuration files etc. I think these will have to be covered by the super-installer/configurator wrapper mentioned below.

> My suggestion would be to split this off as a separate repo, and adding it as a `jupyter` subcommand, i.e. calling it through the `jupyter <subcommand>` machinery.

This is a fantastic idea! I hadn't thought of this before (indeed, it hadn't occurred to me that it was possible outside of the jupyter repos), but it would make a lot more sense to provide a command-line utility called via e.g. `jupyter contrib ...` rather than my existing `themysto ...`, which is a confusing name. I suggest putting all of the current config-modification into something like `jupyter contrib [un]configure all`, and then we can add sub-parts which can be [un]configured separately later. The config modification would then be a separate process from the pip-install. This seems pretty much what @janschulz meant by 

> My preference would be a package which brings all the extensions as data and installs them with a command

I think?

juhasch:
>> installing a single extension should be possible without copying the full repo and the >>configuration page server extension ('I just want this single extension...').

>this is difficult to do straight from the repo, because github doesn't provide urls for single-directory >archive downloads. One could install the pip package (which would include all extensions as data >without installing them to jupyter), then install only the desired extension to jupyter, which has the >desired effect, although it still requires the whole package to be downloaded. 

I believe it would be OK requiring to download the complete repo. It simply should be possible to install a single extension using the official method, also. Nothing is more frustrating than looking for a single thing and being told you also need this and that...

juhasch:
I like the idea of calling it `jupyter contrib` and then doing thinks like
`jupyter contrib install <extension-name> ...` for a single extension
`juypter contrib install --all` for all extensions
`jupyter contrib config...`
etc

This should also make it clear it is not an 'official' thing but it belongs to the Jupyter ecosystem.

juhasch:
And not to forget, with your really nice work there is already a lot done to get there.
As for splitting, I suggest at least for the spin-off packages to put them into jupyter-contrib instead of ipython-contrib now.

jcb91:
> I suggest at least for the spin-off packages to put them into jupyter-contrib instead of ipython-contrib now

good idea, though I can't do any of that as I'm not a member (or whatever the appropriate term is) there...

jfbercher:
I understand the `themysto` rationale and think  it is a right way to go (and really a super work :-). 
I like very much the `jupyter contrib` idea.

Among other possibilities, it is possible to install single extensions via `jupyter nbextension install https://rawgit.com/.../extension.zip --user`. Of course this suppose we can afford storing a zip version of the extension (since github does not give us this functionality). Similarly one could also pip-install a package embedding pre/post processors or templates for extensions that use them. 

It is a very nice work @jcb91

juhasch:
@Carreau : Could you add @jcb91 to jupyter-contrib, and give us both the right to create repos ?

jcb91:
Ok, thanks for feedback all! I've thrown together a standalone repo for the configurator at [jcb91/jupyter_nbextensions_configurator](https://github.com/jcb91/jupyter_nbextensions_configurator) (also [on pypi](https://pypi.python.org/pypi/jupyter_nbextensions_configurator), and I've attempted a build at [conda-forge/staged-recipes](https://github.com/conda-forge/staged-recipes/pull/869), but the post-link step, which is supposed to enable the extension, isn't currently working...).

janschulz:
@jcb91 This might be a nice project for https://github.com/jupyter-incubator?

Carreau:
> @Carreau : Could you add @jcb91 to jupyter-contrib, and give us both the right to create repos ?

Sorry just saw that. Made him owner as well, sorry for the delay. 
It would be really nice to move many of the things here into Jupyter itself. Though with the current policy which look like "only jupyter lab matters" I don't know how well that would be accepted.

jcb91:
@Carreau thanks, I got a notification about `ipython-contrib`, but nothing seems to have changed from `jupyter-contrib` so far.

> It would be really nice to move many of the things here into Jupyter itself.

well, if you've anything you'd like PRed, I can give it a go. But as you say, it may not be the appropriate moment for that...

jcb91:
btw, PR for separating the configurator is #644

jcb91:
@janschulz you mean the configurator, or the `jupyter contrib` idea? I've not looked at the incubator before, the process seems somewhat daunting!

Carreau:
> @Carreau thanks, I got a notification about ipython-contrib, but nothing seems to have changed from jupyter-contrib so far.

D'oh... leaving the tab open and not clicking on "invite"...
Should be ok now.

janschulz:
@jcb91 I mena the configurator page and the "format" (yaml file, ...)

jcb91:
@Carreau thanks, all good now!

@janschulz right, I see what you mean. Yeah, probably worth at least floating the idea around to get some exposure.

PR #629, opened 2016-06-08
use ipython-contrib gitter badge, rather than the @epifanio one

jcb91:


jcb91:
whoops, wrong repo branch! See #630

PR #631, opened 2016-06-12
updates to config serverextension

jcb91:
updates to config server extension:
  * move config server extensions into its own package `nbextensions_configurator`
  * use `self.finish` instead of `self.write` in `nbextensions_configurator`
  * provide magically-named function `_jupyter_server_extension_paths`
  * load `nbextensions_configurator` templates dir directly from server extension
  * don't use a mutable value (list) as a default argument
  * move nbextensions_configurator static files into its package
    * add static path to nbextensions_path in server extension
  * updates to docstrings
  * add more logging
  * get nbextensions_path directly from the running web server, remove function
    to find it, as it depends on various config values.
  * don't list duplicated nbextensions on configuration page. Instead, add a warning about them, and a line in server logs to identify the yaml files in question.
  * catch the more-generic YAMLError instead of only ScannerError
  * add kwarg to return dict from get_configurable_nbextensions; this allows us to use the yaml paths
  * add control to reset parameters
  * refer to nbextensions as enabled/disabled, as jupyter does, rather than activated/deactivated
  * reinstate link hrefs to get correct cursors
  * fix bugs with url encoding


janschulz:
+1 overall (didn't look to deeply and didn't run/test it yet)

I wonder if this functionality makes sense in notebook proper...

jcb91:
> I wonder if this functionality makes sense in notebook proper

Maybe!

Worth noting, though, but with the migration to npm packages in notebook 5.0, it would probably need substantial reworking, (also, it might make sense to switch to using parts of the npm `package.json` rather than a separate yaml file).

(this PR is basically all the config-related parts of #626)

jcb91:
closed as all of this is provided by #644

PR #632, opened 2016-06-12
Delete meta.yaml while conda recipe doesn't work

jcb91:


juhasch:
Let's remove this finally.

Issue #577, opened 2016-04-12
make package more pip-friendly


jcb91:
I've been thinking about how everything is packaged for a while, and I think it might be worth making the repo into a more conventional pip package. Although the nbextensions are largely not python packages in the normal sense, we still ship python code like the `nbextensions.py` server extension, which would benefit from being part of a package. We can adopt a model similar to jupyter, whereby the javascript stuff is installed with the python package. Making the extensions, nbextensions and templates part of a pip-installable package carries several potential benefits:

1. Currently, we edit `.py` configuration files in order to add our server-side extensions to the server's python path, so that they can be loaded correctly. This is a bit magic (not the good kind). If we were to make them part of the pip-package, then after install, they'd be on the path already, so we could just used the fully-qualified module name in the json config, and not edit the `.py` files. The same applies to nbconvert pre- and post-processors. We can just use the `NotebookApp` config items [`extra_nbextensions_path`](https://github.com/jupyter/notebook/blob/792ff7d2d2372e87d32ed2633082c282a2f15aad/notebook/notebookapp.py#L701-L703), [`extra_template_paths`](https://github.com/jupyter/notebook/blob/792ff7d2d2372e87d32ed2633082c282a2f15aad/notebook/notebookapp.py#L690-L694), and the `Exporter`'s [`template_path`](https://github.com/jupyter/nbconvert/blob/54603db3a8d3fb91fc15b1ec634a34d2927dae0c/nbconvert/exporters/templateexporter.py#L106)
2. pip-installed modules can be inherited from - it would make sense to add the various extensions and exporters as separate modules, a la #421
3. there's less ambiguity about where to install things (see, e.g. #418) if they're part of a package - they're in the package!
4. This is the direction Jupyter seem to be taking for 4.x, having nbextensions installable from python packages (see https://github.com/jupyter/notebook/pull/879).

[...]

PR #634, opened 2016-06-13
Turn repo into a proper pip package

jcb91:
this simplifies a lot of issues we have installing things, _but_ means that the configure_nbextensions step must be separated into a post-install step executed by the user. This branch hasn't yet updated any readme to reflect that.


jcb91:
(this is a small part of #626)

janschulz:
I'm not so sure about the name of the package: why themysto and not  `juypter-notebook-extensions`: for me, themysto has something to do with themes, but as far as I know there are no themes for the notebook, just new functionality.

Unfortunately, all the other PRs are also basically unreviewable, as all the renames are in these PRs as well :-(

install.py
```diff
@@ -109,7 +109,8 @@ def install_files():
     for name, src in [('python extensions', 'extensions'),
                       ('templates', 'templates'),
                       ('notebook extensions', 'nbextensions')]:
-        destination = os.path.join(data_dir, src)
+        destination = os.path.join(
+            data_dir, os.path.join('src','themysto', src))
```
janschulz: Why "src/themysto/" and not only "themysto/"?

janschulz:
Why "src/themysto/" and not only "themysto/"?

jcb91:
> I'm not so sure about the name of the package: why themysto and not `juypter-notebook-extensions`

Well, I asked in #577 whether anyone had any ideas for names, but got little repsonse on the name front.
I considered something like `jupyter-contrib-nbextensions`, but with efforts on renaming the repo stalling, decided to skip that for now in favour of something snappier.

> for me, themysto has something to do with themes

an unfortunate similarity, I guess. Themysto came from https://en.wikipedia.org/wiki/Themisto or https://en.wikipedia.org/wiki/Themisto_(moon), with the `i -> y` similar to the [calysto project](https://github.com/Calysto/calysto).


> Why "src/themysto/" and not only "themysto/"?

essentially it avoids a bunch of potential pitfalls in testing and packaging - see https://blog.ionelmc.ro/2014/05/25/python-packaging/#the-structure for details

jcb91:
> all the renames are in these PRs as well :-(

Not sure quite what you mean here? I guess the github interface doesn't render diffs because of the number of changes (and since almost every file was moved, almost every file has changed)? I generally use a local git gui  for looking at changes, so that I can see one commit at a time, rather than having to look at the whole PR at once (sourcetree is my current client of choice). But yeah, it's something of a barrier to review 😟 Any suggestions for how to make it easier?

janschulz:
> essentially it avoids a bunch of potential pitfalls in testing and packaging - see https://blog.ionelmc.ro/2014/05/25/python-packaging/#the-structure for details

Ah, I was bitten the other way with wheel builds on windows in the mpl repo: https://github.com/matplotlib/matplotlib/blob/master/patched_bdist_wheel.py

> [Github and 400 files changed] Any suggestions for how to make it easier?

Not anything else than discuss that rename first and then iterate on the rest. Aka "Get this PR merged first" :-)

jcb91:
> I also like that I don't have to install my package to test it

Yeah, I can see the appeal of that. But, given the use of tox, testing is done in a new tox install each time anyway :)

jcb91:
> Not anything else than discuss that rename first and then iterate on the rest. Aka "Get this PR merged first" :-)

Yeah, I'd have done it already (just adapting the `install.py` script to use the updated paths) if I had a more concrete package name, but, as you point out, `themysto` is perhaps not the one we want to settle on...

juhasch:
Can we close this one with the configurator split to a separate repo ?

jcb91:
well I still think we need to do something about this issue, but I think it makes sense to open new PR once I've ironed out install issues, so I'll close this.

juhasch:
I agree about the install issues. What is your opinion on installing the extensions as a package ?

jcb91:
Oh, definitely the way forward. I've got a working version in [my `pippify` branch](https://github.com/jcb91/IPython-notebook-extensions/tree/pippify), but I haven't quite finalised the script which installs the nbextensions yet...

PR #638, opened 2016-06-22
Clarifying how to not use `--user` option and get a system wide install in Linux and Mac

andyneff:
Update to README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ pip install https://github.com/ipython-contrib/IPython-notebook-extensions/archi
 ```
 - verbose mode can be enabled with -v switch eg pip -v install ...
 - upgrade with a --upgrade.
-- A system install can be done by omitting the --user switch.
+- A system install can be done by omitting the --user switch. However, on Mac and Linux, data files will be installed in the installing user's home directory. Set the environment variable `JUPYTER_DATA_DIR` to override this behavior. For example `JUPYTER_DATA_DIR=/usr/local/share/jupyter pip install ...`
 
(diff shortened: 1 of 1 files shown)


jcb91:
Thanks for this. In fact, since the package isn't really a pip package in the proper sense, the --user switch doesn't really affect much at all other than where the record of the pip-install gets put. Also, the behaviour isn't limited to Mac & Linux, it's the same on windows. Might also be worth stating that you can also set the `JUPYTER_CONFIG_DIR` environment variable to get config written system-wide.

andyneff:
Ah, you are right. I didn't catch [APPDATA is in the home dir](https://github.com/jupyter/jupyter_core/blob/38548d6b1004978591e6e492389a361ff686ac4a/jupyter_core/paths.py#L59). I suppose `%ALLUSERSPROFILE%\jupyter` would be a good suggestion there.

Problem with using JUPYTER_CONFIG_DIR during run time to keep config options system wide is then they have to be writeable via all, etc... And this is not the intended use of `/usr/share`, then we get into `umask` fun, etc... So I'm not sure how you'd want to proceed on that.

jcb91:
> then they have to be writeable via all, etc

I don't think so, unless I've misunderstood you? The `JUPYTER_CONFIG_DIR ` specified by the installing user has to be writable by that user (which we'd expect, obviously), but at runtime, the notebook server searches through and loads config files from all entries in [`jupyter_config_path()`](https://github.com/jupyter/jupyter_core/blob/b4bc8fdc83b678ec7d520009190c7ab6568aaf9b/jupyter_core/paths.py#L164), so the _running_ user only needs _read_ access to the system config files, which should be fine, no?

Issue #643, opened 2016-06-24
Error loading Latex_env


dmyersturnbull:
I followed the installation instructions at https://github.com/ipython-contrib/IPython-notebook-extensions/blob/master/README.md, downloaded https://github.com/ipython-contrib/IPython-notebook-extensions/blob/master/nbextensions/usability/latex_envs/doc/html/latex_env_doc.ipynb, and ran it in this Anaconda environment:
```
jupyter                   1.0.0                    py35_3  
jupyter_client            4.2.2                    py35_0  
jupyter_console           4.1.1                    py35_0  
jupyter_core              4.1.0                    py35_0  
```

Here's the error log when I run `%%javascript 
IPython.utils.load_extensions('calico-document-tools');`:
```
main.min.js:12156 Failed to load extension: ["nbextensions/calico-document-tools"] Error: Script error for "nbextensions/calico-document-tools"
http://requirejs.org/docs/errors.html#scripterror
    at makeError (http://localhost:8888/static/components/requirejs/require.js?v=6da8be361b9ee26c5e721e76c6d4afce:165:17)
    at HTMLScriptElement.onScriptError (http://localhost:8888/static/components/requirejs/require.js?v=6da8be361b9ee26c5e721e76c6d4afce:1732:36)(anonymous function) @ main.min.js:12156
```
LaTeX is not rendered in subsequent cells. Am I supposed to need calico-document-tools? If so, that's not clear from the documentation.

jfbercher:
The icalico extension is something else that was used for convenience in the docs. You do not need it and it has nothing to do with the latex_envs extension. 
The installation instructions in latex_envs doc file are a bit outdated (regarding paths in particular). You are welcome to update it. The best is to use the global installation procedure for all extensions, see readme, and then enable latex_envs in the nbextension tab.

dmyersturnbull:
Okay, good to know. However, LaTeX isn't rendered, and nothing displays in the console.

> The best is to use the global installation procedure for all extensions, see readme, and then enable latex_envs in the nbextension tab.

That's what I did:
`pip install https://github.com/ipython-contrib/IPython-notebook-extensions/archive/master.zip --user`
And then enabled latex_envs in the nbextension tab.

jfbercher:
I can confirm this from a fresh install. 

@jcb91 
[...]

Issue #619, opened 2016-06-06
port extension latex_envs to jupyterhub


epifanio:
Hi,

I'm migrating my environment to a multiuser jupyterhub, I'm having troubles in loading the ```latex_envs``` extension, I guess because the code is using the path to ```htpp://my.server.host:port/hub/``` instead of ```htpp://my.server.host:port/user/username```.

in the *codefolding* extension, i saw some [smart use of the config utility](https://github.com/ipython-contrib/IPython-notebook-extensions/blob/master/nbextensions/usability/codefolding/main.js#L12-L25) which perhaps can be re-used in the latex_envs code, the error in the js console give a 404 for ```thmslnNb4.js``` which I guess is called in:  https://github.com/ipython-contrib/IPython-notebook-extensions/blob/master/nbextensions/usability/latex_envs/latex_envs.js#L48-L53

jcb91:
> i saw some smart use of the config utility 

Your link here doesn't seem to point to lines using the config utility, so I'm not sure what you mean.

Yep, I think this bug is basically down to the hardcoded `nbextensions/...` urls in [nbextensions/usability/latex_envs/latex_envs.js#L50-L52](https://github.com/ipython-contrib/IPython-notebook-extensions/blob/93d19ac20a1af4f56a09bd5fe064346e616edba8/nbextensions/usability/latex_envs/latex_envs.js#L50-L52) - if you replace these with relative urls (so in this case, that should give `"./thmsInNb4"`, `"./bibInNb4"`, and `"./initNb"` respectively), I think requirejs should figure out the rest correctly?

epifanio:
sorry I meant utils:

```var base_url = utils.get_body_data("baseUrl");``` 

which can get the ```base_url```.

I'll try now your suggestions, thanks.

**Comment by jcb91** [comment]
> sorry I meant utils:

Right, I see what you mean. However, I think relative paths are still the way forward, in order not to hard-code a required install location into the extension.

**Comment by epifanio** [comment]
Yes I confirm your right, using:

```
define(["require", 
	'base/js/namespace',
	"./thmsInNb4.js",
	"./bibInNb4.js",
	"./initNb.js"], function (require,Jupyter, thmsInNb, bibsInNb,initNb) {    
```

it now loads the extensions correctly (only for the main user, the admin of the Hub) #617

jcb91:
I guess this should now be fixed by #621, so I'll close. Of course reopen if it's not fixed after all!

PR #645, opened 2016-06-25
LaTeX_envs: update paths + a define wrapper

jfbercher:
- Updated relative paths in `latex_envs.js` and `bibInNb4.js`
- suppressed a function definition in `initNb.js`

in order to address #643. 

Unfortunately, relative paths didn't worked. It would be useful to get something like that working in order to avoid hardcoding paths. 
The define function in `initNb` masked the local functions in that file, that have to be exposed more globally. 

- Also suppressed two unnecessary files


jcb91:
> Unfortunately, relative paths didn't worked. It would be useful to get something like that working in order to avoid hardcoding paths. 

Yep, I forgot that in order for files to work from relative paths, the require statements have to be from within a require/define call themselves.

> The define function in initNb masked the local functions in that file, that have to be exposed more globally.

Right. These should be exported from the `define` call in a similar method to the `load_ipython_extension` for the main extension file.

jcb91:
Ok, I'll merge this now. It will reopen something like #619 (latex_envs doesn't work with custom `base_url` because of hardcoded paths), but I figure that that will be a problem for fewer people than this...

Issue #647, opened 2016-06-26
Body of the Notebook not resized when browser windows width is changed


Kevin-McIsaac:
If I increase or decrease the width of my browser window the body of the notebook  is not resized. If I toggle the TOC off then on, the page is properly resized. 

This is also true if the TOC is off and I resize the width.

jfbercher:
Confirmed. I tried to address this in https://github.com/ipython-contrib/IPython-notebook-extensions/pull/648/commits/f05a432a1d8453e8f8a8969c2bc9b56089d8d597

Kevin-McIsaac:
This does not seem to fix the problem as when I resize the window the right margin does not seem to change.  

Instead of manually compute the width of the notebook (toc2.js lines 541 to 555) a  simpler solution would appear to be just set  #notebook-container width: auto. I tested this by setting this in my custom.css and its working fine.

jfbercher:
For me it works as intended: the right margin is fixed, the body is resized with the window, the toc being displayed or not.  There are other thinks that are done in lines 541 to 555; namely resizing the toc height to the follow window resize and taking into account if the toc is displayed or not. 
Anyway this may be done by pure css; I do not know. You are certainly welcome to PR!

PR #648, opened 2016-06-26
Toc2 window resize update

jfbercher:
Added a commit to addess #647, Unfortunately since I worked on a latex_envs issue directly on master, this one is mixed with the previous one #645. if possible merge this one that include both fixes. Otherwise I will have to rebase and PR later.


jcb91:
Sorry @jfbercher I merged #645 before reading this properly - let me know what you think is the best course of action!

jcb91:
actually, this still merges ok 😄

Issue #654, opened 2016-06-29
TOC2 plus Scratchpad breaking margins


Jessime:
I just upgraded with pip and it looks like TOC2 isn't playing nicely with Scratchpad. Here are a few screen shots to demonstrate what's happening:

1. Everything starts out normal:
![1](https://cloud.githubusercontent.com/assets/8493770/16466568/9cd8303c-3e11-11e6-8e6e-ac505a4e3050.png)

2. Until I open the scratchpad:
![2](https://cloud.githubusercontent.com/assets/8493770/16466581/ac574bce-3e11-11e6-909f-c7de9690fde2.png)

3. Even after I close the scratchpad, the toc now overlaps the notebook.
![3](https://cloud.githubusercontent.com/assets/8493770/16466604/c772b650-3e11-11e6-9b55-17db31f90f22.png)

I have to drag/snap the toc back into the upper left hand corner to make everything return to normal. This just started happening, so I don't know how reproducible this is. Does it happen for anyone else?

Thanks!

jfbercher:
I see the problem. It comes from the fact that both extensions play with the margin-left of the notebook-container. toc2 enlarges the margin in order to make space for itself; When expanding, scratchpad put this margin to 0. Thus there is necessarily an overlap. When collapsing, scratchpad restore the notebook container with the same width, but centered in the window; thus another, different, overlap.

I do not know exactly what to do. A possibility would be to intercept scratchpad collapse/expand and toggle the toc synchronously (I think that @jcb91 did something similar in his collapsible headings -- should study that) ; or patch scratchpad to not modify the left margin. Hum. Actually this last possibility works immediately; just tried it. If everybody agrees with this simple solution I can PR it.

jcb91:
@jfbercher yeah, if that (scratchpad not altering left margin) fixes things without breaking anything, then it sounds ideal!

jfbercher:
"Implemented"  in #656. Hope that it will not introduce unwanted side-effects.

jcb91:
#656 has been merged, so I'm going to go ahead and close this. Please reopen if necessary.

PR #656, opened 2016-06-30
Update Scratchpad

jfbercher:
Just commented two lines in scratchpad main.js in order to remove lest-margin modifications. So doing, the Scratchpad can be used nicely with toc2. This address #654. Hope it does not have side-effects.

Also added a hint in `README` to indicate that scratchpad cell can be executed via `Shift-Enter`.


jcb91:
Looks good to me. I'll merge as-is and wait for complaints about broken margins 😜

To see what the reason for setting the margin was, it might be worth pinging minrk, who wrote the original scratchpad at https://github.com/minrk/nbextension-scratchpad.

PR #657, opened 2016-07-05
Displays execution-time on the right

renaud:
Displays execution time on the right. Easier on the eyes...


jcb91:
so for me, this results in the border being clipped, as the timing area is no longer the full width of the cell:

image

perhaps `text-align: right` might work better?

jcb91:
closing, as this is now available through #907

PR #663, opened 2016-07-13
spring cleaning - remove latex_env temp files

jcb91:
* remove `.ipynb_checkpoints` files from latex_envs
* remove temp files from latex_envs
* add a gitignore to avoid re-adding latex temp files in future
* remove redundant `makezip` script and `creating_the_directory.txt` file.


jcb91:
I wonder if we might also consider removing the compiled html & pdf docs, since they're created form the ipynb anyway...

jfbercher:
I put the compiled html & pdf docs as examples of outputs (I also point to the html in the README if I rememer). 

Anyway I plan to rework all the export part this summer, may be more than that, and then will have to update the doc at the end.

jcb91:
ah, ok!

I forgot to add it to the summary before now, but this PR also removes the `makezip` script and the `creating_the_directory.txt` file. If that also seems sensible, I think we can safely merge?

PR #674, opened 2016-07-22
Fix bug in autoscroll extension

juhasch:
`auto_scroll_threshold` is not a function


juhasch:
Closed in favour of #683

PR #687, opened 2016-07-27
Fix/reponame

juhasch:
Replace IPython-notebook-extensions with Jupyter-notebook-extensions


jcb91:
I've started doing omething similar on top of current master. Will PR shortly

juhasch:
OK. Thanks. I have a small patch for setup.py and readme.md coming.

jcb91:
while we're about renaming things, might it be better to rename the repo with the same name as the python package (`jupyter_contrib_nbextensions`)?

juhasch:
Having the same name sounds good.

jcb91:
ok, so I'll go ahead and do that, then do a quick `s/Ipython-notebook-extensions/jupyter_contrib_nbextensions/g`

PR #691, opened 2016-07-27
CI fixes

jcb91:
to a ref before the restructure, otherwise the test doesn't make sense (and fails!)


jcb91:
not sure why appveyor coverage is failing, but merging anyway

PR #703, opened 2016-07-28
Latex_envs: reworked math numbering and exports

jfbercher:
- In this version we have reworked **equation numbering**. In the previous version, we had implemented a specialized counter and detected equations rendering for updating the counter. Meanwhile, this feature has been introduced in `MathJax` and  we now rely on MathJax implementation. We still have keep the capability of displaying only equation labels (instead of numbers). The numbering is automatically updated and is document-wide.
- It is possible that equation rendering and update is now significantly faster. 
- We have completely reworked the **notebook conversion** to plain $\LaTeX$ and html. We provide specialized exporters, pre and post processors, templates. We also added entry-points to simplify the conversion process. It is now as simple as
```bash
jupyter nbconvert --to html_lenvs FILE.ipynb
```
or 
```bash
jupyter nbconvert --to latex_lenvs FILE.ipynb
```
to convert `FILE.ipynb` into html/latex while keeping all the features of the `latex_envs` notebook extension in the converted version. The LaTeX converter also expose several conversion options (read the docs). 

You may need to reinstall the package to add the entry points, eg via `pip3 install -e jupyter_contrib_nbextensions --user`


jcb91:
looks pretty good, any chance you could get `tox -e lint` to pass (see https://travis-ci.org/ipython-contrib/jupyter_contrib_nbextensions/jobs/148165970 for details)?

jcb91:
thanks @jfbercher, looks good to me, will merge once appveyor completes 👍

jcb91:
... which it has :)

PR #705, opened 2016-07-29
Add conda receipe

juhasch:
Using PyPi as source.


jcb91:
Good plan. I've been attempting to get a conda-forge recipe going at https://github.com/conda-forge/staged-recipes/pull/1159, which has turned up some useful insights from @bollwyvl, so it might be worth taking a look at that. The essential one is that this should require `jupyter_contrib_core >=0.2`, as with https://github.com/jcb91/staged-recipes/pull/1

jcb91:
plus, the `MANIFEST.in` should be adjusted to exclude the conda.recipe folder

juhasch:
@jcb91: Great work as alway! I guess we better use the conda-forge one.

PR #707, opened 2016-07-29
Update README.md

jfbercher:
Add a further note on enabling/disabling/configuring extensions using the nbextension configurator


jcb91:
looks good to me :)

PR #709, opened 2016-07-30
fixed conflicting flags

flying-sheep:
as mentioned in https://github.com/ipython-contrib/jupyter_contrib_nbextensions/issues/607#issuecomment-236361849, the prefix flag is currently broken, so i fixed the conflict check and flag setting logic:

`user` is `True` by default, but has to be `False` when `prefix` is set

### DIFF
--- a/src/jupyter_contrib_nbextensions/application.py
+++ b/src/jupyter_contrib_nbextensions/application.py
@@ -36,6 +36,10 @@ class BaseContribNbextensionsApp(JupyterApp):
                 '%(message)s')
 
 
+USER_HELP = 'Do a user install'
+SYS_PREFIX_HELP = 'Use the sys.prefix as the prefix'
+
+
 class BaseContribNbextensionsInstallApp(BaseContribNbextensionsApp):
     """Base jupyter_contrib_nbextensions (un)installer app."""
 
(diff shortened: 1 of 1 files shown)


src/jupyter_contrib_nbextensions/application.py
```diff
@@ -85,21 +91,36 @@ class BaseContribNbextensionsInstallApp(BaseContribNbextensionsApp):
...
+
+        if len(conflicting_set) > 1:
             raise ArgumentConflict(
-                'cannot specify more than one of user, sys_prefix, or system')
-        return super(BaseContribNbextensionsInstallApp,
-                     self).parse_command_line(argv)
+                'cannot specify more than one of user, system, sys_prefix, prefix, or nbextensions, but found {}',
+                ', '.join(conflicting_set))
+        
+        # it works with flags, but no idea how to do this for string CLI args with traitlets
```
jcb91: > no idea how to do this for string CLI args with traitlets

Nor me! I suspect it may not be simple.

jcb91:
Nice implementation, but it's starting to look very complicated, and hence a bit fragile! I think since we got here from having `user` default to `True`, which was inherited from `notebook.serverextensions.ToggleServerExtensionApp`, it would make more sense (and everyone's lives much easier) to just make `user` default to `False`, which is what `notebook.nbextensions.ToggleNBExtensionApp` does anyway. In addition, since our app is mainly installing nbextensions, I think it makes sense to mirror settings for the nbextensions app anyway.

Thoughts?

flying-sheep:
well, the behavior we want is that

1. setting any of those flags or options sets the others to falsy values
2. specifying more than one of them is caught and reported to be an error.

if you find a less complex way to do it, please tell me :smiley:

jcb91:
> setting any of those flags or options sets the others to falsy values

yes, but in case more than one is set, we have no way of knowing which should take priority. This is the case whether a user sets the value, or the default is non-falsy. I think the current method of complaining where more than one is set is the only sane way to treat this.

> is fragile in case the defaults ever change

sort of, but as noted above, I think it's the only way that makes sense at all, really, so hopefully shouldn't get changed in future. Additionally, since in this case we're in control of the default settings too, I think it seems safe enough.

This is all made even more complicated by the fact that if one wants to make an operation on an install as well as a config (as is the case here), it *might* actually be a reasonable use-case for someone to use both, say, `--user` for the config file operation, and `--nbextensions` for the install operation! I'm going to split the files-related and config-related parts of the install script into separate functions, to cover such use cases (also suggested in https://github.com/conda-forge/staged-recipes/pull/1159#issuecomment-236321100), so I guess that should cover such use cases, if necessary...

flying-sheep:
> yes, but in case more than one is set, we have no way of knowing which should take priority.

sure, that’s where point 2 comes into play.

by the way: many CLI parsers work by letting the last flag take precedence, but it’s clear this project wants to disallow conflicting/duplicate flags.

> Additionally, since in this case we're in control of the default settings too, I think it seems safe enough.

we could simply add a test that checks if all defaults for the conflicting flags are falsey, with a comment explaining that before changing that test, the CLI handling has to be changed. sounds ood?

jcb91:
> we could simply add a test that checks if all defaults for the conflicting flags are falsey, with a comment explaining that before changing that test, the CLI handling has to be changed. sounds good?

Sure, sounds like a reasonable plan 👍

PR #710, opened 2016-07-30
added config dir parameter

flying-sheep:
depends on #709 and Jupyter-contrib/jupyter_contrib_core#1

### DIFF
--- a/src/jupyter_contrib_nbextensions/application.py
+++ b/src/jupyter_contrib_nbextensions/application.py
@@ -46,6 +46,7 @@ class BaseContribNbextensionsInstallApp(BaseContribNbextensionsApp):
     aliases = {
         'prefix': 'BaseContribNbextensionsInstallApp.prefix',
         'nbextensions': 'BaseContribNbextensionsInstallApp.nbextensions_dir',
+        'config-dir': 'BaseContribNbextensionsInstallApp.config_dir'
     }
 
     flags = {
@@ -92,6 +93,9 @@ class BaseContribNbextensionsInstallApp(BaseContribNbextensionsApp):
         '', config=True,
         help='Full path to nbextensions dir '
         '(consider instead using, system sys_prefix, prefix or user)')
+    config_dir = Unicode(
+        '', config=True,
+        help='Custom jupyter config directory')
 
     def parse_command_line(self, argv=None):
         """
@@ -142,7 +146,7 @@ class InstallContribNbextensionsApp(BaseContribNbextensionsInstallApp):
         return install(
             user=self.user, sys_prefix=self.sys_prefix, prefix=self.prefix,
             nbextensions_dir=self.nbextensions_dir, logger=self.log,
-            overwrite=self.overwrite, symlink=self.symlink)
+            overwrite=self.overwrite, symlink=self.symlink, config_dir=self.config_dir)
 
 
 class UninstallContribNbextensionsApp(BaseContribNbextensionsInstallApp):
--- a/src/jupyter_contrib_nbextensions/application.py
+++ b/src/jupyter_contrib_nbextensions/application.py
@@ -36,6 +36,10 @@ class BaseContribNbextensionsApp(JupyterApp):
                 '%(message)s')
 
 
+USER_HELP = 'Do a user install'
+SYS_PREFIX_HELP = 'Use the sys.prefix as the prefix'
+
+
 class BaseContribNbextensionsInstallApp(BaseContribNbextensionsApp):
     """Base jupyter_contrib_nbextensions (un)installer app."""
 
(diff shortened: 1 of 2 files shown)


jcb91:
As mentioned in Jupyter-contrib/jupyter_contrib_core#1, the config-dir kwarg isn't present in notebook 4.2, and as a result doesn't make sense to add to `jupyter_contrib_core.notebook_compat` (which is intended to be a shim providing 4.2-equivalent API). Rather, the config-dir parameter should be handled directly in the `jupyter_contrib_nbextensions` apps and functions. I built this into my `themysto` effort a while back, I think, so I'll see if I can resurrect it from that...

jcb91:
@flying-sheep I've been thinking about this a bit recently. Although I have managed to create working versions which accept a `config_dir` parameter, they necessitate rewriting lots of the upstream (de)activation code, which all depends on the `get_config_dir` function, and as a result feel very brittle to changes in the notebook API. As an alternative methodology, I suggest we consider using a user install, in combination with temporarily setting the environment variable `JUPYTER_CONFIG_DIR` (and potentially also `JUPYTER_DATA_DIR`) to appropriate values. Since the main usage case is installing by package maintainers, I think we can neglect potential side-effects of temporarily altering `os.environ`. Any thoughts?

flying-sheep:
hmm, where do runtime components get the config dir location from?

is it possible that they can’t cope with custom/systemwide installations?

jcb91:
> hmm, where do runtime components get the config dir location from?

I guess you mean where does the notebook server (and similar) look for config? Well, it's slightly OS-dependent, but I think in general they look at all directories in [`jupyter_core.paths.jupyter_config_path`](https://github.com/jupyter/jupyter_core/blob/c600d7e2d5240a02c3308d7cb4f2cfb0fa6c45c5/jupyter_core/paths.py#L164). They certainly seems to cope ok with installations to system & sys_prefix directories (also defined in `jupyter_core.paths`) ok

jcb91:
I should add, the environment variable `JUPYTER_CONFIG_DIR` is returned by [`jupyter_core.paths.jupyter_config_dir`](https://github.com/jupyter/jupyter_core/blob/c600d7e2d5240a02c3308d7cb4f2cfb0fa6c45c5/jupyter_core/paths.py#L27-L39), which in turn is used as the first entry in the list returned by [`jupyter_core.path.jupyter_config_path`](https://github.com/jupyter/jupyter_core/blob/c600d7e2d5240a02c3308d7cb4f2cfb0fa6c45c5/jupyter_core/paths.py#L164-L171)

flying-sheep:
ah, makes sense. so setting `$JUPYTER_CONFIG_DIR` during installation will work.

jcb91:
> the kernel/nbconvert/whatever will also find it, right?

As long as it's also set for their processes, yes, they should find it ok

flying-sheep:
OK, then i’ll close this. will there be a release with the system-by-default fix in it soon?

flying-sheep:
OK!

it’s a bit awkward, but it works: https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=jupyter_contrib_nbextensions

(well, almost. i have to patch a config file which stores the template location wrong)

jcb91:
> will there be a release with the system-by-default fix in it soon?

Yes, this was already merged into master in #735 (which also included #725). I'll try to make pypi & conda releases today, assuming I can get twine to play on windows...

jcb91:
Ok, I have pushed 0.2.0 to pypi, conda build to follow once it's also visible on `pypi.io` (seems to take a while to transfer over).

> i assume the user-by-default will also mess with `$JUPYTER_CONFIG_DIR`, no?

I'm not really sure what you mean here?

flying-sheep:
great!

what i meant was wrong: `actually i (counterintuitively) need to use `--user` in combination with the `$JUPYTER_*_DIR` variables to get a true system install.

it takes more fiddling than other kinds of install, but now i have something quite ideal:

```
/
├── etc/jupyter/
│   ├── jupyter_nbconvert_config.json
│   ├── jupyter_notebook_config.json
│   └── migrated
└── usr/
    ├── bin/jupyter-contrib-nbextension
    ├── lib/python3.5/site-packages/jupyter_contrib_nbextensions/
    │   └── ...
    └── share/jupyter/nbextensions/
         └── ...
```

jcb91:
@flying-sheep great, glad you got it working :smile: Thanks for the explanation, too! So are we ok to close this PR then?

flying-sheep:
yes, but it would be great to have a “true system install” option

jcb91:
sure, though since that's OS-dependent, I think it'd have to be implemented upstream before we could use it reliably :stuck_out_tongue_winking_eye:

PR #715, opened 2016-08-05
[WIP] Add docs

juhasch:
This is work in progress.

The idea is this to genrate docs using Sphinx:
* have some general documentation as `rst` files
* `md2rst.py` script:  convert `md` files of individual extensions to `rst` and copy images using pandoc to the `docs/source` directory
* generate HTML documentation

Open questions:
* Is there a better approach than converting `md` files to `rst` first and copying them ?
* How to automate the conversion process (run `md2rst.py` script)
* Add to Readthedocs


juhasch:
@jcb91 Do you think we could have CI running the md to rst conversion using pandoc automatically, so we can automate generating documentation for RTD ?

jcb91:
Yes, absolutely. I think the bets way to do it would be to have a dedicated tox environment (actually, we already have one at [tox.ini#L59-L64](https://github.com/ipython-contrib/jupyter_contrib_nbextensions/blob/48bd32b76bcd55f7a5d0595b30681512c79202dd/tox.ini#L59-L64), though it will need modification), which gets run in a travis `on_success` phase via `tox -e docs` or similar. That should take care of the build, then for the upload to RTD, we can add secure environment variables to the travis build, which get omitted from pull request builds, so that the built docs only get uploaded once the PR has been merged into the main repo...

juhasch:
Thanks @jcb91.

Closing this in favour of smaller junks.
First PR is #733 with contents of docs directory.
Follwoing PRs will update the README.md files.

jcb91:
No problem, good work on this. I've just noticed that RTD directly support GitHub webhooks, so it might even be simpler to get them to perform the build: http://docs.readthedocs.io/en/latest/webhooks.html#github

PR #721, opened 2016-08-12
only limit normal output, i.e. don't hide errors

John-Colvin:
The current behaviour hides errors (or any other output) if the "stream" output is already too large, which is a serious problem.

### DIFF
--- a/src/jupyter_contrib_nbextensions/nbextensions/limit_output/main.js
+++ b/src/jupyter_contrib_nbextensions/nbextensions/limit_output/main.js
@@ -43,14 +43,16 @@ define([
 
         oa.OutputArea.prototype._handle_output = oa.OutputArea.prototype.handle_output;
         oa.OutputArea.prototype.handle_output = function (msg) {
-            if (this.count === undefined) { this.count=0; }
-            if (this.max_count === undefined) { this.max_count = MAX_CHARACTERS; }
-            this.count = this.count + String(msg.content.text).length;
-            if(this.count > this.max_count) {
-                if (this.drop) return;
-                console.log("limit_output: output exceeded", this.max_count, "characters. Further output muted.");
-                msg.content.text = msg.content.text.substr(0, this.max_count) + params.limit_ouput_message;
-                this.drop = true;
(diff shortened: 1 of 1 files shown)


src/jupyter_contrib_nbextensions/nbextensions/limit_output/main.js
```diff
@@ -43,14 +43,16 @@ define([
...
         oa.OutputArea.prototype.handle_output = function (msg) {
-            if (this.count === undefined) { this.count=0; }
-            if (this.max_count === undefined) { this.max_count = MAX_CHARACTERS; }
-            this.count = this.count + String(msg.content.text).length;
-            if(this.count > this.max_count) {
-                if (this.drop) return;
-                console.log("limit_output: output exceeded", this.max_count, "characters. Further output muted.");
-                msg.content.text = msg.content.text.substr(0, this.max_count) + params.limit_ouput_message;
-                this.drop = true;
+            if (msg.header.msg_type === "stream") {
```
juhasch: Good idea to not mute error messages.  I would still like the other two message types `stream` and `execute_reply` to be limited, so I think
```
if (msg.header.msg_type !== "error") {
```
would be better.

John-Colvin: you mean `data` and `execute_reply`, yes? I'm not familiar with this stuff, what are they used for?

juhasch: IPython sends `execute_reply` or `display_data` messages, other Kernels send `stream` messages. Not sure if/when IPython sends `stream` messages.
Actually, filtering `display_data` does not make any sense, so only the two types `execute_reply` and `stream` should be filtered.

juhasch:
I have created a new PR https://github.com/ipython-contrib/jupyter_contrib_nbextensions/issues/729
Does this fix the issues for you ?

juhasch:
Closed by PR #729.
@John-Colvin Please open an issue if you still have problems with the extension.

PR #722, opened 2016-08-12
make file ASCII compatible

xiangcyUber:
Currently the MANIFEST.in contains a character that is incompatible with ASCII encoding, which makes pip3 unable to install the library because Python3 uses ASCII encoding. The pull request changes the character to a normal ASCII compatible space.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,4 +1,4 @@
-# explicit includes
+# explicit includes
 include COPYING.rst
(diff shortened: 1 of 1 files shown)


juhasch:
Thanks for catching.

jcb91:
:+: thanks @xiangcyUber not sure where that came from, I guess I must have pasted it from somewhere :confused:

Issue #365, opened 2015-10-14
Code formatting?


janschulz:
I just found http://www.cleancss.com/python-beautify/, so code formatting is posisble in javascript. This as a button or a hotkey would be great, especially if it would work with the line length plugin.

jcb91:
Love this idea. Will look into it.

jcb91:
So, on further investigation, it's not so easy to just use the code behind your example (which seems to be  based on [beautify-web/js-beautify](github.com/beautify-web/js-beautify), since at heart it's a javascript reformatter (with extensions for CSS and HTML). It doesn't cope with, for example,
* comments(!)
* decorators using the `@decorator` syntax
* the functions call syntax, e.g. `myfunction(*args, **kwargs)`
plus probably other things too.

While it would be possible to write a javascript-only formatter, it'd be quite a lot of effort from scratch. Something like normalising numbers of newlines between functions etc might be easy to do from scratch? Python seems (to me) to have quite a complex grammar though, so I don't know how complicated it might become.

JanSchulz:
:-(

The other way would be to open a new kernel, load a python beautifier (https://github.com/google/yapf) and open a new channel to get cleanup requests.

jcb91:
yep, something like `yapf` would be possible, but it then becomes a requirement of the extension, I guess, which may get confusing for people trying to use it. Possible though, I think.

jcb91:
On the other hand, I did also find [SecPlus/pep8.js](github.com/SecPlus/pep8.js), which spots pep8 violations without fixing them, so that might be worth adapting

mraxilus:
I would very much be interested in an extension which does something similar to yapf.

jfbercher:
Attempted an extension in #724 . 
Hope it will help!

jcb91:
closing as fixed by #724, #726

PR #724, opened 2016-08-15
yapf code autoformatter/prettifier extension

jfbercher:
Triggered by a remark by @mraxilus in #365  I attempted a little extension to autoformat/prettify Python code in a notebook's code cell.  It uses the Yapf python package to format the code. 
Of course, the extension is **not** kernel agnostic, but seems to do the work... I let the maintainers here decide if it can be added to the repo. 

It was an interesting exercise to learn how to implement bi-directional communication between javascript and the Jupyter kernel. Other kernels may be added in the future (eg R with AutoformatR).  As usual, Suggestions, corrections are welcome. In the hope it will be useful. 

This extension uses the [YAPF](https://github.com/google/yapf) Python module to reformat IPython code cells. 
**Of course, you must have `YAPF` package available somewhere in your Python path; ** 
Install via `pip install yapf [--user]`. 
 
Then the extension provides
- a toolbar button
- a keyboard shortcut (default: Ctrl-L)
for reformatting the current code-cell. 
Syntax shall be correct. The extension will point basic syntax errors, see demo. 
demo

Of course, this extension is *not* language agnostic... However, this first attempt may be the basis for a more general "code prettyfier" extension.

Installation
------------

If you use [jupyter-contrib-nbextensions](https://github.com/ipython-contrib/jupyter_contrib_nbextensions), proceed as usual. 

Otherwise, you can still install/try the extension from my personal repo, using
```
jupyter nbextension install https://github.com/jfbercher/yapf_ext/archive/master.zip --user
jupyter nbextension enable yapf_ext-master/yapf_ext
```

History: 
---------
@jfbercher, august 14, 2016, first version.


juhasch:
Nice!

mraxilus:
Wow, definitely what I was looking for. Great job.

juhasch:
Let's merge.

janschulz:
> It was an interesting exercise to learn how to implement bi-directional communication between javascript and the Jupyter kernel. Other kernels may be added in the future (eg R with AutoformatR). As usual, Suggestions, corrections are welcome. In the hope it will be useful.

Are the widget "comm" channels a way to do it? I never really looked at them, but my sketchy understanding is that they are a way to let UI communicate with the kernel and get results back.

jfbercher:
@janschulz  I did not use the comm manager (but will investigate - thanks for the link). I simply used the fact that one can specify a callback on the output of  `kernel.execute()` on the iopub channel; as in 
```
Jupyter.notebook.kernel.execute(code_input, { iopub: { output: code_exec_callback } }, { silent: false });
```
I have a prettifier for R  which works and look for javascript (via js-beautify). Then I will include both.

mraxilus:
One issue I found is that it seems to fail on docstrings. 
For example trying to run the extension on a cell like this:

```python
def test():
    """This is a docstring."""
    pass
```

Fails with this error message:
selection_006

I'm not sure if it's just my setup. 
Let me know if you experience it as well.

jfbercher:
@mraxilus I see the problem. This is a problem of quotes in strings. I embedded the text returned by `cell.get_text` in triple quotes. Thus need to escape/replace the (triple) quotes inside. 
I will update shortly.

jfbercher:
Updated in #726

ericgtaylor:
Hi, I'm using the YAPF python formatter extension, and I'm getting this bug where YAPF (or something) inserts a quote at the beginning of my cell. Any idea why this is happening? Thanks!
screen shot 2017-01-23 at 9 51 19 am

jfbercher:
Yes, this was an issue in a previous version of the extension. The `yapf_ext` (don't remember the exact name) was deprecated in favor of the `code prettifier` #726 extension, which we have largely reworked and improved recently #837. 
Please try to update your nbextensions and select the new code_prettify extension. If the bug still exists, we will try to address it of course.

PR #725, opened 2016-08-17
Change --user default to False

flying-sheep:
As agreed upon in #709


jcb91:
@flying-sheep apologies for the delay getting round to looking at this. Altering this option also needed the tests fixing to reflect the new defaults, but that's done now.

flying-sheep:
no probs!

PR #728, opened 2016-08-26
Add conda installation doc

juhasch:


jcb91:
Looks good, a couple of notes:

* the conda installation can be made one line with the `-c` option: `conda install -c conda-forge jupyter_contrib_nbextensions`
* it's worth noting that the conda installation also enables the serverextension, so the second step of the manual install isn't needed
* maybe add a link to the conda-forge site?
* adding `[skip CI]` to commit messages will skip travis etc -can be useful for commits like this which only affect docs

juhasch:
Thanks @jcb91 for the notes. I will make the changes accordingly.

PR #736, opened 2016-09-04
Update docs building

juhasch:
* Simplify docs generation (only `make html` required)
* Prepare for readthedocs
* Remove `extensions.rst`. Will be built by `md2rst.py`


docs/make.bat
```diff
@@ -46,15 +45,14 @@ if "%1" == "clean" (
 	for /d %%i in (%BUILDDIR%\*) do rmdir /q /s %%i
 	del /q /s %BUILDDIR%\*
 	for /d %%i in (source\*) do rmdir /q /s %%i
-
+	del /q /s source\extensions.rst
```
jcb91: could we maybe use the full package name of `jupyter_contrib_nbextensions` here too? Or at least specify `nbextensions` over `extensions`

juhasch: Do you think the name matters ?
It is just a filename that is being generated and contains the list of notebook extensions.

jcb91: No, it doesn't matter, in the sense that it'll still work fine regardless. But, it's just slightly less surprising/confusing to keep all the names consistent

juhasch:
@jcb91 Readthedocs does not seem to know Pandoc, so it will not build HTML docs:
```
 conda env create --name latest --file /home/docs/checkouts/readthedocs.org/user_builds/jupyter-contrib-nbextensions/checkouts/latest/docs/environment.yml

Fetching package metadata: ....
Solving package specifications: .
Error:  Package missing in current linux-64 channels: 
  - pandoc
```
 Do you have a hint for me how to proceed ?

jcb91:
@juhasch apologies for the delayed reply. I've not built anything on rtd before, so I'm not really in a position to offer any advice, I'm afraid. Perhaps we could do this not through conda? Or wait for the bugfix :smile:

juhasch:
Sure, I will investigate further. I guess your suggestion not to use conda for now is what I will try next.

PR #741, opened 2016-09-15
Make RTD build work

juhasch:


jcb91:
looks like a good plan to me!

juhasch:
Year, we got docs now :smile:
http://jupyter-contrib-nbextensions.readthedocs.io/en/latest/

juhasch:
Well, not quite. The extension descriptions are not included. It just would have been too easy...

jcb91:
haha too true! By the way, do we actually need to convert everything to rst? Since sphinx semms to support markdown docs directly, and we're writing in markdown to begin with, so not using any rst-only features, it seems like it might be possible to skip...

juhasch:
I would love to skip this step. Although sphinx supports md, I did not find a good way because:
a) you can't easily access the source code directories, especially on Windows. 
b) images don't work is you give sphinx md files.

jcb91:
> you can't easily access the source code directories, especially on Windows (i.e no symlinks). 

Sure, but we could simply copy to a temporary build directory without conversion

> images don't work is you give sphinx md files.

Ah, that is a bit of a deal-breaker...

juhasch:
Now I got it working for my local repo:
http://juhasch-jupyter-contrib-nbextensions.readthedocs.io/en/latest/index.html

I just need to do a little cleanup and it should be fine now. 
Actually recommonmark is set up for sphinx, so I might try it again so the pandoc step is no longer required.

jcb91:
ah, nice :sunglasses: 
Some of the top-level headings in the per-extension readmes need a little tweaking though, it seems. E.g, the headings for the latex_envs section `Demo/documentation` appears in the sidebar as though it's an extension itself (see http://juhasch-jupyter-contrib-nbextensions.readthedocs.io/en/latest/latex_envs/readme.html#demo-documentation)

Sterling work though :smile: :+1:

juhasch:
Yes, a few changes are still needed. I already changed most of the readme files and only skipped the ones with open PRs (at that time).

Unfortunately, it is difficult to generate PDF docs or similar. Hard to embed animated GIFs there...

jfbercher:
Very nice work @juhasch !

PR #742, opened 2016-09-15
[skip CI]Update docs

juhasch:
- add a RTD badge
- extend nbextension example


docs/source/internals.rst
```diff
@@ -10,3 +10,91 @@ Each notebook extension typically has it's own directory containing:
...
+        return {
+            load_ipython_extension : load_ipython_extension
+        };
+    });
+
+And for the `config.yaml` file:
+
+::
+
+    Type: IPython Notebook Extension
```
jcb91: We should use `Jupyter Notebook Extension` as default now

docs/source/internals.rst
```diff
@@ -10,3 +10,91 @@ Each notebook extension typically has it's own directory containing:
...
+    ], function (
+        $,
+        IPython,
+        utils,
+        configmod,
+    ) {
+        "use strict";
+
+        // define default values for config parameters
+        var params = {
```
jcb91: rather than having parameters sit in the root of the config file, it can be much neater to group them inside an object named after the extension - see `collapsible_headings` for an example. This also makes the update simpler, as it can be just a `$.extend(true, params, config.data.collapsible_headings);`, since jquery extend is robust to `undefined` values

docs/source/internals.rst
```diff
@@ -10,3 +10,91 @@ Each notebook extension typically has it's own directory containing:
...
+            for (var key in params) {
+                if (config.data.hasOwnProperty(key))
+                    params[key] = config.data[key];
+            }
+        };
+
+        // will be called when the config parameters have been loaded
+        config.loaded.then( function() {
+            update_params();
+
```
jcb91: maybe add an example usage of the config parameter, to make it clearer that you need to wait for them to be loaded before attempting to use them?

docs/source/internals.rst
```diff
@@ -10,3 +10,91 @@ Each notebook extension typically has it's own directory containing:
...
+    Description: My super duper extension
+    Parameters:
+    - name: my_config_value
+      description: A configuration parameter
+      input_type: number
+      min: -1
+      step: 1
+      default: 100
+
+When supplying a `readme.MD` file, please supply a main heading, as this will be linked in the generated documentation
```
jcb91: `.MD` -> `.md`?

juhasch:
@jcb91 Thanks for the remarks. I tried to address them.

docs/source/internals.rst
jfbercher: Yes, that's a nice way to go!

docs/source/internals.rst
```diff
@@ -59,17 +63,17 @@ The file names do not need to have the shown names, they can be choosen freely.
 
         // return public methods
         return {
-            load_ipython_extension : load_ipython_extension
+            load_jupyter_extension : load_jupyter_extension
```
jcb91: Unfortunately, I think notebook still only looks for `load_ipython_extension` - so although the function may be named `load_jupyter_extension`, we also need to list it under both keys

jcb91:
looks good :+1:

Issue #731, opened 2016-08-28
Some problem about  latex_envs


shanhaiying:
1. latex_envs extension do not support nested enumerate environment.

![image](https://cloud.githubusercontent.com/assets/1595652/18031491/922806c0-6d15-11e6-92c0-c9aaa746a120.png)

![image](https://cloud.githubusercontent.com/assets/1595652/18031493/a3523a24-6d15-11e6-9003-9b1979d1f872.png)
1. latex_envs can not work for  "\right}".

![image](https://cloud.githubusercontent.com/assets/1595652/18031480/4c7561ea-6d15-11e6-8c59-e04061a5b790.png)
![image](https://cloud.githubusercontent.com/assets/1595652/18031485/6a9cfdea-6d15-11e6-8193-b27eb64b89d5.png)

jfbercher:
- The fact that it was not possible to nest several environments of the same type was a known limitation, mentioned in the doc, due to the fact that js regular expression are not suited for recursive things. 
- The second problem was linked to an issue in accounting for markdown markup within $\LaTeX$ environments.

I have developped a solution to the recursivity problem, namely a bunch of specialized "regex"-like functions that enable to find/extract nested structures. This in turn also enabled to solve the second problem from ground, I think.  There are also several improvements in the exporter to latex and html and elsewhere. 

This gives a *major update* that you can find at https://github.com/jfbercher/jupyter_latex_envs
Installation instructions are in the README there (the Pypi version is not  ready yet).  Give it a try and make comments/ specific issues [there](https://github.com/jfbercher/jupyter_latex_envs/issues). The full documentation is [here](https://rawgit.com/jfbercher/jupyter_latex_envs/master/src/latex_envs/static/doc/latex_env_doc.html). 


![image](https://cloud.githubusercontent.com/assets/7596356/18135896/c08f22f6-6fa3-11e6-9b03-5136391aad54.png)

shanhaiying:
After the update of latex_envs,   a new problem arises.

`\begin{thm}
aa $$bb$$ cc
\end{thm}`

![image](https://cloud.githubusercontent.com/assets/1595652/18175509/1f587f34-70a4-11e6-8dee-cb5ed01b4d04.png)

jfbercher:
Thanks for testing all that. 
Please make new issues at https://github.com/jfbercher/jupyter_latex_envs. 
Fixed.

jcb91:
[...]

PR #746, opened 2016-09-21
replace vendored `latex_envs` with its pypi package

jcb91:
as mentioned in #731.

The conda-forge recipe will also need updating, which I think may require a conda-forge feedstock for the jupyter_latex_envs pypi package


jcb91:
Hmm, also just realised that this will mess up docs generation from the readme. We could (perhaps!) get around that by using `jupyter_nbextensions_configurator.get_configurable_extensions` inside a tox environment. Note sure for now...

jfbercher:
@jcb91 Thanks for that. 
I had not made this kind of PR myself because there is an additional small trouble: in the last versions of [jupyter_latex_envs](https://github.com/jfbercher/jupyter_latex_envs) setup I automatically install and enable the extension (\*); Thus it would not be fair here to have this extension automatically enabled for all users, and some auxiliary steps would have to be included in the setup. 
For this reason I was just in the process of updating the version here  with the recent updates and new features. I have also some other PR incoming. 

(\*) I was thinking to propose something similar here to remove the need for  installation step 2.

jcb91:
> I automatically install and enable the extension

Yeah, I noticed that. However, it doesn't work (for me) when installing directly from pip with

    pip install jupyter_latex_envs

I don't think it's a good idea to merge the pypi-package install step with the nbextension install step, since the variety of different ways in which people install each component is too great to get it all to work reliably for everyone all the time. As a case in point, it seems like checking `sys.argv` is insufficient for this purpose. Also, attempting to use the `--sys-prefix` flag results in

```
$  pip install https://github.com/jfbercher/jupyter_latex_envs/archive/master.zip -- sys-prefix
Usage:
  pip install [options] <requirement specifier> [package-index-options] ...
  pip install [options] -r <requirements file> [package-index-options] ...
  pip install [options] [-e] <vcs project url> ...
  pip install [options] [-e] <local project path> ...
  pip install [options] <archive url/path> ...

no such option: --sys-prefix
```

If you *really* want a one-step install command, I'd suggest writing a customised setup install command in the `setup.py`. However, I'd not recommend a one-step install at all, since it'll be unintuitive by doing something unexpected by (at least some) users. Perhaps provide a custom setup command e.g. `python setup.py install_and_enable`?

jfbercher:
Yeah, the one step install procedure is perhaps not stable. it works for me in several configs but I still mentioned the two other steps in README and doc. 
I believe that the three steps install is not intuitive at all for most users who may just want to "pip-install and try". Of course they should read the docs before trying to install!

It seeems that `--sys-prefix` supposes that you have also positioned `--py`, ie `--py --sys-prefix`, which works. 
Let's go as it is and install it as a pip dependency, this is nice.

jcb91:
> It seeems that `--sys-prefix` supposes that you have also positioned `--py`

Wow, I had never come across pip's `--py` option before, and can find no reference to it! Nice spot 👍 where did you find it?!

> I believe that the three steps install is not intuitive at all for most users

Oh, I absolutely agree, it's not a good solution, and not intuitive. However, its one saving grace is that although not intuitive, it is *standardized* by the jupyter docs, so it works for everyone without contradicting upstream documentation. I guess there isn't a good answer, really 😞 

Anyway, as far as I can tell, this PR doesn't seem to activate the `latex_envs` one-step installer, so I think we're ok to use it as-is...

jfbercher:
@jcb91 
What about this PR? 
There are some (I believe interesting :-)  new features available in `latex_envs`. 

If we do not go on with this one I can still make a (Big) PR to update the extension here.

jcb91:
This PR was fine for merging (assuming fixes for conflicts which have since appeared), _but_, it causes our conda build to break unless latex_envs also has a working conda recipe...

jcb91:
I can knock together a latex_envs conda-forge recipe, if you'd like a hand - let me know if that'd be helpful.

jfbercher:
Didn't really notice that there were something to do for conda. 
Yes, help for that recipe will he much useful since I do not know anything on conda[-forge] recipe!

jcb91:
ok, I'll give it a go. Have rebased to fix conflicts

jcb91:
Right, this should now hopefully work (fingers crossed, touch wood!).

@jfbercher if you could comment on/close any issues for which this provides solutions, that'd be super helpful 😄

jcb91:
uff, just realised that this has removed latex_envs from the [list of provided extensions on readthedocs](http://jupyter-contrib-nbextensions.readthedocs.io/en/latest/nbextensions.html), since the list is constructed from this repo's nbextensions directory. Not yet sure how best to handle that

jfbercher:
@jcb91 Thanks a lot for your help for this PR. 

I have commented on some issues related to the earlier version of `latex_envs`, but I do not have the rights to close them. 
I have also commented on some old issues for `toc2`. (there still remains some even older ones)

jcb91:
Thanks!

>  I do not have the rights to close them

Oh, sorry, I'd assumed your were on the team already, given how many fixes you provide! I think it would be a good idea to have you in ipython-contrib - @juhasch @janschulz, thoughts?

janschulz:
If I have rights here, he, for sure, should have them, too... :-) Given my current activity level, it would probably be good to remove me, too...

juhasch:
Sure.

PR #748, opened 2016-09-21
code_prettify: Corrected insufficient re replacement that led to spurious `\n` for blank lines

jfbercher:
This is for the Python prettifying part. 
Blank lines were incorrectly (of course) replaced by the `\n` string. 

In short I fell into a regex trap where I forgot that regex gobble matching characters., thus missed some matches.  The simple solution here solves this issue.


jcb91:
this (multi-regex replace) feels like something that'd benefit from tests! 😉

PR #749, opened 2016-09-21
toc2: fixed zero-size navigation menu

jfbercher:
- fixed issue with zero-length navigation menu 
The size of the resizable navigation menu is stored in notebook's metadata. The update is done automatically each time the resize event is sent. The size was not stored before any resize. At reload, the size was read as empty and the navigation menu was reduced to a zero size. 
The menu size is now stored just before saving the notebook, which solves his issue. 
- store toc parameters as an object in notebook.json (instead of messing the section with all that)
- changed loading and merging of toc parameters in system and notebook metadata. 

There is still a big cleanup to do in the code...


jcb91:
looks good to me...

PR #750, opened 2016-09-22
Add select_keymap nbextension

aiguofer:
This extension allows you to choose between CodeMirror key maps from a dropdown


jfbercher:
Thanks for that! I think that it will be a good idea to have something like that in the jupyter_contrib_nbextensions collection. I just had a look at the code and it seems nice. I will try to test the extension later this evening. 
Just two small remarks 
- you use local storage for keeping the current selection ; this is a good idea, but perhaps with the limitation that it will not be shared between different browsers. I should be possible to use `config.update(data)` where `data` is a js object, see  [here](https://github.com/jupyter/notebook/blob/master/notebook/static/services/config.js#L56-L71). 
-  capitalize may be set by css

I use [notebook_input_mode](https://github.com/asford/notebook_input_mode) for switching to vim mode. You should have a look at it, there are probably lot of things  interesting to get from it (it is left in public domain) ; namely the general command mode keybindings for vim.  Thanks again.

aiguofer:
Great, thanks for the feedback!

I really didn't like the way I added the logic for making additional key changes (and restoring the previous ones) so I'll come up with a better way to handle those and implement the ones from vim mode. I wish I would have seen that repo before I started on mine.

I chose local storage because we're sharing one server instance between a few people and we have different preferences (which is the reason why I wrote this in the first place). Although it doesn't share to other browsers, if you do use multiple browsers it's quite easy to make the selection once in each.

As for the css, that should be a simple change to make, I'll give that a try.

Not sure if I'll have time to work on this today, but will get to it soon :)

asford:
I've just merged @jfbercher's PR to support the configurator in asford/notebook_input_mode#5.

Is there a quality bar to folding extensions into this repo? @aiguofer, would you be open to collaborating on incorporating the existing extension into this PR so I can official deprecate that extension?

aiguofer:
Definitely! I should have some time to work on it tomorrow.

jfbercher:
@asford Wonderful! This is a very good news. I think that the extensions are already nice and merging the two is certainly a good way to go. I will be glad to test and review!

aiguofer:
Hey guys, I pulled over all the vim customizations from asford/notebook_input_mode as well as a few ideas (mainly updating the cells off the "defaults").

I took a stab at implementing a way to "reset" keybindings before switching modes, although I'm thinking that if there's other extensions that load after this one and change the keybindings, those settings will be lost if we change the keymap later. I see @asford just printed out a comment asking the person to reload, but I'd really like to be able to do this.

I'm also curious, @asford, why did you flatten the command_shortcuts?

asford:
@aiguofer This looks great!

That flatten operation is a little cryptic. I (of course) should have commented it better in the original implementation; it took a little digging to remember why it's needed. 

Jupyter's default keyboard shortcuts are specified in [keyboardmanager.js](https://github.com/jupyter/notebook/blob/4.2.x/notebook/static/notebook/js/keyboardmanager.js#L52) using a flat dictionary. The dict key is the shortcut key sequence/chord and the dict value is the associated action. Shortcuts requiring a key sequence (eg `d,d`) separate entries in the sequence with a comma. 

Key events are managed via `ShortcutManagers` implemented in [keyboard.js](https://github.com/jupyter/notebook/blob/4.2.x/notebook/static/base/js/keyboard.js#L360), which store shortcuts as a tree of key events with action names as leaves. Each tree level corresponds to a sequential key event that may lead to an action. For single-events shortcuts the tree is flat, however for sequential shortcuts the tree is nested.

The ShortcutManagers setup in [keyboardmanager.js](https://github.com/jupyter/notebook/blob/4.2.x/notebook/static/notebook/js/keyboardmanager.js#L40) store the currently active shortcuts in the tree representation. The flatten operation converts this tree back to the original dict representation, allowing an incremental update to the defined shortcuts in which a new shortcut is added and any old (now redundant) shortcut leading to the same action is removed.

src/jupyter_contrib_nbextensions/nbextensions/select_keymap/main.js
```diff
@@ -0,0 +1,308 @@
...
+        orig_ctrl_y(cm);
+        // remove selection after yank
+        cm.setSelection(cm.getCursor(), cm.getCursor());
+    };
+
+    CodeMirror.commands.yank_no_selection = yank_no_selection;
+    CodeMirror.commands.leave_current_mode = leave_current_mode;
+    // end custom CodeMirror command definitions
+
+    function create_menu() {
```
asford: It would be best to add a menu mirroring the current layout of the notebook `edit` page, which add keymaps as a sub-heading under the "edit" tab. For example:

````
$("#edit_menu").append('<li class="divider"></li>');
$("#edit_menu").append('<li class="dropdown-header">Key Map</li>');

available_keymaps.forEach(function(keymap) {
    $("#edit_menu").append(
        `<li><a id="keymap-${keymap}" href="#" title="Toggle ${keymap} keybindings" onClick="switch_keymap('${keymap}')" style="text-transform: capitalize;">${keymap}</a></li>`);
});

**Comment by jfbercher** [comment]
Great work @aiguofer  and @asford.
Really. 

I have played with it this afternoon and it works like a charm! :-))
I read the code and see that I have lot of things to learn from it, namely the use of the `underscore` library. Anyway it seems nice and robust. 

I do not really know how to avoid that another extension or somebody overwrite the key bindings. Anyway it should be possible to remap the keys as late as possible using eg 
```
$([Jupyter.events]).on("notebook_loaded.Notebook", function() { do remaps here }
```

I agree with @asford  suggestion of moving the keymaps menu to the edit menu, and I further suggest to put it  as a `submenu`of the edit menu. This should do the work  (moving the menu below the "move cell down" item)
```
function create_menu() {
            var menu = [
                '<li class="divider"></li>',
                '<li class="dropdown-submenu">',
                '<a href="#">Keymaps</a>',
                '<ul class="dropdown-menu" id="keymap_menu">'
            ];

            available_keymaps.forEach(function(keymap) {
                menu.push(`<li ><a id="keymap-${keymap}" href="#" title="Toggle ${keymap} keybindings" onClick="switch_keymap('${keymap}')" style="text-transform: capitalize;">${keymap}</a></li>`);
            });

            menu.push('</ul></li>');

            $('#move_cell_down').after(menu.join('\n'));
        }
```
Thanks again!

aiguofer:
The menu entry sounds good, and I agree a submenu in the edit menu might be best (it's already quite a long menu).

I tried playing around with the `notebook_loaded.Notebook` event but it seems that's already fired by the time the extension loads.

jcb91:
> I tried playing around with the notebook_loaded.Notebook event but it seems that's already fired by the time the extension loads.

Yes, this event can fire either before or after the nbextension loads, depending on network latency & the like. An approach I've seen used elsewhere is to check for the existence of `Jupyter.notebook`, and attach a `notebook_loaded.Notebook` callback only if it isn't yet defined.

This largely looks good, but I have a suggestion: as mentioned earlier, it's probably a nice idea to have the selected map stored in a config key, instead of local storage. I can also see the argument for using local storage, but the two are not incompatible - you could add a boolean config key to define where the extension ought to look for the map, to allow the user to choose which behaviour to adopt.

In addition, since the edit page also offers a keymap selector, would it make sense to apply (most of?) the same alterations to shortcuts made here in the notebook mode there as well?

aiguofer:
I made the changes to add an option to choose whether to store keymap on client side or server side, let me know what you guys think.

I also looked at the [editor](https://github.com/jupyter/notebook/blob/master/notebook/static/edit/js/menubar.js) (I wasn't aware of that menu) and it looks like they're just setting the basic CodeMirror keymaps. Since there isn't a `command_mode` in that view, some of the customizations wouldn't apply, but others would certainly be helpful. From what I've gathered, the changes would have to be added to the editor directly; if that's the case, it seems like having this in a centralized location that can be used by notebooks and editors would make the most sense. I'm not sure yet what it would take to get this integrated upstream in such a manner, but certainly worth considering.

jcb91:
@aiguofer looks good to me 👍 

>  From what I've gathered, the changes would have to be added to the editor directly

An extension can be made to load in any view (notebook, edit, tree are the only views in core notebook so far) by enabling the nbextension in the appropriate config section (see e.g. [nbextensions/tree-filter/index.js#L23](https://github.com/ipython-contrib/jupyter_contrib_nbextensions/blob/3f1122c28723e4f726596dfc153c1976c29339e6/src/jupyter_contrib_nbextensions/nbextensions/tree-filter/index.js#L23) which uses the tree section for its config parameters and enable/disable status). In addition, there is a `common` config section, where enabling the nbextension in `common` gets it loaded in all views. For use with the `jupyter_nbextensions_configurator`, the yaml file can declare which section should be used to enable/disable and store parameters with the `Section` key - see [nbextensions/tree-filter/tree-filter.yaml#L6](https://github.com/ipython-contrib/jupyter_contrib_nbextensions/blob/3f1122c28723e4f726596dfc153c1976c29339e6/src/jupyter_contrib_nbextensions/nbextensions/tree-filter/tree-filter.yaml#L6) for example.

> I'm not sure yet what it would take to get this integrated upstream in such a manner, but certainly worth considering

Definitely worth considering, since it's an added feature which (partially) exists in the edit view already, and isn't enabled by default, so shouldn't interfere with anyone's existing experience unless they explicitly activate a different keymap...

aiguofer:
Well... I've been playing around with trying to get this to work with the editor. Unfortunately, it seems like the APIs and the way things are set up are different enough that it would involve a lot more code to get it to work on both views. Additionally, the Key Map menu in the edit view is part of core, so I'd have to remove those entries to enable this without being confusing. Unfortunately, I imagine this wouldn't be looked at favorably.

It'd be great to eventually have the same options in each view, but maybe for now we can just have this extension in the Notebook view.

jcb91:
> maybe for now we can just have this extension in the Notebook view

Sure, we can always improve later :) So, I think we can now go ahead and merge as-is? All in favour? 👍

juhasch:
👍

PR #753, opened 2016-10-01
Html embed exporter

juhasch:
- Add an exporter for embedding graphics in HTML nbconvert output
- Update documentation on exporters
- Add link to documentation in README.md


jcb91:
I'm not sure why codecov seems to have decided to add statuses to PRs again, I thought it was disabled 😦

jcb91:
but anyway, I think it's complaining because your new exporter isn't tested to the extent of the rest of the codebase.

jfbercher:
The new exporter embedding linked images is certainly a good idea.

juhasch:
@jcb91: True, a test is still missing. I will add one before merging.
@jfbercher: It would be nicer to have entrypoints for postprocessors, so one could add a `--embed` flag or so. But for now I think this is the best option.

PR #754, opened 2016-10-03
select_keymap: Fix bugs introduced in code cleanup

aiguofer:
In one of my commits I had standardized use of quotes, which introduced a bug in the `onClick` call. This fixes it.


aiguofer:
Also had broken `openDialog` when I refactored the code; fixed now

jcb91:
looks reasonable

Issue #756, opened 2016-10-06
Drag and drop fails with jupyterhub


Chris7:
Drag and drop fails with jupyterhub because it constructs the url to send the data to:

`location.host + '/api/contents'...`

This does not send the request to the user's server (located at /user/xxx) and will cause a 405 error

juhasch:
I don't know anything about Jupyterhub. 
Can you say how the correct url should look like exactly, and how to find that out inJavascript ?

jcb91:
I think it should be possible to just use base.url as we do for config api.

juhasch:
Can you try if it works now ?

Chris7:
Thanks for getting around to this! I haven't tested it yet but I will let you know as soon as I do.

jcb91:
I'm going to assume this worked. @Chris7 please reopen if not!

Issue #755, opened 2016-10-04
How to make folder and save image in dragdrop extension?


re4lfl0w:
I tried to drag and drop image in my jupyter notebook.
I check how the extension are saving image?

```
<img src='test.jpg'> in markdown. and save test.jpg in notebook same directory.
```

I want to save image in 'images' directory.
## Problem
1. How do I make images directory?
2. How to save image in images directory?

I saw main.js but I didn't understand.

juhasch:
The path to the images is set as `IPython.notebook.path` in the `send_to_server()` function calls. You can add a subdirectory there. Ideally, this would be configurable and not hardcoded, like some other extensions already allow it.

To create a new subdirectory, I guess this would be possible using the contents API. You can look at the source code in the notebook at `notebook/static/services/contents.js` on how it is used.

juhasch:
I had to fix the drag&drop extension, so I added this feature. Does this work for you ?

jcb91:
assuming this was fixed by #758. Please reopen if not.

PR #758, opened 2016-10-06
[skip ci] Fix dragdrop for Jupyterhub and add subdirectory

juhasch:
- use base_url instead of location.host when generating url to send image to server
- add option to specify subdirectory
- update docs

Addresses #756 and #755


jcb91:
just to mention, I think `[skip CI]` is needed in the commit message of each new commit when pushing to a PR branch, having it in the PR title isn't enough to get the new commits also skipped.

Issue #711, opened 2016-07-31
Pypi & conda-forge package maintainers needed


jcb91:
So far, I'm the only maintainer of the jupyter_contrib_nbextensions` package [on pypi](https://pypi.python.org/pypi/jupyter_contrib_nbextensions). It would be a good idea to get some other maintainers who can make releases and the like. Any volunteers @juhasch, @jfbercher, @janschulz?

jcb91:
This will also apply to the [conda-forge feedstock](https://github.com/conda-forge/staged-recipes/pull/1159), once I get it working...

jfbercher:
I will be glad to help (don't know conda though)

janschulz:
I can help with conda (but not as the main maintainer)

jcb91:
Awesome, thanks both! Any pypi maintainers, let me know your pypi username so I can add you there. `conda-forge` uses github usernames, I think, so that should be easier.

juhasch:
I will be happy to help, too.

PR #759, opened 2016-10-08
Remove embed postprocessor call and bump to 0.2.1

juhasch:
Forgot to remove stale import.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -37,7 +37,7 @@ officially related to the Jupyter development team.
 The maturity of the provided extensions varies, please create an issue if you
 encounter any problems.
 """,
-        version='0.2.0',
+        version='0.2.1',
         author='ipython-contrib and jupyter-contrib developers',
         author_email='jupytercontrib@gmail.com',
         url=('https://github.com/'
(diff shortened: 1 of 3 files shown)


juhasch:
@jcb91 I think it would make sense to push new packages to PyPi and conda-forge due to the new exporter ?

jcb91:
Bumping version should also update [.bumpversion.cfg#L2](https://github.com/ipython-contrib/jupyter_contrib_nbextensions/blob/1dc9596af4f8976ee54ee0973f1f041346a91ba2/.bumpversion.cfg#L2) and create an annotated tag. I use

    tox -e bump

to run the whole version-bump-plus-git-commit-and-tag process (see commands at [tox.ini#L132-L134](https://github.com/ipython-contrib/jupyter_contrib_nbextensions/blob/1dc9596af4f8976ee54ee0973f1f041346a91ba2/tox.ini#L132-L134)).

>  I think it would make sense to push new packages to PyPi and conda-forge due to the new exporter?

Sure. I use similar tox environments to orchestrate the pypi build & upload: 

    tox -e clean,pypi_build,pypi_upload

I'm still the only pypi maintainer (nobody volunteered on #711 yet), so currently I'll have to do that, but it would make sense for you to get added there too for things like this. conda-forge takes its source from pypi.io, so the new recipe has to wait until the new version appears there, but essentially just submit a pr similar to https://github.com/conda-forge/jupyter_contrib_nbextensions-feedstock/pull/2 on the feedstock, then if the build runs ok, we merge it & the rest happens automagically (neat!).

PR #765, opened 2016-10-15
[Freeze] Extend functionality to markdown cells

kukanya:
I've received a request to make Freeze work with markdown cells, so here it is.
Read-only state for markdown cells equals "input can be unrendered, but cannot be edited", frozen state equals "input cannot be unrendered".
I've also done some cleanup along the way.

Thanks in advance.


jcb91:
looks good to me, thanks!

Issue #767, opened 2016-10-18
 Collapsible Headings conflicts with Code prettify


lll9p:
When I clicked the code formatting button, the errors below comes out.

```
TypeError: cell is null
http://localhost:6666/nbextensions/collapsible_headings/main.js?v=20161018093041
Line 80
```

firefox49.0.1 
jupyter-contrib-nbextensions (0.2.1)
notebook (4.2.3)

jcb91:
this results from code_prettify attempting to select an undefined cell, as a result of the commented out line at [src/jupyter_contrib_nbextensions/nbextensions/code_prettify/code_prettify.js#L127](https://github.com/ipython-contrib/jupyter_contrib_nbextensions/blob/ca8801be9e66b154e37b9f19652a5a5f0d5a2492/src/jupyter_contrib_nbextensions/nbextensions/code_prettify/code_prettify.js#L127)

jfbercher:
Indeed, there is another error in `code_prettify`; sorry for that. 
There was some code remaining from a previous version where the index of the selected cell was passed to the `yapf_format` function. But actually the index is not defined anywhere anymore. That this actually works most of the time may be because since the index is undefined, then the  select(index) does nothing and [line 130](https://github.com/ipython-contrib/jupyter_contrib_nbextensions/blob/ca8801be9e66b154e37b9f19652a5a5f0d5a2492/src/jupyter_contrib_nbextensions/nbextensions/code_prettify/code_prettify.js#L130) selects the correct cell. 
I have just removed the reference to the index and corrected the code accordingly. 
Tested on my configuration. Hope it will fix this.

jcb91:
thanks @jfbercher, I think your diagnosis seems correct, and usually it works because selecting an undefined cell isn't a problem. As a result, I think it makes sense to fix this on the `collapsible_headings` side as well, as @lll9p has done in #768

jcb91:
should now be fixed by #768, #770

PR #768, opened 2016-10-18
Fix level test code in plugin collapsible_headings

lll9p:
#767


jcb91:
although making the level-test code more robust is not a bad idea, it doesn't actually fix the original issue, which lies in `code_prettify` as mentioned in https://github.com/ipython-contrib/jupyter_contrib_nbextensions/issues/767#issuecomment-254475026

jcb91:
this seems like a good idea to have in addition to #770, as mentioned there.

Issue #764, opened 2016-10-13
Using the global config file breaks (conda) environment encapsulation


michaelpacer:
I was testing some of this in a conda environment, and then when I attempted to use nbconvert in my root environment (which does not have the extensions installed) this broke my build. 

This is because it directly edited the `~/.jupyter/jupyter_nbconvert_config.json` file to include values that expect to be able to find jupyter_contrib_nbextensions. 

This violates the environmental encapsulation and requires installing this every environment to have a working install of nbconvert.

jcb91:
Sorry to hear this is causing you problems. Could you clarify your install
process? E.g. did you use the conda-forge recipe, or the pip package? The
conda-forge recipe is intended to use the sys-prefix flag to avoid writing
to user- or system-level config files causing problems like you're
experiencing, so if that's what you tried already, then we've not passed
the flag correctly somewhere in our install...

jcb91:
to clarify, if you want to maintain environment encapsulation and use the provided `jupyter contrib nbextension install` command, you need to use the `--sys-prefix` flag to avoid having it write to config files in either the user or system directories.

michaelpacer:
I cloned the git repo & and built with an editable install.

It is not clear that the way that it is going to perform the installation was by writing a file.

How would you recommend having an editable install that also is also only going to be relative to one conda environment? I.e., should I do as I have done, but use the `--sys-prefix` flag?

jcb91:
> It is not clear that the way that it is going to perform the installation was by writing a file.

Ok, in that case it might be worth updating our docs to make clear that the install both copies files (analogous to `jupyter nbextension install`) and edits config files. Any PRs to this effect welcome 😉 

> How would you recommend having an editable install that also is also only going to be relative to one conda environment? I.e., should I do as I have done, but use the `--sys-prefix` flag?

yes, exactly. In brief,

```bash
git clone https://github.com/ipython-contrib/jupyter_contrib_nbextensions.git
pip install -e ./jupyter_contrib_nbextensions
jupyter contrib nbextension install --sys-prefix
```

To ensure edits to javascript & css in the repo are reflected in the installed files, add the `--symlink` flag as well (unavailable on Windows without admin rights, unfortunately).

jcb91:
hopefully addressed in #774 and #788 - please reopen if you think we should do something more

PR #774, opened 2016-10-19
switchable install

jcb91:
* add cli switches to install only the nbextensions files, or only the config parts
* add tests for switchable installs


jcb91:
this allows the conda recipe to keep the files as part of the build phase, with only the config as part of post-link scripts. Additionally, installing only files (no config) can be done system-wide while maintaining environment encapsulation (see #764)

PR #780, opened 2016-10-22
[ScrollDown] Initial commit

kukanya:
Hi!

One of the small problems with Jupyter notebook occures when a cell has a lot of output (perhaps it prints something on each iteration). Although the output is streamed to the notebook, you need to scroll it manually. So I've created an extension which does this scrolling for you. Hope it is helpfull!


jcb91:
neat, nice idea!

src/jupyter_contrib_nbextensions/nbextensions/scroll_down/readme.md
```diff
@@ -0,0 +1,3 @@
+# Description
```
jcb91: Sorry, I should have caught this earlier, but the main heading in the readme should be the extension name, so that it appears correctly in the built docs at http://jupyter-contrib-nbextensions.readthedocs.io/en/latest/nbextensions.html

PR #784, opened 2016-10-31
Fix limit output missing braces

azjps:
There are missing braces causing the indentation to not match the code flow. Due to the currently mismatching braces, the limit output extension will usually display an error message indicating it is limiting output, although it is not (but not indicating such in console.log).

Fixed this bug and also added ability to display the current and maximum output lengths in the error message to make this type of error easier to debug in future.


jcb91:
Good spot, thanks! Obvious once it's pointed out haha 😜

juhasch:
Thanks for spotting.

Issue #792, opened 2016-11-11
Code folding and command line execution


FlavienMakesInfographics:
Hi,

I came across a behaviour I did not expect, some strange interaction between the Code folding extension and `jupyter nbconvert`.

Let say that a notebook called `test_folding.ipynb` contains the following code
```python
if True:
    print('Whatever')
```

The folding hides `print('Whatever')`, putting a double-headed arrow on the previous line. Now, if the notebook is saved with the code folded and executed through
```
jupyter nbconvert --to notebook --execute test_folding.ipynb --output test_folding_executed.ipynb
```
the new notebook `test_folding_executed.ipynb` does *not* contain anymore `print('Whatever')` but only the double-headed arrow. Here is the resulting cell
```JSON
  {
   "cell_type": "code",
   "execution_count": 1,
   "metadata": {
    "code_folding": [
     0
    ],
    "collapsed": false
   },
   "outputs": [
    {
     "name": "stdout",
     "output_type": "stream",
     "text": [
      "Whatever\n"
     ]
    }
   ],
   "source": [
    "if True:↔\n"
   ]
  }
```

Is that an expected behaviour? On a side note, it badly screwed one of my notebook. 

I am using Jupyter version 4.2.0 and IPython version 5.1.0.

Thank you,
F.

FlavienMakesInfographics:
An update on the previous, the following code does produce the proper notebook
```python
import nbformat
from nbconvert.preprocessors import ExecutePreprocessor

with open('test_folding.ipynb') as f:
    nb = nbformat.read(f, as_version=4)

ep = ExecutePreprocessor(timeout=600, kernel_name='python3')
ep.preprocess(nb, {'metadata': {'path': 'notebooks/'}})

with open('test_folding_executed.ipynb', 'wt') as f:
    nbformat.write(nb, f)
```

jcb91:
[...]

PR #794, opened 2016-11-15
[WIP]Only execute codefolding preprocessor when requested

juhasch:
- [x] Only execute preprocessor when configuration variable is set
- [x] add test
- [x] add documentation

Addresses #792


juhasch:
@jcb91 : I would like to move the `autoclass` API docs to a separate chapter `internals`, and instead only have the description on how to use the different exporters, pre- and preprocessors under `exporting` ?

jcb91:
> I would like to move the `autoclass` API docs to a separate chapter `internals`, and instead only have the description on how to use the different exporters, pre- and preprocessors under `exporting`?

Sure, seems like a sensible move. I think we even already have an internals file :)

juhasch:
Is there any deeper meaning why `exporting.rst` is `rst` and `internals.md` is `md` format ?
Otherwise I will convert the internals stuff to `rst`.

jcb91:
No, I just kept to markdown for anything that didn't require rat-only
features, to make it easier to compare with the main (markdown) readme.
Convert away :)

docs/source/internals.rst
```diff
@@ -0,0 +1,111 @@
+.. module:: jupyter_contrib_nbextensions.nbconvert_support
+
+Notebook extension structure
+============================
+
+The nbextensions are stored each as a separate subdirectory of
+`src/jupyter_contrib_nbextensions/nbextensions`
```
jcb91: in rst, inline literals may need two backticks? I believe the notebook repo has somehow enabled use of single backticks, but I'm not sure how. The only reference to single backticks that I can find is [this](http://www.sphinx-doc.org/en/1.4.9/markup/inline.html), which seems to suggest that by default, single backticks have no special meaning

jcb91: it seems the single backticks [give the default role](http://docutils.sourceforge.net/docs/ref/rst/roles.html), which [is configurable](http://docutils.sourceforge.net/docs/ref/rst/directives.html#default-role). Perhaps we should add a

```python
default_role = 'code'
```

to `conf.py`?

juhasch: Good catch. Yes, I think we should set the default role.

juhasch:
@jcb91 Any idea why one test fails ? I don't see how my changes would affect it.

jcb91:
It's the 4.x notebook branch, and it's just failing to build notebook,
before even running our tests, so it's not really anything to do with your
changes, I think. Looks fine to merge to me!

Issue #791, opened 2016-11-10
[toc2 extension feature request] highlight TOC section with executing cell


dinya:
Is it possible to highlight TOC section which contains executing cell?

I think it will be very easy to navigate notebooks with a large number of cells.

jcb91:
This could certainly be possible, but the current toc2 machinery doesn't have a mechanism to add it in simply, so it'd need a bit of a rewrite. My best guess for how to go about it would be to loop through cells backwards instead of the current forwards while rebuilding the toc, allowing you to decide whether the current 'section' is being executed, so that once you get to the section heading, you can alter its rendering...

dinya:
@jcb91 The simplest and therefore the best solution!

> allowing you to decide whether the current 'section' is being executed, so that once you get to the section heading, you can alter its rendering...

Do you mean this way allows me to add rules to customize toc entries (for example highlighting)? Or you talk about plugin realization in general?

jcb91:
> Do you mean this way allows me to add rules to customize toc entries (for example highlighting)? Or you talk about plugin realization in general?

well, I meant that's how I'd make the plugin implement it, but the other also makes sense. To clarify, I'd do the customization by just applying a css class (plus adding some css rules), which would then allow you to add additional rules, or override the provided ones by adding to your own `custom.css`. Does that make sense?

jfbercher:
@dinya I don't completely understand what you mean by "executing cell". Do you mean the section of the _currently selected cell_, which if it is a code cell can be executed? (I mean sometimes execution is very short and it would not be as easy to catch the event). 

Actually we do not really loop over the cells (only for finding the index of the "toc_cell"  if needed) but other the headers. So this would require finding the current "executing" (may be 'selected'?) cell, and then the corresponding nearest (above) header and adding a style attribute to it. 

A point is that the toc is currently  refreshed *only* when a markdown cell is created/updated. I do not if we have an event to detect the execution of a cell; otherwise detecting execution would require to patch CodeCell prototype. Ok, while writing this I think that it is probably what you meant. Feasible, but not that easy.

jcb91:
Ah, I'd thought more in terms of cells queued for execution/executing now, but yeah, maybe selected is what was meant?

> Actually we do not really loop over the cells [...] but over the headers.

Ah, true, I'd forgotten that! Yeah, in this case it makes it more additional work than it would have been!

dinya:
[...]

Issue #762, opened 2016-10-12
Can't save notebooks when toc2 extension is enabled


DavidJBianco:
I've noticed that when I have the toc2 extension enabled, I can load notebooks just fine, but pressing the "Save" button or using the save command under the File menu doesn't work.  Looking at the jupyter server output without toc2 enabled, I get a log message that confirms that a save operation took place.  However, with the toc2, no such message.  And the notebook knows it hasn't been saved, since I get the message about unsaved changes when I try to navigate away.  

Has anyone else seen this?

DavidJBianco:
I should mention that this occurs whenever I have toc2 enabled, even if I have none of the other extensions here enabled.  I do have the IPython widgets loaded, but nothing else.

DavidJBianco:
I also found this in the JavaScript console, which seems to be the culprit.

![pcr_analysis_hunt__sqrrl_query_](https://cloud.githubusercontent.com/assets/9140349/19311658/4e8fdbc0-905d-11e6-8783-175c513e70c2.png)

jfbercher:
It seems that there is no toc section in your notebook's metadata, though it should have be created when loading the extension via main.js. Can you confirm that by looking at edit menu / Edit notebook metadata?

DavidJBianco:
Yeah, I don't see a toc section.  Here's the whole thing:

> {
>  "kernelspec": {
>    "name": "python2",
>    "display_name": "Python 2",
>    "language": "python"
>  },
>  "widgets": {
>    "state": {
>      "98936d8870f84cffa288e5ced44a8b24": {
>        "views": [
>          {
>            "cell_index": 20
>          }
>        ]
>      }
>    },
>    "version": "1.2.0"
>  },
>  "hide_input": false,
>  "language_info": {
>    "mimetype": "text/x-python",
>    "nbconvert_exporter": "python",
>    "name": "python",
>    "pygments_lexer": "ipython2",
>    "version": "2.7.11",
>    "file_extension": ".py",
>    "codemirror_mode": {
>      "version": 2,
>      "name": "ipython"
>    }
>  },
>  "celltoolbar": "Initialisation Cell"
>}

jfbercher:
Ok; It seems that config load fails somewhere. 
- Are you able to find something related to config in the js console? 
[...]

PR #795, opened 2016-11-16
toc2 update: Highlight toc headings for section with selected/edited/running cells + fix save issue

jfbercher:
## New feature
As suggested by @dinya in [#791](https://github.com/ipython-contrib/jupyter_contrib_nbextensions/issues/791), **added highlighting of the section that contains the currently edited/selected/executing cell.** 
Colors can be customized by changing `.toc-item-highlight-select` and `.toc-item-highlight-execute` classes in css. (actual highlight colors are gold yellow and red, but the red doen't render well in the gif anim :-< ) 
demo3

## Fix
- **Fixed saving issue** due to a race condition in loading/writing metadata; see issues [#1882](https://github.com/jupyter/notebook/issues/1882#issuecomment-260671282) and [#762](https://github.com/ipython-contrib/jupyter_contrib_nbextensions/issues/762)


jcb91:
Nice.

> Colors can be customized by changing .toc-item-highlight-select and .toc-item-highlight-execute classes in css

It's possible to edit css from javascript, allowing these to be configurable - see ruler nbextension for configurable colors, and code_font_size nbextension for js editing of css rules.

jfbercher:
@jcb91 Hum, that is a good idea. 
May be that it would be easier in that case to directly specify the background colors (via a `.css(.,.)) and have them configurable in the yaml file.  
[I will implement that in a next iteration (or welcome somebody else to do it), but I will be quite busy in the coming weeks.]

jfbercher:
@jcb91 
Colors are now configurable! 
[created a style section from javascript and populated it with css embedding values read from the config]

jcb91:
Awesome, looks great! One question though - looks like maybe a typo using `hoover` instead of `hover`? Hoover being the vacuum-cleaner brand/US president 😉

src/jupyter_contrib_nbextensions/nbextensions/toc2/main.js
```diff
@@ -99,48 +112,102 @@ define(["require", "jquery", "base/js/namespace",  'services/config',
...
-        if (!IPythonKernel) {
-            $('#save_html_with_toc').remove()
-        }
-        else{
-          if ($('#save_html_with_toc').length==0) addSaveAsWithToc();
-        }
-    });
+  function create_highlights_css() {
+      var sheet = document.createElement('style')
+      sheet.innerHTML = `#toc-level0 li > a:hover {  display: block; background-color: ${cfg.colors.hoover_highlight} }
```
jcb91: The backtick is not supported in android browser or internet explorer (see https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Template_literals#Browser_compatibility), so it might be a good idea to use a more basic string concatenation operation here, as otherwise it'll presumably be a syntax error, causing the whole nbextension to crash & burn, rather than failing mor gracefully.

src/jupyter_contrib_nbextensions/nbextensions/toc2/toc2.yaml
```diff
@@ -32,5 +32,16 @@ Parameters:
   description: Display Table of Contents as a navigation menu
   input_type: checkbox
   default: true
-  
+- name: toc2.colors.hoover_highlight
+  input_type: color
+  description: Hoover color in toc
+  default: ["#DAA520"]
```
jcb91: This is a list of strings (because of the `[...]`), whereas we only want one color (string). I guess this is a hangover of my example from the ruler nbextension, which *does* actually use a list of colors. Sorry, not a great example!

src/jupyter_contrib_nbextensions/nbextensions/toc2/main.js
jfbercher: Ok, it is nicer to use template strings, but indeed it might crash the extension in IE/Android. I change this to simple concatenation.

src/jupyter_contrib_nbextensions/nbextensions/toc2/toc2.yaml
jfbercher: Yep. I asked myself, but did not investigated since it worked immediately as such.

jcb91: True, I'm being a bit pedantic here, since as you point out, it works regardless of spelling, as long as it's consistent.

jcb91:
right, this looks good, I'm gonna go ahead and merge!

PR #802, opened 2016-11-27
also remove latex_envs nbconvert_support things

jcb91:
@jfbercher any thoughts on whether we ought to import relevant classes into nbconvvert_support from latex_envs, as I've done so far? It seems neat for back-compatibility, but will probably get to be a pain keeping them in sync eventually, so I think maybe it's be better to just remove them altogether, and let people use `latex_envs`'s classes directly when they want to...


docs/source/exporting.rst
```diff
@@ -48,8 +48,6 @@ http://nbconvert.readthedocs.io/en/latest/api/exporters.html
 
 .. autoclass:: LenvsHTMLExporter
```
jfbercher: I think that `LenvsHTMLExporter` can be removed too (exporter from latex_envs to HTML)

src/jupyter_contrib_nbextensions/nbconvert_support/__init__.py
```diff
@@ -2,7 +2,8 @@
 
 import os
 
-from .latex_envs import LenvsHTMLExporter, LenvsLatexExporter
+from latex_envs.latex_envs import LenvsHTMLExporter, LenvsLatexExporter
```
jfbercher: Not sure that we have to keep these two imports since the entry points are defined in the `latex_envs.py` module.

docs/source/exporting.rst
jcb91: yes, true, but this has actually already been removed in #794, so doesn't need doing again here

src/jupyter_contrib_nbextensions/nbconvert_support/__init__.py
jcb91: yep, ok, I'll update shortlt

jfbercher:
@jcb91 yes, I agree that it would be be better and simpler to let people use latex_envs's classes directly.  I will update the documentation to reflect that.

Issue #650, opened 2016-06-27
Centered title/menus when using toc2-sidebar view looks odd


Kevin-McIsaac:
Now that toc2-sidebar expands the sidebar + notebook to fit the windows having the title/menu bar centered looks odd and as the windows size is changed the menu moves. 

A more typical approach, like in Google Docs, is to have the menus set to the left.  I've set the following in .jupyter/custom/custom.css
div#menubar-container, div#header-container {
  width: auto;
}

which fixes the title/menu to the left

jfbercher:
I finally understood, 5 months later, what you meant here! 
I have tested this (+ an extra padding-left) and agree that it will look much better. I propose to include this in the toc stylesheet in a next PR.

jfbercher:
Closed by #803

PR #803, opened 2016-11-27
toc2: update for html export + notebook menus set to the left

jfbercher:
- Since previous PR, the highlighting colors for the sections in toc with selected/edited/running cells are user configurable and are set directly by javascript (config is read from system). However, when exporting to html, the style is not defined. This PR simply adds a default config (for hover highlighting) which is used by documents exported to html and overwritten in the live notebook.   

- In #650, @Kevin-McIsaac suggested to move title/menus/icons on left. 
Indeed, since the introduction of the toc sidebar, the toc2 extension resizes the main container to fit the whole window and add the sidebar. Centered menus then sometimes look odd. I agree with the suggestion. The PR adds a small styling in toc css to move the menu to the left. However, I am not sure if it should be done here, or left to the users, or put as a configuration option.  Thoughts?

### DIFF
--- a/src/jupyter_contrib_nbextensions/nbextensions/toc2/main.css
+++ b/src/jupyter_contrib_nbextensions/nbextensions/toc2/main.css
@@ -10,11 +10,11 @@ by selecting a color in the nbextension-configurator*/
 #navigate_menu li a:hover {background-color: #f1f1f1}
 
 
-/* Move menus and tooolbar to the left, following @Kevin-McIsaac suggestion */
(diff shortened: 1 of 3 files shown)


jfbercher:
Moving title/menus/icons on left:
> However, I am not sure if it should be done here, or left to the users, or put as a configuration option. Thoughts?

Finally, I have put this as a user configurable parameter (with a default to true, that is "move")

jcb91:
> I have put this as a user configurable parameter (with a default to true, that is "move")

sounds good to me!

Issue #808, opened 2016-11-30
Toc2 numbers elements in html export, contrary to parameters


NicolasWoloszko:
In the html export, toc2 automatically numbers the elements of the toc, even though this parameter is desactivated in the settings.

NicolasWoloszko:
It also losses indentation

jfbercher:
Yes, that is the default behavior when exporting to html. Note that you just have to press the "n" link in toc to turn off the numbering. That is not too bad. 

Anyway, it is easy to read the parameter and use it in the export step.

jfbercher:
As far as the second comment is concerned, are sure that this is due to toc and not to standard html export?

NicolasWoloszko:
> As far as the second comment is concerned, are sure that this is due to toc and not to standard html export?
I am no expert...

jfbercher:
Hopefully fixed by #810

PR #810, opened 2016-11-30
toc2: html export -- take into account number_section 

jfbercher:
OK, here is a possibility to address #808 
Just read the parameter in notebook's metadata and populate the parameter in the export template. 

However, this has a drawback and I am not sure that it should be merged: if the metadata toc (the toc section in metadata) does not exist, then conversion fail; even if we try to catch the exception; since actually the exception arises in the jinja2 processing. 
I think that it can be useful to convert notebook using the `html_toc`parameter even if they have not be created with toc2 activated, since it still generates the toc in that case. Thoughts?

### DIFF
--- a/src/jupyter_contrib_nbextensions/templates/toc2.tpl
+++ b/src/jupyter_contrib_nbextensions/templates/toc2.tpl
@@ -9,6 +9,15 @@
 <script type="text/javascript" src="http://ajax.googleapis.com/ajax/libs/jquery/1.9.1/jquery.min.js"></script>
 <script type="text/javascript" src="http://ajax.googleapis.com/ajax/libs/jqueryui/1.9.1/jquery-ui.min.js"></script>
 
+<style>  /* defined here in case the main.css below cannot be loaded */
+.lev1 {margin-left: 80px}
+.lev2 {margin-left: 100px}
+.lev3 {margin-left: 120px}
+.lev4 {margin-left: 140px}
+.lev5 {margin-left: 160px}
+.lev6 {margin-left: 180px}
+</style>
+
 <link rel="stylesheet" type="text/css" href="https://rawgit.com/ipython-contrib/jupyter_contrib_nbextensions/master/src/jupyter_contrib_nbextensions/nbextensions/toc2/main.css">
 
 <script src="https://rawgit.com/ipython-contrib/jupyter_contrib_nbextensions/master/src/jupyter_contrib_nbextensions/nbextensions/toc2/toc2.js"></script>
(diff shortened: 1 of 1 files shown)


jcb91:
> I think that it can be useful to convert notebook using the html_tocparameter even if they have not be created with toc2 activated, since it still generates the toc in that case.

Yes, sounds reasonable.

>  if the metadata toc (the toc section in metadata) does not exist, then conversion fail; even if we try to catch the exception; since actually the exception arises in the jinja2 processing.

This can be avoided with robust jinja processing, though, right?

src/jupyter_contrib_nbextensions/templates/toc2.tpl
```diff
@@ -16,7 +16,7 @@
 <script>
 $( document ).ready(function(){
             var cfg={'threshold':6,     // depth of toc (number of levels)
-             'number_sections':true,    // sections numbering
+             'number_sections': "{{nb['metadata']['toc']['number_sections']}}"=="True" ? true : false,  // sections numbering
```
jcb91: it would probably be more robust to use something like

```jinja2
'number_sections': {{ 'true' if nb.get('metadata', {}).get('toc', {}).get('number_sections', False) else 'false' }},
```

jfbercher: Wow.  Very nice!

NicolasWoloszko:
I have modified the toc2.tpl file on my laptop and implemented the change you suggested but i still have the same problem on export : 
capture d ecran 2016-12-01 a 11 49 30

On the HTML file, the TOC has no indetation and is numbered although I desactivated the numbering parameter in the settings.

jfbercher:
@NicolasWoloszko You will have to re-install the `jupyter_contrib_nbextensions` once this will be merged since the templates are copied in a subdirectory of the python package (its exact location varies depending on the system you use). Otherwise make the modifications in the jupyter_contrib_nbextensions/.../templates directory and reinstall by `python setup.py install` and then step 2 in README, eg `jupyter contrib nbextension install --user`

jfbercher:
@NicolasWoloszko 
**Indentation** I have checked and for me it works. 
It seems that the css is not read/present. Actually, the css is read from the `jupyter_contrib_nbextensions` site. If you try to read your html file without internet connection, it would produce this effect. A possibility is to include the relevant style directly in the notebook: try to paste
```
%%HTML
<style>
.lev1 {margin-left: 80px}
.lev2 {margin-left: 100px}
.lev3 {margin-left: 120px}
.lev4 {margin-left: 140px}
.lev5 {margin-left: 160px}
.lev6 {margin-left: 180px}
</style>
```
in a code cell.

NicolasWoloszko:
That's working thank you very much !

jfbercher:
toc2: 
- added reading of threshold (number of levels to number) and sidebar (display the sidebar or floating window) parameters from notebook metadata instead of using static defaults
- included the css used for the "static toc cell" possibly included, in case it can't be read from `main.css` (no network connection)

jcb91:
haha, let's merge now? We can always add further PRs later 😆

jfbercher:
Yes, we can merge now. I just wanted to finish the work to include the other parameters.

jcb91:
yeah, sorry, it did make sense to include, I was just getting an itchy trigger finger, so to speak.

Issue #811, opened 2016-12-01
can I limit the output for just one cell?


foice:
let's say I want output to be limited in all cells, as a general rule for the notebook. I set a limit at 1000 chars. 

now I am in a cell in which I want to see more than the limit, say I want all or 2000 chars. is there  a way to "relax" the bound for that one cell I want to see more?

jcb91:
In the current implementation, there's no way to do that. However, it would be pretty simple to implement, say by adding a value to the cell's metadata which would override the regular limit. I'll post back here once I get around to knocking something together.

foice:
Thanks a lot for picking it up!

jcb91:
OK, I've knocked together what I think should work as a a solution in PR #813. Let me know whether or not it works as you'd like

jcb91:
should be fixed by #813

PR #813, opened 2016-12-02
[limit_output] updates including allowing per-cell override of limit

jcb91:
- allow cell-specific override of output length via `cell.metadata.limit_output`
- make output-limited notes persist through notebook save & reload by adding them as actual output, rather than just calling `OutputArea.append_display_data`
- limit outputs even over several distinct output events. Previously, stdout outputs separated in time would each be limited separately, rather than taking all of the outputs as a whole.

### DIFF
--- a/src/jupyter_contrib_nbextensions/nbextensions/limit_output/main.js
+++ b/src/jupyter_contrib_nbextensions/nbextensions/limit_output/main.js
@@ -80,12 +80,11 @@ define([
             return old_handle_output.apply(this, arguments);
         };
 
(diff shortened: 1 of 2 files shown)


jcb91:
(fixes #811)

juhasch:
Looks good to me.

jcb91:
there's a long OSX backlog on Travis, so CI checks may take a long time to go through, but since these changes won't be covered by the tests anyway, I don't think we necessarily need to wait for them if it seems to work for people in their own testning

juhasch:
I think there really is no need to wait for the OSX tests, too.

Issue #651, opened 2016-06-28
Hinterland should only hint for nontrivial prefixes


dmyersturnbull:
**Steps to reproduce (simplest example):**
1. In an empty line in a Python cell, press `#` to start a comment.
2. Press enter to end the line.

**Actual result:** The line now reads `#%%!` (the first completion).

**Expected result:** The line should be an empty comment (just `#`).

The above example isn't especially annoying, but in a day of using Hinterland, I've repeatedly had to trim `%%!` from meaningful lines of code, not just comments.

jcb91:
Yes, this was a problem I ran into when writing it with the colon `:`. It's made a little more tricky to get right in general because of not knowing which language kernel may be in use. As a result, I think perhaps it makes sense to have a user-configurable set of characters after which hinting is **not** triggered, with the default including the `#`.

> I've repeatedly had to trim %%! from meaningful lines of code, not just comments.

Is this a result of other problematic characters, or something worth addressing more generally?

jcb91:
@dmyersturnbull any suggestion for what constitutes a non-trivial prefix?

dmyersturnbull:
@jcb91 Good question. Perhaps trivial prefixes are just `^[^A-Za-z\!]$` with the `!` just to catch external commands. That might be too conservative for a definition of trivial.

jcb91:
Hmm, I hadn't thought about the external command prefix. I'm not sure I follow your regex, given the `^[^A-Za-z`, you want to consider trivial anything beginning with a non-alphabetic character - wouldn't this include the external-command `!` already? Unless you mean you think external-command lines should be considered _non_-trivial, but for me they don't give useful completion suggestions, so that seems unlikely? Also, does this test apply to just the character preceding the cursor (seems like it would be enough?), or to the entirety of the line before the cursor? It's possible also that people will want to use unicode characters for variables etc, which would put a hole in this regex, but I can't think of a (simple) better one right now, other than listing every character with a special definition in python syntax. Anyway, this is getting complicated, so I think it would be best to make this a configurable option, since it will also vary between kernel languages...

jcb91:
[...]

PR #815, opened 2016-12-02
[hinterland] make all regexes configurable

jcb91:
for deciding whether to trigger autocompletion/tooltip.
Addresses #651

### DIFF
--- a/src/jupyter_contrib_nbextensions/nbextensions/hinterland/hinterland.js
+++ b/src/jupyter_contrib_nbextensions/nbextensions/hinterland/hinterland.js
@@ -96,7 +96,7 @@ define(function (require, exports, module) {
 							ch: cur.ch - 1
 						}, cur);
 						if (	pre_cursor !== '' &&
-								config.include_regexp.test(pre_cursor) &&
+								(config.include_regexp.test(pre_cursor) || config.tooltip_regexp.test(pre_cursor)) &&
 								!config.exclude_regexp.test(pre_cursor) ) {
 							if (config.tooltip_regexp.test(pre_cursor)) {
 								cell.tooltip.request(cell);
@@ -162,11 +162,11 @@ define(function (require, exports, module) {
 					config[regexp_names[ii]] = new RegExp(config[regexp_names[ii]]);
 				}
 				catch (err) {
-					console.warning(log_prefix, 'error parsing', regexp_names[ii] + ':', err);
+					console.warn(log_prefix, 'error parsing', regexp_names[ii] + ':', err);
 				}
 			}
 		}, function on_error (err) {
-			console.warning(log_prefix, 'error loading config:', err);
+			console.warn(log_prefix, 'error loading config:', err);
 		}).then(function on_success () {
 			patch_cell_keyevent();
 			add_menu_item();
--- a/src/jupyter_contrib_nbextensions/nbextensions/hinterland/hinterland.js
+++ b/src/jupyter_contrib_nbextensions/nbextensions/hinterland/hinterland.js
@@ -1,13 +1,24 @@
 define(function (require, exports, module) {
 	'use strict';
 
+	var $ = require('jquery');
+	var utils = require('base/js/utils');
 	var keyboard = require('base/js/keyboard');
 	var Cell = require('notebook/js/cell').Cell;
 	var CodeCell = require('notebook/js/codecell').CodeCell;
 	var Completer = require('notebook/js/completer').Completer;
+	var ConfigSection = require('services/config').ConfigSection;
 
 	var log_prefix = '[' + module.id + ']';
 
+	// default config (updated on nbextension load)
+	var config = {
+		enable_at_start: true,
+		exclude_regexp: ':',
+		include_regexp: '',
+		tooltip_regexp: '\\(',
+	};
+	// flag denoting whether hinting is enabled
 	var do_hinting;
 
 	// ignore most specially-named keys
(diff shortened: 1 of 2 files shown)


src/jupyter_contrib_nbextensions/nbextensions/hinterland/hinterland.js
```diff
@@ -139,9 +142,36 @@ define(function (require, exports, module) {
...
+			var regexp_names = ['exclude_regexp', 'include_regexp', 'tooltip_regexp'];
+			for (var ii=0; ii < regexp_names.length; ii++) {
+				if (config[regexp_names[ii]] === '') {
+					continue;
+				}
+				try {
+					config[regexp_names[ii]] = new RegExp(config[regexp_names[ii]]);
+				}
+				catch (err) {
+					console.warning(log_prefix, 'error parsing', regexp_names[ii] + ':', err);
```
jfbercher: console.warn() instead of console.warning() for me

src/jupyter_contrib_nbextensions/nbextensions/hinterland/hinterland.yaml
```diff
@@ -2,5 +2,38 @@ Type: Jupyter Notebook Extension
 Main: hinterland.js
 Name: Hinterland
 Link: README.md
-Description: Enable code autocompletion menu for every keypress in a code cell, instead of only enabling it with tab
+Description: |
+  Enable code autocompletion menu for every keypress in a code cell, instead of
+  only enabling it with tab
 Compatibility: 4.x
+Parameters:
+- name: enable_at_start
```
jfbercher: hinterland prefix missing for parameters (since the hinterland section is read in the js source)

src/jupyter_contrib_nbextensions/nbextensions/hinterland/hinterland.js
```diff
@@ -93,14 +95,15 @@ define(function (require, exports, module) {
...
 							ch: cur.ch - 1
 						}, cur);
-
-						if (pre_cursor !== '' && !ignore_re.test(pre_cursor)) {
-							if (pre_cursor === '(') {
+						if (	pre_cursor !== '' &&
+								config.include_regexp.test(pre_cursor) &&
+								!config.exclude_regexp.test(pre_cursor) ) {
+							if (config.tooltip_regexp.test(pre_cursor)) {
 								cell.tooltip.request(cell);
```
jfbercher: The `tooltip_regexp` character(s) must be included in the `include_regexp` characters, otherwise this is not fired. Perhaps that the the two regexp have to be merged (or more simply, the test modified perhaps  `(config.include_regexp.test(pre_cursor)  || config.tooltip_regexp.test(pre_cursor) ) && ..`  (untested)

src/jupyter_contrib_nbextensions/nbextensions/hinterland/hinterland.yaml
jcb91: thanks, good spot!

src/jupyter_contrib_nbextensions/nbextensions/hinterland/hinterland.js
jcb91: True! Too used to python logging, which seems to accept either

src/jupyter_contrib_nbextensions/nbextensions/hinterland/hinterland.js
jcb91: right, good point. Probably simplest/best to alter the test

src/jupyter_contrib_nbextensions/nbextensions/hinterland/hinterland.yaml
```diff
@@ -2,5 +2,38 @@ Type: Jupyter Notebook Extension
...
+  input_type: text
+  default: ':'
+- name: hinterland.include_regexp
+  description: |
+    A regular expression tested against the character before the cursor, which
+    must match in order for autocompletion to be triggered. If left blank, the
+    value of the notebook's Completer.reinvoke_re parameter is used, which can
+    be modified by kernels, but defaults to /[%0-9a-z._/\\:~-]/i
+  input_type: text
+  default: ''
```
jfbercher: The regex entered here is case sensitive, because the /i flag is not added for now when creating the regex object. Perhaps warn the user on that and provide an example regex, eg [%0-9A-Za-z._/\\:~-]

src/jupyter_contrib_nbextensions/nbextensions/hinterland/hinterland.yaml
```diff
@@ -2,5 +2,38 @@ Type: Jupyter Notebook Extension
...
+- name: hinterland.tooltip_regexp
+  description: |
+    A regular expression tested against the character before the cursor, which
+    if it matches, causes a tooltip to be triggered, instead of regular
+    autocompletion. For python, this is useful for example for function calls,
+    so the default regex matches opening parentheses.
+  input_type: text
+  default: '\\('
```
jfbercher: It seems that a single '\\' is enough: '\\(' ; the corresponding re is /\(/ which is correct (and works)

jfbercher: Actually both work... sorry :)

src/jupyter_contrib_nbextensions/nbextensions/hinterland/hinterland.yaml
jcb91: good point, worth adding

src/jupyter_contrib_nbextensions/nbextensions/hinterland/hinterland.yaml
jcb91: right. Depends on the regex engine, some will break given a single unescaped bracket as it would denote the start of a capturing group, but I guess javascript is prepared to treat a single bracket as a literal...

jcb91: ah, wait, sorry, I think this is actually to the complexities of YAML string parsing, which shares some similarities with bash: strings in single-quotes are more-or-less like python's 'raw' strings, which I hadn't realised. Those in double quotes allow use of `\` to escape special characters. I'm guessing that it still worked only because the yaml default doesn't actually get used, as I'd expect to end up with a regex of /\\(/? Anyway, I think I'll switch to use the double-quoted string in the YAML...

jfbercher:
It works nicely and is very useful. 
Thanks @jcb91!

jcb91:
nice, glad to hear it's useful :)

PR #818, opened 2016-12-09
Added new extension 2to3_cell

EWouters:
Converts python2 code in a notebook's code cell to python3 code

Based on [code_prettify](https://github.com/jfbercher/code_prettify) and [2to3_nb.py](https://gist.github.com/takluyver/c8839593c615bb2f6e80)

Please let me know what you think!


jfbercher:
@EWouters,
Making a 2to3 extension is a good idea!
[Actually I made a unfinished draft of something like that for my personal use, so this is very welcome.]
I will review the code, test the extension and suggest possible improvements (I already saw a possible improvement with respect to kernel language use). 
Thanks for this contribution!

EWouters:
Awesome, I'm looking forward to see what you'll find. In the mean time I added you to the [readme](https://github.com/EWouters/jupyter_contrib_nbextensions/blob/master/src/jupyter_contrib_nbextensions/nbextensions/2to3_cell/README.md).

jfbercher:
@EWouters I have pushed [some suggestions](https://github.com/EWouters/2to3_cell/pull/1) directly to your repo. Sort and choose what you find useful.  I have also added a [gif-demo](https://github.com/jfbercher/code_prettify/blob/2to3/demo.gif): it works nicely!

jfbercher:
Fine. Let's wait for others to test and/or comment.

jcb91:
I think we may as well merge this, and iterate later as necessary?

EWouters:
It sounds like a good idea to provide some JavaScript library to handle the kernel execution. However, I think `code_prettify` and `2to3_cell` would remain separate extensions as their use case is quite different.

I'm ok with merging this version. In my fork I have a [branch] (https://github.com/EWouters/2to3_cell/tree/cell-magic?files=1) that also includes a cell magic to execute python2 code on a python3 kernel, but I don't think many people would find that useful, given the possibility to just convert the code.

jcb91:
> It sounds like a good idea to provide some JavaScript library to handle the kernel execution

Yeah, a library would be the right approach. It can get kinda tricky with nbextensions, as they have no package-manager type way of specifying dependencies (they can `require` things, but have no way to ensure that something they require gets installed). However if they're both just part of this repo, then that would mean we can orchestrate their install correctly.

> In my fork I have a [branch](https://github.com/EWouters/2to3_cell/tree/cell-magic?files=1) that also includes a cell magic to execute python2 code on a python3 kernel

Well, we may as well include it, since you've made it nicely configurable, and it defaults to off, so it shouldn't be intrusive for anyone not using it...

src/jupyter_contrib_nbextensions/nbextensions/2to3_cell/2to3_cell.yaml
```diff
@@ -0,0 +1,11 @@
+Type: IPython Notebook Extension
+Name: 2to3_cell
+Description: Converts python2 code in a notebook's code cell to python3 code
+Link: README.md
+Main: 2to3_cell.js
+Compatibility: Jupyter (4.x)
+Parameters:
+- name: conv2to3.hotkey 
+  description: Hotkey to convert the current cell from python2 to python3.
+  input_type: text
```
jcb91: You can use `input_type: hotkey` to get the configurator to provide a key-binding editor for you (see for example [toggle_all_line_numbers/main.yaml#L11](https://github.com/ipython-contrib/jupyter_contrib_nbextensions/blob/3f1122c28723e4f726596dfc153c1976c29339e6/src/jupyter_contrib_nbextensions/nbextensions/toggle_all_line_numbers/main.yaml#L11))

jfbercher:
@EWouters The `2to3` extension has been included as a plugin of the `KernelExecOnCells` shared library we designed while improving `code_prettify` and integrating other extensions using the same functional model. See #837. Thanks for triggering that!

PR #820, opened 2016-12-15
Hinterland update - disable hints in comments (configurable)

### LINKED ISSUE #819: Hinterland: Disable automatic hinting when typing a comment
So the automatic hinting is awesome , but there is no need for it when we are typing comments , as it is mildly annoying. 

So I think autocomplete when commenting should be disabled

Thanks

jcb91:
fixes #819


jcb91:
this PR also makes the hint-request timeout configurable, and changes it to 20ms from 200, as the lag was noticeable and annoyed me.

Issue #822, opened 2016-12-22
Collapsible headings do not show up anymore


foice:
Hello, I have used collapsible headings for sometime now, but as of today they not show up any more in my notebooks. I have tried several times to switch off and back on this extension with the http://localhost:8888/nbextensions page, but it does not seem to help. I have also tried on Chrome and Firefox and in both cases the handles to close the heading for the section do not show up.

Any idea what else I could try to get the collapsible headings back?

jcb91:
Well, it sounds like the extension's attempting, but failing, to load. Could you post any output from the 
[javascript console](http://webmasters.stackexchange.com/questions/8525/how-to-open-the-javascript-console-in-different-browsers)? I'd expect to see a complaint about failing to load `collapsble_headings`.

As to what's causing it, what changes have you recently made to your notebook environment? Any new nbextensions/server extensions? Updated versions? Custom js?

Finally, does this apply equally to existing and new notebooks? I would guess yes, but worth a check...

foice:
Failed to load extensions: <unavailable> <unavailable>  main.min.js:12156
GET 
http://localhost:8888/nbextensions/widgets/notebook/js/extension.js [HTTP/1.1 404 Not Found 20ms]
Session: kernel_created (57e00ef1-02ff-413e-8959-ede33a3978ca)  main.min.js:23423
Starting WebSockets: ws://localhost:8888/api/kernels/75e48fb9-ba45-4591-8881-9d1bc5a66646  main.min.js:22709
Widgets are not available.  Please install widgetsnbextension or ipywidgets 4.0  main.min.js:33966
Kernel: kernel_connected (75e48fb9-ba45-4591-8881-9d1bc5a66646)  main.min.js:22356
init_cell : running all initialization cells  main.js:26
init_cell : finished running 48 initialization cells  main.js:36
Kernel: kernel_ready (75e48fb9-ba45-4591-8881-9d1bc5a66646)  main.min.js:22356

foice:
I have widgetsnbextension and ipywidgets (5.1.5)

jcb91:
>It works no problem with small notebooks, is very consistent. For large notebooks I get "script is too slow" warning, which I accept to keep going, and almost always I get no collapsible headings.

[...]

PR #823, opened 2016-12-28
[collapsible_headings] rate limit heading updates

jcb91:
to address #822


jcb91:
closing this in light of the fact that it isn't what's causing the problem in #822, and similar updates coming from work on making collapsible_headings work in exported html notebooks

PR #826, opened 2016-12-31
New notebook extension: code snippets

adrn:
This PR adds a new notebook extension for inserting code snippet cells into the current notebook. When enabled, the extension inserts a drop-down menu into the toolbar with a list of named code snippets specified by the user in `nbextensions/snippets/snippets.json`. The extension ships with an example snippet.

This is based off of an [old extension of mine](https://github.com/adrn/macro-cell).


jcb91:
nice :) reminds me of https://github.com/moble/jupyter_boilerplate, which I've always thought we should add, but never got round to asking @moble about it

src/jupyter_contrib_nbextensions/nbextensions/snippets/main.js
```diff
@@ -0,0 +1,72 @@
...
+                                             .attr("class", "form-control select-xs")
+                                             .change(insert_cell);
+        Jupyter.toolbar.element.append(dropdown);
+    });
+
+    // will be called when the nbextension is loaded
+    function load_extension() {
+        config.load(); // trigger loading config parameters
+
+        $.getJSON("/nbextensions/snippets/snippets.json", function(data) {
```
jcb91: Sorry, should have picked this up before, but this path will fail on anyn server not using the default base_url of `/`, for example on jupyterhub. A better solution would be to use a relative path `./snippets.json`

jcb91: Also, the `$.getJSON` callback relies on the config.loaded already having added the select, so it might be better to chain them directly. I should have mentioned these before merging, really, sorry!

PR #831, opened 2017-01-04
Use util function for upload to add authentication

### LINKED ISSUE #830: Drag-and-drop plugin fails with "'_xrsf' missing from POST"
Platform: OS X 10.12.1, Jupyter 4.2.1, Python 3.6.0, Chrome 55.

Well I really need this function, but it can't work in Safari and Chrome. Here is the log in Javascript Console:

```
[Warning] namespace.js:10 accessing `utils` is deprecated. Use `require("base/js/utils")` 
[Error] PUT http://localhost:8888/api/contents/21.jpg 403 (Forbidden)
[Log] Data transfer for drag-and-drop failed.
```

And the log in Jupyter:

```
[I 17:15:31.188 NotebookApp] Uploading file to
[W 17:15:31.206 NotebookApp] 403 PUT /api/contents/21.jpg (::1): '_xsrf' argument missing from POST
[W 17:15:31.208 NotebookApp] 403 PUT /api/contents/21.jpg (::1) 4.24ms referer=http://localhost:8888/notebooks/test.ipynb
```

It seems that the server rejected the uploading file. So is there any solution?

juhasch:
Fix dragdrop extension due to security fix https://github.com/jupyter/notebook/pull/2005

Should fix #830

### DIFF
--- a/src/jupyter_contrib_nbextensions/nbextensions/dragdrop/main.js
+++ b/src/jupyter_contrib_nbextensions/nbextensions/dragdrop/main.js
@@ -49,10 +49,10 @@ define([
             type : "PUT",
             data: data,
             contentType: 'application/json',
-            dataType : "json",
+            dataType : "json"
         };
-    IPython.utils.promising_ajax(IPython.contents.api_url(path), settings);
-    }
+    utils.promising_ajax(IPython.contents.api_url(path), settings);
+    };
 
     var send_to_server = function(name, msg) {
         var path = utils.url_path_join(utils.url_path_split(IPython.notebook.notebook_path)[0], params.subdirectory);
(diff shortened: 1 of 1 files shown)


src/jupyter_contrib_nbextensions/nbextensions/dragdrop/main.js
```diff
@@ -73,15 +73,14 @@ define([
             data : JSON.stringify(data),
             headers : {'Content-Type': 'text/plain'},
             async : false,
-            success : function (data, status, xhr) {
+            error : function() {console.log('Data transfer for drag-and-drop failed.'); }
```
jcb91: This will get ignored - see [notebook/static/base/js/utils.js#L718](https://github.com/jupyter/notebook/blob/4727efcaf8f8993715831358c1e6b23dad607753/notebook/static/base/js/utils.js#L718)

jcb91: I think you'll want instead a

```javascript
utils.promising_ajax(url, settings).then(
    function on_success (data, status, xhr) {/*...*/},
    function on_error (reason) {/*...*/}
);
```

pattern

juhasch:
I think I only copied the usage pattern from somewhere else. But you are right and I also should add the error path that is unused now.

Issue #740, opened 2016-09-09
hide_input Need better documentation for using with nbconvert


travis23:
I am trying to get hidden input cells to persist with nbconvert --to html. The documentation in readme could use more detail. As it is, I can't get to where I need to go. 

The readme for hide_input states 
`To export a notebook with hidden cells using nbconvert, you need to add a custom template and a custom filter:`

* It would be nice to have and example of the custom template. 
* Can I put the custom template in the same location as the .ipynb file?
* Where is the custom filter placed? Does it go in the template or in the jupyter notebook? 
* Can you provide a sample nbconvert command that utilizes the custom template?

davidgasquez:
Same here! Would love some updated explanations regarding the template usage. :smile:

juhasch:
We now hav docs at http://jupyter-contrib-nbextensions.readthedocs.io/en/latest

I will try to add some more documentation there.

davidgasquez:
Hey @juhasch, I've been able to use the template if I place all in one folder (`template.tpl`, `config.py`, `notebook.ipynb` and `filter.py`) and run `jupyter-nbconvert --config config.py`. I think what is a bit confusing for users right now is where to place all those files to make it system wide.

Thanks for sharing the docs!

jcb91:
@davidgasquez 

> I think what is a bit confusing for users right now is where to place all those files to make it system wide

the main install does attempt to alter nbconvert config to include the package's templates directory in [install.py#L149-L152](https://github.com/ipython-contrib/jupyter_contrib_nbextensions/blob/9bf33fdea280e24298ebf87676521361ee977b98/src/jupyter_contrib_nbextensions/install.py#L149-L152), but it's possible that that either didn't happen correctly, or isn't being read.

There's another subtle complication, that if you specify a config with a different filename (e.g. `config.py` in your example), this changes the default config name that jupyter looks for. That is to say, it will look in the normal jupyter config path as usual, but for files named `config.json` or `config.py` instead of the usual `jupyter_nbconvert.json` and `jupyter_nbconvert.py`, which are the ones that our install procedure implements. Which exact file gets altered depends on which of the `--user`, `--sys-prefix` or `--system` flags you pass to the

    jupyter contrib nbextensions install

command.

If you want to add your own templates system-wide, you can put them all in a folder (let's say `/user/josh/nbconvert_stuff/templates` for example), then edit a config file using a python snippet:

```python
[...]

PR #833, opened 2017-01-05
Readme and docs

jcb91:
to address #740 and comments therein


juhasch:
Looks good.
Note:
In future there will be https://github.com/jupyter/nbconvert/pull/500 as alternative to `strip_output_prompt`.

jcb91:
> In future there will be jupyter/nbconvert#500 as alternative to strip_output_prompt.

Good to know, perhaps we should note it in the docs?

PR #834, opened 2017-01-05
fix init cell extension bug on trust warning

lspvic:
When load a Jupyter notebook, the trust warning model dialog always displayed and init cell are not executed although I am very sure that the notebook is trusted.
I debugged the script, found that the notebook's `trust` property was not initiated with `null` when checking, as below:

this is because the init_cell extension is loaded before notebook is fully loaded. I found when the notebook is big or the network is slow, the bug occurs, so I do a check on whether the notebook is loaded, if not, binding an  `notebook_loaded` event to execute the init run.


jcb91:
Good spot, neat fix, thanks @lspvic!

PR #836, opened 2017-01-06
Remove strip_output_prompt

juhasch:
Use updated template from https://github.com/jupyter/nbconvert/pull/500


jcb91:
looks like the import in https://travis-ci.org/ipython-contrib/jupyter_contrib_nbextensions/jobs/189658004#L2118 needs removing, along with the relevant tests & edit to the conda recipe import-test

juhasch:
Yep, some things were still missing.

src/jupyter_contrib_nbextensions/nbextensions/hide_input_all/readme.md
```diff
@@ -18,3 +18,6 @@ Internals
 
 The codecell hiding state is stored in the metadata `IPython.notebook.metadata.hide_input`.
 If it is set to `true`, all codecells will be hidden on reload.
+
+Exporting using `nbconvert` is currently not supported.
```
jcb91: I think the same metadata value is set as in hide_input, in which case the same export advice as there should work? Or perhaps I've misunderstood/misremembered

juhasch: It currently does not work. It should, so I am adding this now.

jcb91: Right, apologies!

PR #835, opened 2017-01-06
merge code_prettify updates from upstream

jcb91:
as discussed in https://github.com/jfbercher/code_prettify/issues/7, this PR merges the history from the upstream code prettify repo https://github.com/jfbercher/code_prettify

Issue #832, opened 2017-01-04
Add yapf as a dependency?


jcb91:
Migrated from https://github.com/conda-forge/jupyter_nbextensions_configurator-feedstock/issues/10#issue-198835802, @michaelaye commented:

> I'm always met with the "yapf not installed" error when using the code prettyfier. Should it maybe made a dependency for this package?

jcb91:
@jcb91 replied:

> mm, I suppose it could be. I'd hesitate to add it explicitly, but I guess the notebook runs on python anyway. Also, @jfbercher & I have recently been adapting code_prettify to be more configurable, so yapf isn't necessarily a dependency at all anymore for someone using a different formatter. Not sure, thoughts @juhasch, @jfbercher?

juhasch:
I think if `yapf` is the default formatter, we should add it as dependency. If not, I think it will be sufficient to describe the dependencies in the configuration options.

We already have a growing list of dependencies, so I guess it would be good to find a way to have a way to limit this growth. This is also true e.g. for the latex_envs and highlighter extensions, that are listed as dependency.

jcb91:
> I think if `yapf` is the default formatter, we should add it as dependency.

It remains the default, even after the refactor, so I guess we'll add it.

> it would be good to find a way to have a way to limit this growth

True, though  can't think of an obvious way to do that given that essentially we're attempting to provide a whole collection at once. Attempting to provide a selection, we'd end up building something of a package manager, which sounds a bit complicated for my taste 😟 

Anyway, for now, we add `yapf`, it's pretty lightweight anyway.

jfbercher:
I am inclined to agree with @juhasch  that we should not add it to the list of dependencies, because (a) it is only used by `code_prettifier`and we should not install a dependency for every user that is only used by one extension, that some/many users may not select in the collection (b) the dependency is clearly stated in the documentation. I will try to look if it is possible to more clearly alert the user of the missing dependency in case he uses the extension without reading the doc.

juhasch:
We definitely don't want to build a package manager.

The reason I came up with this is that I wanted to build conda packages locally. I failed because of the dependency on the highlighter extension, which failed due to a Windows vs. Linux path issue.

jfbercher:
Perhaps `jupyter_highlight_selected_word` instead of highlighter`?

jcb91:
[...]

PR #837, opened 2017-01-07
merge upstream code-prettify updates

jcb91:
including their original history. This is the hopefully fixed version of #835.


jcb91:
This also removes the need for #818, and includes the updates mentioned in #832

jfbercher:
Just to add some details on this PR :

The PR builds on `code_prettify` extension, including major improvements by @jcb91 -- eg
- json exchange model for messages with the kernel (removing the need for sometimes adhoc regex replacements)
- support for multicell selections (enabling to transform the whole notebook in one go, for instance). 

The PR defines a more robust  architecture and provides a general kernel-language agnostic library, [KernelExecOnCells](https://github.com/jfbercher/code_prettify/blob/master/README.md) for creating Jupyter nbextensions which transform code cell text using calls to the active kernel. The library uses a series of options, describing the configuration of the plugin -- almost everything is configurable!  Default values for these options are specified as an object in the plugin source file, and can be overriden by values loaded from config.

This scheme has been applied to create several nbextensions which are also included in the PR. For instance, to prettify code, see the [code-prettify](https://github.com/jfbercher/code_prettify/blob/master/README_code_prettify.md) nbextension, the [jupyter-autopep8](https://github.com/jfbercher/code_prettify/blob/master/README_autopep8.md) nbextension or to refactor python 2 code for python 3, see the [2to3](https://github.com/jfbercher/code_prettify/blob/master/README_2to3.md) extension, following a suggestion by @EWouters. These nbextensions are defined as simple plugins of the main KernelExecOnCells library.

juhasch:
Thanks :+1:

PR #838, opened 2017-01-07
nbTranslate

jfbercher:
# nbTranslate -- helps translating notebooks -- provides multilanguage support

*Suggestions for a better name would be very welcome, as well as general improvements directions*

This extension

- helps converting markdown cells in a notebook from a language to another (optionally using **Google translate**),
- enables to selectively display cells from a given language in a multilanguage notebook. 

Basically, the extension allows to copy the original cell into a new one for editing and translating. Optionally, the cell source text can be passed through `google translate` and the result inserted in the new cell. Basic markdown structures (e.g. bold, emphasis, lists) are preserved/restored after conversion, to the best extent, but this is not perfect, and usually one has to correct the text and structures afterward. Similarly, equations are extracted before conversion and restored in the result.

The extension has been written to play nicely with
- latex_envs (must be upgraded to version >= 1.3.6)
-  toc2

An **exporter** is provided  to extract one language from the multilanguage notebook. 

--- 
demo1

---
demo2

#### See the README [here](https://github.com/jfbercher/jupyter_nbTranslate/blob/master/README.md) for further details.


juhasch:
I just tried it. Very nice!
On thing I noticed is that in markdown cell this
```
<img src="images/dashboard_files_tab.png" width="791px"/>
```
gets translated to
```
\u003cImg src = "images / dashboard _dateien_ tab.png" width = "791px" /\u003e
```

jfbercher:
Thanks @juhasch. That is right. Html structures are also translated! I have added a small fix for that,  similar to what is done for maths: remove the structures before conversion, put marks at their original places,  and reinsert html after conversion at these places. 
This is not always perfect since sometimes translation changes  elements order and then the html structures can be mixed up; eg `<something> text </something>` may become  `</something<something> text` in some situations. A more complete approach would need to implement something like an html parser and call the translation for the html data.  But a bit complicated for this simple app. 

[*I would not be surprised if there exists a switch to instruct google translate to ignore html structures. But I have not be able to find it..*]

juhasch:
I think not always being perfect is OK. The `` tags are frequently used in notebooks, so it makes sense to handle them.

jcb91:
neat!

> I think not always being perfect is OK

👍 Yes, me too. If you want it to work reliably on everything, I think it'd be possible by extending a marked parser/lexer/renderer, but it'd certainly be a non-trivial amount of work! Regex works perfectly well enough most of the time, particularly as it's a non-destructive operation where the original text is retained.

The build failure in the conda_env travis case is a result of the updated jupyter_latex_envs requirment version - you'll need to update the [conda-forge/jupyter_latex_envs-feedstock](https://github.com/conda-forge/jupyter_latex_envs-feedstock) recipe to have the newer version made available from conda-forge. Incidentally, this is the reason I put the conda_recipe environment on travis, so that we'd spot things like this that break the conda build that otherwise I know I'd just miss at least half the time. For updating the feedstock recipe, see https://github.com/conda-forge/jupyter_highlight_selected_word-feedstock/pull/3 as an example. Essentially, unless requirements or imports have changed, you can just update version number & checksum (which I get from pypi.io's download files, to make sure the version I'm referencing is available there, but could equally be done locally on the sdist tar.gz). See [the instructions in the readme](https://github.com/conda-forge/jupyter_latex_envs-feedstock/tree/9a28e43dce0f124d50892b317217d08ae1a45b5b#updating-jupyter_latex_envs-feedstock) for an important point - make sure to fork the feedstock & submit a PR, as *anything* pushed directly to the conda-forge repo gets uploaded to conda-forge channel. Let me know if you have problems 👌

jfbercher:
I have updated the conda-forge/jupyter_latex_envs-feedstock recipe. The tests for nbTranslate now pass.

jcb91:
Looks good to me!

Issue #646, opened 2016-06-26
Maximum level of nested sections to display in side bar


Kevin-McIsaac:
Firstly, what a brilliant extension, thank you!  

I'm finding that  the "Maximum level of nested sections to display on the tables of contents" setting is not respected by the side bar toc. Not sure if this is a feature or not. My preference would be to either 
1. Have it followed by the side bar
2. Perhaps add the capability of hiding specific headings

jfbercher:
Thanks. 
I do not have this problem, if I understand correctly the question. The 'Maximum level of nested sections' is used at creation of the notebook but is not updated for the notebook at hand if it already exists. If you want to change this depth, edit the notebook-metada in Edit menu and change the number for "toc", "threeshold". For now it is the only way to do that.

Cartman0:
I also found.
Maybe same as jfbercher's.

I found it in created old extension(IPython-contrib<0.2)  notebook file.

changed  from `"toc_threshold": "6";` to `"threshold": "6";`  ,
and delete old property, then solve it.

jfbercher:
I may have finally understood @Kevin-McIsaac's  initial request and pushed a PR #846 for it. This should also avoid issue like the one reported by @Cartman0. 
The drawback is that you can't have some notebooks with a given toc depth and some others with another value.

PR #846, opened 2017-01-19
[toc2] - always read threshold parameter from system config - address #646

jfbercher:
The threshold parameter is now always read from system config. This allows to change the toc depth through the nbextensions_configurator, even for existing notebooks. Hope that this address #646.


jcb91:
Looks fine to me, good to merge?

jfbercher:
ok!

PR #853, opened 2017-01-24
Only use H1 header once

juhasch:


jcb91:
Now that I look again, this is actually already covered by #839, I think? Not sure what the build failure here is, seems to be some conda recipe dependency again, so nothing to do with the actual changes...

juhasch:
Yes, #839 seems to cover it, so closing this one.

PR #854, opened 2017-01-26
[toc2] Implement table of figures

ptitdoc:
Table of Figures can be enabled by selecting a dom_search_pattern (such as :caption), or a specific user defined class (such as .figure).
The code is also adapted to implement multiple table of figures in the future.

Currently, the table of figure items are added to the TOC panel, and also optionally generate a new Figures TOC cell in the beginning of the Notebook.

Finally if the user defined custom styles (using class attributes), the style will also be used for the TOC item.


jfbercher:
Thanks for that! 
I already looked at the code and still need to give it a try. 
I am not sure to exactly follow the use case. You insert figures with `<figure>`and `<caption>`tags? Can you provide an example notebook (either in a gist or as an example for the extension)?  Also I am not sure if we cannot use the more generic parameters and implementation you provide to design some kind of plugin interface to the main toc.

ptitdoc:
Hello, about the use cases:
- a generic use case would be to document figures by adding caption tags `<caption>Graph showing something nice</caption>`. This will numberize and make a summary table of all captions.
- my own use case is to add recommendations to a notebook. I add html tags such as `<p class=info>` `<p class=warn>` `<p class=reco>` along with a custom css file to highligth these informations. In this case, in dom_search_pattern, I use ".info .warn". In such case, I ends with a table of recommendations that summarize all my analysis of a notebook. The custom.css file will also automatically format items in the table.

I will try to share a notebook.

About the plugin interface, do you have use cases ? Because in my view, tables of figures can be used for a lot of cases ('figures' is just a name that can be changed by the way). Plus I doubt the stability that can be achieved using plugins of plugins.

ptitdoc:
One point however is that in my current code only a single table of figure is allowed.
However I tried organized/generalized the code to easily add multiple table. The thing is, I don't know how to deal with configuration parameters to use an arbitrary number of tables. 

This only way I identified is "create all the parameters for 10 tables of figures", and add a parameter to ask how many table of figures the user want (but limited to 10 in this case).

jfbercher:
@ptitdoc Sorry sorry I didn't have the time to test and review this for now. I will try to do this before then end of next week.

ptitdoc:
No problem.

Just note that I tested it last week with the defaults parameters, and it works with  `<caption>My caption here</caption>` tags.

I will probably add as suggested the parameters for 5 different tables of figures, and a master parameter to select how much table of figures the user wants (0 to disable table of figures to 5). Because for now, it is not possible to disable tables of figures.

ptitdoc:
As suggested, I added a parameter to choose the number of table of figures (default 0), which allows disabling tables of figures.
I will also add in the future an arbitrary number of table of figures (probably 5) that can be parametrized.

ptitdoc:
Replaced with a reworked pull request:

https://github.com/ipython-contrib/jupyter_contrib_nbextensions/pull/1307

Issue #859, opened 2017-01-27
'Move selected cells' shortcuts not working on mac os 


ggrrll:
Hi, 

the shortcuts  of 'Move selected cells' are not working for me
( Alt-up and Alt-down )

I am on:
- macOS 10.12.2
- notebook 4.3.1

Thanks for your support

jcb91:
Not sure why this is, you could try checking the [javascript console](http://webmasters.stackexchange.com/questions/8525/how-to-open-the-javascript-console-in-different-browsers) for any error/warning messages.

Alternatively, I use the keyboard_shortcut_editor to bind the move cell up & move cell down actions to ctrl-up and ctrl-down. These actions work with multi-cell selections, so provide all of the functionality of this nbextension, at leats in the latest notebook version.

jcb91:
Perhaps I should add to the readme to note this...

jfbercher:
Ooops; this is an old extension that may have been deprecated. It should be very easy to update it for notebook >=4.2 using the `Jupyter.notebook.move_selection_up` (or `down`) actions.

jcb91:
Also, have you tried on any other OS? It's possible that the bindings are being intercepted by the browser...

jfbercher:
Deprecate or not?

jfbercher:
Th extension fails because `get_selected_indices` was replaced by `get_selected_cells_indices`

jcb91:
Ah, good diagnosis! Well, we could keep it for those on older notebook versions, as long as we make it work also with newer versions (preferably using the action). Additionally, if we keep it, then we should update the readme to reflect the fact that it exists in core/can be bound as a shortcut using the shortcut editor...

ggrrll:
(same negative result on safari and firefox)

jcb91:
Yep, this will be broken on all browsers, since, as @jfbercher's quick diagnosis reveals, it's the result of the notebook javascript API having changed a function name. The quickest way to get the functionality is to enable the `keyboard_shortuct_editor`, then add new command-mode shortcuts from the `Help->keyboard shortcuts` menu item

jfbercher:
I will update the extension this evening.

jcb91:
should be fixed by #860

PR #860, opened 2017-01-27
[move_selected_cells] updated to Jupyter 4.2+

jfbercher:
As discussed in #859.


jcb91:
look good, only I'd make rubberband a requirement only for versions of notebook which don't already support multiple-selection.

jcb91:
that last commit seems to have reverted the readme?

jcb91:
looks good to me :+1:

Issue #828, opened 2017-01-04
 Link error: Error: post-link failed for: conda-forge::jupyter_nbextensions_configurator-0.2.3-py27_0


yunyundong:
I want to install nbextensions for jupyte notebook, I used the command 

    conda install -c conda-forge jupyter_contrib_nbextensions

but it failed. there was a message:

    LinkError: Link error: Error: post-link failed for: conda-forge::jupyter_nbextensions_configurator-0.2.3-py27_0

besieds, I used pip to install:

    pip install https://github.com/ipython-contrib/jupyter_contrib_nbextensions/tarball/master
    jupyter contrib nbextension install --system

in the process of install, there is no error messages. but after i restart the jupyte server, there is a 'nbextension' tab ,but it is empty, no extensions to be enabled. how can i do? by the way ,my anconada version is 4.2

by the way ,how can I uninstall the nbextensions clean? including remove the install packages.
Thanks in advance.

jcb91:
> there is a 'nbextension' tab ,but it is empty

Yes, this is provided not by this repo directly, but by the [jupyter_nbextensions_configurator](https://github.com/Jupyter-contrib/jupyter_nbextensions_configurator) package, which gets installed as a dependency. Using `pip uninstall <package>` doesn't remove packages installed as dependencies of `<package>` (see [this SO answer](http://stackoverflow.com/a/27713702)).

Therefore, to remove things completely, you'll have to do a little more work, and uninstall each of the dependencies which you no longer require. You could use the [pip-autoremove](https://github.com/invl/pip-autoremove) package, or use a [list of the dependencies as of today](https://github.com/ipython-contrib/jupyter_contrib_nbextensions/blob/8946d0779fb54165c3b466bda0c07650c7105166/setup.py#L64-L75). However, you probably want to keep some of the dependencies, as they'll be required by the notebook itself.

I'd recommend

    pip uninstall jupyter-contrib-core jupyter-highlight-selected-word jupyter-latex-envs jupyter-nbextensions-configurator psutil

To begin with

tarokiritani:
I have got a similar error when installing this repo.

I did
`conda install jupyter_contrib_nbextensions -c conda-forge`

and got this error:
`LinkError: Link error: Error: post-link failed for : conda-forge::jupyter_latex_envs-1.3.6-py35_0`

What I found a bit odd is I already have "jupyter-latex-envs" package installed on my environment.

tt293:
[...]

PR #874, opened 2017-02-05
[requirements] nbTranslate needs traitlets 4.1

jcb91:


jcb91:
(for the tag api, similar to what we've previously found in #828)

jfbercher:
Thanks

PR #882, opened 2017-02-11
Fix/limit output2

juhasch:
Depending on kernel type, limiting different message types is important.
For Python, `stream` type messages are the most common use case.

### DIFF
--- a/src/jupyter_contrib_nbextensions/nbextensions/limit_output/readme.md
+++ b/src/jupyter_contrib_nbextensions/nbextensions/limit_output/readme.md
@@ -38,12 +38,11 @@ The `limit_output_message` parameter can be formatted to display the
 `limit_output` length and the current `output_length`, using the respective
 replacement fields `{limit_output_length}` and `{output_length}`.
 
+### Parameter Overview
 
-Parameter | Type | Description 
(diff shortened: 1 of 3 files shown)


src/jupyter_contrib_nbextensions/nbextensions/limit_output/readme.md
```diff
@@ -37,3 +37,13 @@ counted, and if either exceeds `limit_output` characters will be truncated to
 The `limit_output_message` parameter can be formatted to display the
 `limit_output` length and the current `output_length`, using the respective
 replacement fields `{limit_output_length}` and `{output_length}`.
+
+
+Parameter | Type | Description
```
jcb91: Unfortunately table markdown isn't supported by the recommonmark parser which is currently producing the readthedocs pages. I've used bullet lists instead, although they aren't really as good as a table would be

jcb91:
nice!

src/jupyter_contrib_nbextensions/nbextensions/limit_output/readme.md
juhasch: Thanks for letting me know. Fixed it.

Issue #879, opened 2017-02-09
freeze state of cells is not persisted when reloading notebook or refreshing page (browser refresh)


odedbd:
The freeze extension answers a very real use case for me, but unfortunately it doesn't seem to persist the cell's status after closing the notebook and reopening it, or even when simply refreshing the page. This seems strange as the documentation indicates that the state is saved in the cell's metadata and should be preserved.

Moreover, I used the Edit cell metadata option to verify that the cell does indeed have in its metadata the settings that it should be frozen, namely-
```
 {
  "scrolled": true,
  "deletable": true,
  "collapsed": false,
  "hide_input": false,
  "editable": true,
  "run_control": {
    "frozen": true,
    "read_only": true
  },
  "trusted": true
}
```

Yet it is not colored as frozen and indeed I can modify its contents after refreshing the page. Setting it as frozen again will work until I refresh the page again.

jfbercher:
What other extensions do you use? 
Can you disable all the other extensions to see if there is not a conflict with another one?

odedbd:
Good idea. I have a quite a few extensions enabled, but this is on my work computer so I will only be able to check the list and test for a conflict on Sunday (weekend where I live is Friday-Saturday).

Thanks!

jcb91:
Pinging @kukanya as  freeze's author. This may also be simply related to the notebook being saved or not. In order for metadata changes to be correctly interpreted as changes to the notebook, the notebook's state should be designated 'dirty' (see [notebook/static/notebook/js/notebook.js#L385-L396](https://github.com/jupyter/notebook/blob/bfb47641e049b49bf82f4c6269fa6b9d8617c18f/notebook/static/notebook/js/notebook.js)). If you freeze a cell, then directly refresh the page, the chances are that the metadata changes will not have been saved, but the notebook isn't set as dirty, so won't warn you about leaving the page. As a result, on reloading, the cell won't be frozen anymore. You can check whether this is the cause of your issue just by making sure to save the notebook before refreshing the page. Actually, it could be worth setting the notebook as dirty on freezing/unfreezing cells to avoid this kind of problem, regardless of whether it's actually causing this particular issue.

kukanya:
@jcb91, thank you for making me aware.

I'll try to reproduce the issue this weekend and work on a fix if it's reproducible for me.

And yes, the extension is not designed to apply changes you do directly to the metadata. I guess I'll work on that if I have time, too.

[...]

PR #883, opened 2017-02-12
make execute_time robust to loading before notebook finishes loading

jcb91:


jcb91:
this should fix the timing issue described in #879 for the execute_time nbextension

PR #884, opened 2017-02-12
Freeze Update

### LINKED ISSUE #879: freeze state of cells is not persisted when reloading notebook or refreshing page (browser refresh)

kukanya:
Hopefully fixes #879.

### DIFF
--- a/src/jupyter_contrib_nbextensions/nbextensions/freeze/main.js
+++ b/src/jupyter_contrib_nbextensions/nbextensions/freeze/main.js
@@ -139,9 +139,9 @@ define([
             }
         ]);
 
-        if (typeof Jupyter.notebook === "undefined") {
-            events.on("notebook_loaded.Notebook", initialize_states);
-        } else {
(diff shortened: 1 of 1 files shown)


jcb91:
looks good to me, thanks :+1:

Issue #871, opened 2017-02-01
ToC2: cell sizes are widened 


RinaldoB:
Hi, turning on the Table of Contents extension has an side-effect for me. The width of the cells is extended to fit the browser window. Without ToC2 the width of the cells is limited beyond a certain size.

jfbercher:
This is not really an issue: this behavior is intended, so to have the sidebar to display nicely. 

We could add another configuration option for that if people find it sufficiently useful but this will further complicate the code (have to test if the toc2 is displayed or not, floating or displayed as a sidebar).

RinaldoB:
I support the option to widen the cells only if the ToC is visible.

The current behavior alters the notebook display, only if the ToC2 extension is enabled. Personally I intend to use a ToC only a couple of my notebooks, however all of the notebooks are affected by this, so I'm forced to turn the extension off.

odedbd:
I find the widening effect generally helpful. If it ends up being applied only when the ToC is visible it would be great if "wide_notebook" could be available as a standalone extension

jcb91:
yep, making a standalone is certainly possible, and sounds reasonable

jfbercher:
An attempt to address this is on #887. 

For a standalone extension, something like this
`$("#view_menu").append("<li id=\"toggle_width\" title=\"Enlarge notebook\"><a href=\"javascript:{var percent = $('#notebook-container').width()/$('#notebook-container').parent().width()*100; var new_width = percent < 90? '98%': '82%'; $('#notebook-container').css({'width':new_width, 'margin-left':'auto'})}\">(Un)stretch notebook</a></li>")` in `custom.js` should do the trick.

jcb91:
or even just

```css
#notebook-container {
    width: auto;
}
```

in `custom.css`!

KadeG:
Just spent a bit tracking down what was widening my cells. I wish it mentioned it in the readme. It makes for difficult reading in text-heavy notebooks.

jfbercher:
@KadeG  See #887. In addition I just updated the README :-)

jfbercher:
Closed by #887.

KadeG:
@jfbercher You're awesome!

fkjogu:
For me, the cells are still widened even if I have switched this off in the options.

Using jupyter 4.3.0, Python 3, and whatever is current in PIP.

fkjogu:
[...]

PR #887, opened 2017-02-13
[toc2] Add a parameter to enable/disable cell widening

jfbercher:
This addresses #871. 
A parameter is added to enable/disable cell widening to fit the browser window (parameter is set to false to be consistent with the previous behavior). This works independently of the state of the toc display (sidebar/floating/hidden). Loading of config parameters have been a bit cleaned. Also reworked the code so that everything (I hope)  resizes correctly with the main browser window.

### DIFF
--- a/src/jupyter_contrib_nbextensions/nbextensions/toc2/README.md
+++ b/src/jupyter_contrib_nbextensions/nbextensions/toc2/README.md
@@ -2,7 +2,7 @@
 
 ## Description and main features
 
-The toc2 extension enables to collect all running headers and display them in a floating window, as a sidebar or with a navigation menu. The extension is also draggable, resizable, collapsable, dockable and features automatic numerotation with unique links ids, and an optional toc cell. Sections of currently selected/edited or running cells are highlighted in the toc. Finally, the toc can preserved when exporting to html.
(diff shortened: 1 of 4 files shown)


jcb91:
looks like a good plan to me, but I'll leave merging for to you, for when you're happy with it @jfbercher :)

jat255:
Even when disabling the new widen option, I find that the top menu items and the body of the notebook are still widened. The menu bar takes up the full width, and the body cells now take up 82% of the width (compared to a fixed 1140px width without it). I would like to use this extension, but on my widescreen monitor it is difficult to read at full (or 82% width). Is it feasible to completely disable the width scaling, but keep the ToC feature? 

In my opinion, the docking is nice, but a Table of Contents extension should not completely reformat the whole page. I had to selectively disable all of my extensions to figure out which one was the culprit, and I definitely did not expect it to be this one!

Otherwise, I very much like this extension 😄

jfbercher:
I am not sure to understand well. For the menu & menubar, you can disable moving on the left by disabling the parameter "Move notebook's title and menu on the left instead of being centered". This feature was recommended/requested by others users. As for the 82% width, I used it to look like the default; isn't it the case? Anyway we could let it configurable.

jat255:
You're right about the menu bar. I simply missed that option. All is well when it is unchecked.

As for the width, I would agree that configurable would be the most versatile and satisfy the majority of users. I have a fairly wide monitor (1920px), so my 82% is probably wider than someone on a 1600 or 1200px screen.

jcb91:
I think the default is actually fixed at several steps (the standard bootstrap small/medium/large screen stops), rather than a percentage...

jfbercher:
@jcb91 I don't know all that, but it seems indeed that the best is to reset the width by a `.css({'width': ""})` to return to the default behavior. I will update that.

Issue #847, opened 2017-01-20
nbextensions and nbconvert


tarokiritani:
Is it possible to use nbextensions with nbconvert? I ran `jupyter nbconvert --to html nynotebook.ipynb`. Then, toc2 seems to be effective in the resulting html. However, collapsible headings is no longer working.

Based on the [previous issue #198](https://github.com/ipython-contrib/jupyter_contrib_nbextensions/issues/198), I understand that in order to use nbextensions with nbconvert I need to

1. make "jupyter_nbconvert_config.py" in jupyter configuration folder (in my win 8.1, User\.jupyter),
2. put nbextension .py files in extensions folder,  and
3. I somehow need to write a template file in User\.jupyter\extensions.

Am I mistaken? I have not found collapsible_headings.py in my env, and have not figured out the format of the template file. Could anyone give me pointers?

jcb91:
> Is it possible to use nbextensions with nbconvert?

Some of them support nbconvert, to various degrees. This is down to which have had templates, exporters etc written for them. See [jupyter-contrib-nbextensions.readthedocs.io/en/latest/exporting](http://jupyter-contrib-nbextensions.readthedocs.io/en/latest/exporting.html) for docs on the subject.

> Based on the previous issue #198, I understand that in order to use nbextensions with nbconvert I need to [...] Am I mistaken?


Well, that advice is a little out of date, and a lot of that process is now done automatically by the package installation. Again, refer to the docs at [jupyter-contrib-nbextensions.readthedocs.io/en/latest/exporting](http://jupyter-contrib-nbextensions.readthedocs.io/en/latest/exporting.html) for the latest info.

> I ran jupyter nbconvert --to html nynotebook.ipynb. Then, toc2 seems to be effective in the resulting html. However, collapsible headings is no longer working

I'm not completely familiar with how toc2 might function even without extra options passed to nbconvert, although I know it does provide an exporter for embedding into html output (see http://jupyter-contrib-nbextensions.readthedocs.io/en/latest/exporting.html#export-table-of-contents). The support for `collapsible_headings` exports is currently very limited (essentially, it'll just remove any hidden cells from the output), basically because I haven't written a version which can be embedded yet. Since you sound like you might use it, I'll give it some thought to it & see what I can come up with...

tarokiritani:
[...]

Issue #605, opened 2016-05-13
Collapsible headings: nbcovert notebook


brecht-d-m:
Is it possible to convert a notebook to HTML with collapsible headings still "enabled" in the converted HTML file? I tried to add ```"pre_collapsible_headings.CollapsibleHeadingsPreprocessor"``` to the preprocessors in ```jupyter_nbconvert_config.json```, but this seems not to do it (```jupyter nbconvert notebook_name.ipynb```). 

If it is possible, do I need to change/add other files?

jcb91:
True, I tested the preprocessor recently and it doesn't work as intended.
I've made a fix in my trav branch, but haven't backported it yet. Will aim
to do it soon...

brecht-d-m:
Your fix does indeed fix a part of my issue. What I'm truly looking for, is that the functionality that the collapsible headings extension has in the notebooks, is preserved in the exported HTML file. I mean that I could also collapse and uncollapse parts of the HTML file. Or is this not possible?

jcb91:
Ah, I see! Sorry, I was a bit slow on the uptake there. Hmm. I'm not
completely sure, but I think it's maybe possible, given that jquery is (I
think?) included as a link in the output HTML. I'll have a think and see if
I can patch in some JavaScript through the preprocessor...

jfbercher:
You can probably do this with a template like I suggested in #600 for collapsing outputs (jquery is included in full.tpl).
It will be nice :-)

jcb91:
Ah yes, I knew I'd seen a reference to something similar recently somewhere
:)

brecht-d-m:
A template did indeed solved it. If somebody is interested, I have put the code [here (template)](https://gist.github.com/brecht-d-m/01529f956bb90032508ff439f466325d) and [here (javascript file)](https://gist.github.com/brecht-d-m/88f5a989a1818191277852559f4c2b0c)

jcb91:
@brecht-d-m interesting approach! It certainly seems rather more intuitive to place cells inside each other, though as you note, it does give rise to potential DOM-related issues for other templates/postprocessors/css etc.

PR #892, opened 2017-02-15
[collapsible_headings] new features!

jcb91:
* unbind event handlers during notebook loading, so collapsing/uncollapsing headings don't slow down the load, as mentioned in #822
* allow left-arrow keyboard shortcut to skip to higher-level headings, for faster navigation
* add support for embedding collapsible_headings into nbconverted html documents, as requested in #605, #847. This is done through use of an exporter model which can be applied more generally through subclassing, though at the moment, the only other example that comes to mind of embedding things in html is toc2.


jcb91:
I haven't yet updated any of the docs for this...

jcb91:
or tests! d'oh

jfbercher:
:+1: :+1:

jcb91:
Ok, I'm gonna merge this as-is, maybe add support for exporting alongside toc2 in another PR...

PR #894, opened 2017-02-18
[CI] test newer notebook & python versions

jcb91:


jcb91:
for the record, this adds python 3.6 tests on OSX & linux (Travis), but not Windows (Appveyor don't yet provide python 3.6), along with (python) test runs for notebook 4.3, 4.4 and 5.0.0b1. We still don't have proper javascript tests.

PR #895, opened 2017-02-19
Modularity - don't run things before `load_ipython_extension` is called

jcb91:


jfbercher:
Wow! I was thinking on updating this. So, well done; thanks.

Issue #885, opened 2017-02-12
handling `notebook_loaded.Notebook` events correctly


jcb91:
As discussed in #879, sometimes (particularly with big notebooks, slow connections, etc), nbextensions get loaded before the notebook has finished loading. In other cases, the notebook may have fully loaded before the nbextension gets loaded. Finally, the event is fired again as part of loading checkpoints, so it's not necessarily a one-time-only event! Nbextension code should be able to deal with any of these three scenarios.

The best way to approach this and still get any initialization which needs to be applied to the whole notebook performed correctly seems to be:

1. make sure any whole-notebook initialization function can be called multiple times without causing problems like duplicating html elements, etc.
2. bind the initialization function to the `notebook_loaded.Notebook` event, so that it gets re-initialized correctly whenever the notebook is (re)loaded
3.As part of the `load_ipython_extension` function used to load the nbextension, check if `Jupyter.notebook._fully_loaded === true`, which indicates we missed the first `notebook_loaded.Notebook` event. If this is true, call the initialization function(s) mentioned in 1 directly

I've had a quick skim through our nbextensions looking for ones which may need their loading behaviour altered, and come up with the following lists: 

The following need updates to handle loading correctly:

 - [x] `execute_time` - handled in #883
 - [ ] `freeze`
 - [ ] `hide_input`
 - [ ] `hide_input_all`
 - [ ] `ruler`
 - [ ] `scroll_down` -ok but needs update to handler for new '.output' instances
 - [ ] `splitcell`

These already use the correct approach:
 - `autoscroll`
 - `collapsible_headings`
 - `init_cell`
 - `nbTranslate`
 - `toc2`

These I'm uncertain about:

 - [ ] `python-markdown` - does some running of things on `kernel_ready.Kernel`, but I'm not sure whether this needs special handling
 - [ ] `runtools` - has some odd dummy require thing going on to combat race conditions. I'm not immediately sure what's going on there
 - [ ] `table_beautifier` - probably ok (relies on catching `kernel_ready.Kernel`), but I'm not certain
 - [ ] `highlighter/highlighter.old.js` - do we even need to keep this file?!
 
These seem to run at least some code without `load_ipython_extension`
being called, which I don't think is supposed to happen, so perhaps need a bit
of reworking:
 - [ ] `codefolding`
 - [ ] `exercise`
 - [ ] `exercise2`
 - [ ] `highlighter/highlighter.js`
[...]

PR #897, opened 2017-02-19
Fix runtools initialization

juhasch:
Fixes `notebook_loaded.Notebook` handling and removes hack for race condition.
See #885 for background.


jcb91:
looks good to me, although the `else` means that the `initGutter` won't be executed when a notebook is e.g. reloaded from a checkpoint - but I guess that since the `create_cell` event should handle those ok, that's correct?

juhasch:
Yes, I think this should be find IMO. We only need to make sure the gutter is there when a notebook is created.

jcb91:
okey doke!

PR #899, opened 2017-02-20
[highlighter] Minor typo update + cleaning 

jfbercher:
- Minor typo update which prevented correct display of readme in nbextensions_configurator
- Cleaned dead highlighter.old.js file


jcb91:
I've coincidentally just patched the configurator to fix the display issue with syntax-highlighting schemes which can't be found, so the ipython highlight scheme will just display unhighlighted, rather than preventing display of the entire readme. However, I haven't yet looked into how to add the ipython syntax (which adds support for magics) to the readme renderer in the same way that it's added for the general notebook...

jfbercher:
@jcb91 Nice that you improve highlighting support in the configurator!
I am merging this PR for now (the highlighted code in question is pure python).

jcb91:
Good plan to merge :+1: configurator update will be a few days while I get testing sorted out, anyway.
But given the `%%javascript` magic, isn't it mainly javascript? :laughing:

jfbercher:
Yes, you are right, and the code in question was definitively ipython! Let it as is for now, and we will update the highlight mode after and specify that the code to test the extension must be typed in a ipython code cell (or in the JS console as pure javascript).

PR #900, opened 2017-02-20
Make sure gutter is loaded before initializing it.

juhasch:
Overlooked that at #897.

### DIFF
--- a/src/jupyter_contrib_nbextensions/nbextensions/runtools/main.js
+++ b/src/jupyter_contrib_nbextensions/nbextensions/runtools/main.js
@@ -6,8 +6,7 @@ define([
     'jquery',
     'require',
     'base/js/events',
-    'codemirror/lib/codemirror',
-    'codemirror/addon/fold/foldgutter'
+    'codemirror/lib/codemirror'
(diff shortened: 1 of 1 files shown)


jcb91:
Is this necessary? The define should only execute once the addon's been loaded anyway, no?

juhasch:
Unfortunately yes. Without, it looks like this:
bildschirmfoto 2017-02-20 um 22 31 03

jcb91:
ok, that's strange!

I think it must be more of a timing problem though, since the `define([` call is analogous to the `require([` one in terms of loading order, so if anything it should load the addon _earlier_. But whatever, if this fixes it for now, it's not worth worrying about :smile:

juhasch:
It is a timing problem, and this is the least ugly hack I could come up with.
The gutter stuff can be strange...
If you have an idea what to do about it, you're welcome. But don't waste your time on it.

jcb91:
> The gutter stuff can be strange...

No kidding!

> If you have an idea what to do about it, you're welcome. But don't waste your time on it.

Yeah, it looks too complicated to quickly think about, so if this ides fixes it for now, I think you've got the right plan here!

PR #902, opened 2017-02-23
Updated regex pattern for better match

jozjan:
Fix to match markdown img tag that contains:
1. space in the filename
2. alt and title

such as:
```markdown
![arduino with a photoresistor](images/arduino with a photoresistor.png "Arduino Physical Layout")
```

that normally renders to:
```html
<img src="images/arduino with a photoresistor.png" alt="arduino with a photoresistor" title="Arduino Physical Layout">
```

### DIFF
--- a/src/jupyter_contrib_nbextensions/nbconvert_support/embedhtml.py
+++ b/src/jupyter_contrib_nbextensions/nbconvert_support/embedhtml.py
@@ -29,7 +29,7 @@ class EmbedHTMLExporter(HTMLExporter):
         if url.startswith('http'):
             data = urlopen(url).read()
         elif url.startswith('data'):
-            img = '<img src="' + url + '" ' + match.group(2) + ' />'
+            img = '<img src="' + url + '"'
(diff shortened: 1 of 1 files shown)


juhasch:
Thanks. Yes, I guess the regex can definitely be improved.
In your approach you removed the ability to add additional parameters like `width=50px`.
Could you add that again ?

jozjan:
Hi @juhasch , can you please give me an example of such situation?

I guess the simplified regex `'<img\s+src="([^"]+)"'` should only match the the `img` tag followed with a space and the `src` attribute with the image uri inside the quotes. No other character inside the `img` tag is being matched. 
If such match is found, the `regex.sub()` will only replace the matched string - thus not touching other possible attributes such as `alt`, `width`, etc. in the remaining part of the `img` tag.
Maybe I am missing something :-/

juhasch:
OK, I took a closer look. You are right, it works as it should.
I was thinking of something like ``

PR #905, opened 2017-03-01
[codemirror_mode_extensions] add new nbextension

jcb91:
This is a small nbextension adding an extension to the codemirror octave mode, because I got frustrated while using an octave kernel from metakernel. I've submitted this extension to codemirror & it's been merged, but it won't make it into notebook for a while, so I'm adding it here for now. We could also add similar extensions for other codemirror modes in future

### DIFF
--- a/src/jupyter_contrib_nbextensions/nbextensions/codemirror_mode_extensions/codemirror_mode_extensions.yaml
+++ b/src/jupyter_contrib_nbextensions/nbextensions/codemirror_mode_extensions/codemirror_mode_extensions.yaml
@@ -1,0 +1,7 @@
+Type: Jupyter Notebook Extension
+Name: CodeMirror mode extensions
+Description: |
+  Extends some CodeMirror modes with extra features. Currently just adds
+  support for comment/uncomment and folding for octave/MATLAB mode.
+Main: main.js
+Compatibility: 4.x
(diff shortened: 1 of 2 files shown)


juhasch:
👍 
Commenting works for me. I can't get codefolding to work with Matlab though.

jcb91:
ok, I didn't look at codefolding yet, just added what I'd found on other codemirror modes. Did you try octave as well, or just MATLAB?

For future reference, the CodeMirror PR is https://github.com/codemirror/CodeMirror/pull/4570

juhasch:
Actually, this is my 'fault'. In the codefolding extension, the folding options are hardcoded currently.
Let me see if this can be changed.

juhasch:
OK, I fixed the codefolding extension and removed the hardcoded folding type setting in PR #908. 

However, for Matlab/Octave I don't think indent folding is correct. This is more of a convention, the language itself doesn't care about indenting.

jcb91:
> However, for Matlab/Octave I don't think indent folding is correct. This is more of a convention, the language itself doesn't care about indenting.

Yes, you're absolutely right, I just added it as a stop-gap, since I didn't have the time to find out how to implement more complex folding in CodeMirror. I think it requires the mode to implement it itself, but I'll have a think.

juhasch:
I think we can merge it.

jcb91:
Ok, I've made a proper fold mode for octave, but haven't managed to get it to load correctly year, so I'll merge as-is for now.

juhasch:
Cool, I look forward to that!

PR #906, opened 2017-03-01
fix loading of skill, remove stale nbextensions no_exec_dunder and nbviewer_theme

jcb91:
* no_exec_dunder functionality is provided by freeze
* nbviewer_theme hasn't been updated for 4.x, and seems to be a subset of css_selector anyway.
* skill is a bare codemirror mode, which doesn't provide the expected load_ipython_extension method, causing the notebook to record a misleading error. Add a main.js script to load the codemirror mode.


juhasch:
👍

PR #911, opened 2017-03-07
Use https instead of http for jquery

piti118:
It's 2017 and using http causes some problem with https hosting as most browser disallow loading script from http source if the resource is https.


jfbercher:
Looks good to me. Thanks.

PR #912, opened 2017-03-08
[init_cell] allow celltoolbar preset to load from metadata setting

jcb91:


jcb91:
The conda builds are currently error-ing on travis, as a result of a failure to install conda-build, which is occurring for reasons I don't yet understand. I'm going to go ahead and merge this anyway

PR #915, opened 2017-03-11
Fix Firefox drag&drop

### LINKED ISSUE #914: Drag and Drop doesn't work in Firefox
Drag and Drop works in IE and Edge but not in Firefox52 on Windows10 64bit latest.  The checkbox in configurator tab is checked and javascript console shows "Loading extension: dragdrop/main", but drag-and-drop of an image file responds nothing.

juhasch:
Fixes #914


jcb91:
as mentioned in #912, the travis conda recipe build is currently failing to start correctly, as a result of failing to install conda-build, for reasons I don't yet understand. At any rate, the reason isn't related to this PR, so I think we can go ahead and merge.

PR #917, opened 2017-03-16
Fix documentation links in snippets

moble:
See moble/jupyter_boilerplate#12


jcb91:
as mentioned in #912 and #915, the travis conda recipe build is currently failing to start correctly, as a result of failing to install conda-build, for reasons I don't yet understand. At any rate, the reason isn't related to this PR, so I think we can go ahead and merge.

PR #919, opened 2017-03-17
Scratchpad

### LINKED ISSUE #916: Tab-complete window appears under scratchpad.
If I open the scratchpad window and try to use the tab complete, it only becomes visible after I close the scratchpad. For example if I do:

```
inspect.<tab>
```

it seems like it doesn't do anything. But if I close the scratchpad, my notebook looks like:

screenshot from 2017-03-15 13-53-26

Is it possible to change the window ordering so that the tab-complete window is visible and usable?

jcb91:
Fixes #916


jcb91:
just the conda build still error-ing, unrelated to anything in this PR

PR #920, opened 2017-03-17
[CI] temporary fix for travis conda build

jcb91:
as they cause travis condarecipe build to fail reporting a signal-releated TypeError. This should hopefully be fixed by conda in future, hence the blacklist rather than using versions prior to 4.3.14 and 2.1.6

[skip Appveyor]


jcb91:
Hmm. I thought this would do it (seems to have worked for the configurator), but it seems there's something more complicated going on here (or possibly another, separate-but-related issue)

PR #922, opened 2017-03-22
[varInspector] A variable inspector.

jfbercher:
# Python Variable Inspector

## Description and main features

The Python Variable Inspector extension enables to collect all defined variables and display them in a floating window. The window not only display the name of variables but also  their type, size in memory and content. The columns are sortable. The window is draggable, resizable, collapsable. The list of displayed variables is automatically updated at each cell execution. Variables can be deleted from workspace by pushing a link. Position and state (displayed/collapsed) are stored in the notebook's metadata and restored at startup. 

#### Demo:
demo

 
## Configuration
The initial configuration can be given using the IPython-contrib nbextensions facility. It includes:

- Display at startup or not (default: false) 
- Width of columns (actually the max number of charracter to display in each column)
- The state of the toc cell (default: false, ie not present)
- The list of types to exclude from display (default: ['module', 'function', 'builtin_function_or_method', 'instance', '_Feature'])

## Notes
- The displayed size of variables use the `getsizeof()` python method. This method doesn't work for all types, so the reported size is to be considered with some caution. The extension includes some code to correctly return the size of numpy arrays, pandas Series and DataFrame but the size for some other types may be incorrect. 
- The extension builds on some code provided [here](https://github.com/ipython/ipywidgets/blob/master/docs/source/examples/Variable%20Inspector.ipynb) (essentially the `_fill` method)

## History

- @jfbercher march 22, 2017 -- initial release


src/jupyter_contrib_nbextensions/nbextensions/pyVariableInspector/main.js
```diff
@@ -0,0 +1,397 @@
...
+
+    var st = {}
+    st.config_loaded = false;
+    st.extension_initialized = false;
+
+    function read_config(cfg, callback) { // read after nb is loaded
+        // create config object to load parameters
+        var base_url = utils.get_body_data("baseUrl");
+        var initial_cfg = $.extend(true, {}, cfg);
+        var config = new configmod.ConfigSection('notebook', { base_url: base_url });
```
jcb91: instead of creating a new section, you actually `Jupyter.notebook.config` if you want the notebook config section.

src/jupyter_contrib_nbextensions/nbextensions/pyVariableInspector/main.js
```diff
@@ -0,0 +1,397 @@
+define(["require", "jquery", "base/js/namespace", 'services/config',
+    'base/js/utils', 'notebook/js/codecell'
+], function(require, $, IPython, configmod, utils, codecell) {
```
jcb91: could we have `Jupyter` instead of `IPython` please?

src/jupyter_contrib_nbextensions/nbextensions/pyVariableInspector/main.js
```diff
@@ -0,0 +1,397 @@
...
+        }
+    );
+
+    function toggleVarInspector() {
+        toggle_varInspector(cfg, st)
+    }
+
+    var varInspector_button = function() {
+        if (!IPython.toolbar) {
+            $([IPython.events]).on("app_initialized.NotebookApp", varInspector_button);
```
jcb91: accessing events from the global namespace is (I think) deprecated - you should add `base/js/events` to the `define([` requirement list, and then call `events.on`

src/jupyter_contrib_nbextensions/nbextensions/pyVariableInspector/main.js
```diff
@@ -0,0 +1,397 @@
...
+        })
+
+        // event: on cell execution, update the list of variables 
+        $([IPython.events]).on('execute.CodeCell', varRefresh);
+        $([IPython.events]).on('varRefresh', varRefresh);
+    }
+
+    var create_varInspector_div = function(cfg, st) {
+        function save_position(){
+            IPython.notebook.metadata.pyVarInspector['varInspector_position'] = {
```
jcb91: you can use `IPython.notebook.metadata.pyVarInspector.varInspector_position`

src/jupyter_contrib_nbextensions/nbextensions/pyVariableInspector/pyVarInspector.yaml
```diff
@@ -0,0 +1,31 @@
+Type: IPython Notebook Extension
```
jcb91: Jupyter please!

src/jupyter_contrib_nbextensions/nbextensions/pyVariableInspector/main.js
jcb91: and given that it's in the `pyVarInspector` object already, it doesn't really need the `varInspector_` prefix for the name, so `IPython.notebook.metadata.pyVarInspector.position`

src/jupyter_contrib_nbextensions/nbextensions/pyVariableInspector/main.js
```diff
@@ -0,0 +1,397 @@
...
+    var load_ipython_extension = function() {
+        load_css(); //console.log("Loading css")
+        varInspector_button(); //console.log("Adding varInspector_button")
+
+        // Wait for the notebook to be fully loaded
+        if (IPython.notebook !== undefined && IPython.notebook._fully_loaded) {
+            // this tests if the notebook is fully loaded 
+            console.log("[pyVarInspector] Notebook fully loaded -- pyVarInspector initialized ")
+            varInspector_init();
+            //          setTimeout(function(){
```
jcb91: can we strip the commented stuff out?

src/jupyter_contrib_nbextensions/nbextensions/pyVariableInspector/main.js
```diff
@@ -0,0 +1,397 @@
...
+        'window_display': false,
+        'lenName': 16,
+        'lenType': 16,
+        'lenVar': 40,
+        'types_to_exclude': ['module', 'function', 'builtin_function_or_method', 'instance', '_Feature']  
+    }
+
+    //.....................global variables....
+
+    var liveNotebook = !(typeof IPython == "undefined")
```
jcb91: does it make sense to have a variable inspector in a non-live notebook?!

src/jupyter_contrib_nbextensions/nbextensions/pyVariableInspector/README.md
```diff
@@ -0,0 +1,25 @@
...
+#### Demo:
+![](demo.gif)
+
+ 
+## Configuration
+The initial configuration can be given using the IPython-contrib nbextensions facility. It includes:
+
+- Display at startup or not (default: false) 
+- Width of columns (actually the max number of character to display in each column)
+- The state of the toc cell (default: false, ie not present)
```
jcb91: toc cell? I guess this might be a copy-paste hangover? Actually, for all these options, I think it's nice to state the variable name in the config, e.g. 

```markdown
 - `pyVarInspector.types_to_exclude`: The list of types to exclude from display (default:...
```

juhasch:
Sounds great!

Trying this extension for the first time, I get this error:
```
Uncaught (in promise) TypeError: Cannot read property 'types_to_exclude' of undefined
    at :8888/nbextensions/pyVariableInspector/main.js?v=20170318173437:46
main.min.js:12412 
```
After changing something in the configuration parameters it works.

jfbercher:
@jcb91 Thanks for the code review. I have taken all this into account. 
@juhasch Thanks. Indeed, configuration implicitly assumed that a `pyVariableInspector` was present in `notebook.json`. I have fixed this. 

In the new PR, I have 

- done some cleanup,
- reworked window restoration on a new session,
- reworked configuration loading, extension initialization (condition for initializing was incorrectly on `notebook._fully_loaded` while the correct condition is to test for a python kernel to be available),
- added a local copy of jquery tablesorter plugin
- added an icon for `nbextensions_configurator

janschulz:
Just a curious question: how hard would it be for other kernel to adapt this (or the other way around: to adapt this code to other kernels)? E.g. what would the R kernel need to implement to get the same variable view?

Background: This is now the second time (or more? First was the code prettyfier), where a UI function needs support code in the kernel.

jcb91:
@janschulz from what i can see, the kernel currently produces html which gets consumed directly. To simplify provide multi-kernel support, and to separate html from the kernel side, I'd be inclined to switch to a json-based model similar to code-prettify, where the kernel code would need to return an json array of object with keys (say) name, type, size, repr. Then the js nbextension side can handle formatting that correctly...

juhasch:
I agree with @jcb91, a JSON interface would be preferable.
Also, the message box 'pyVariableInspector only works with a Python kernel' is annoying when working with different Kernels. I think simply disabling the extension would be better in case of a non-Python kernel. 

For multi-kernel support, the kernels-specific code could be loaded instead of being hardcoded. This also makes it easier to see what is going on, especially not running code fragments hidden in a js file.

jfbercher:
@juhasch 
- the alert box is removed in favor of a warning in the console, and the toolbar button is removed (the extension is not initialized though)

@jcb91 @janschulz  @juhasch 
- indeed I had in mind that this could be extended to include support for multiple kernels, as we did for `code_prettify`
- support for R kernel should not be difficult
- I have created a [dedicated repo](https://github.com/jfbercher/jupyter_varInspector) and invited you as collaborators. Feel free to update the code! I will shortly try to open an issue there with ideas for the roadmap.

jfbercher:
**varInspector**

The last two commits implement
   - multiple kernels support
    - config per kernel,
    - kernel code read from external files,
    - json communication between javascript/kernel implemented
and finally
   -  support for R kernel 

In addition, I renamed the extension as `varInspector` since it is not restricted to python anymore. 

I think that this kind of think will be probably available in JupyterLab, so it might not be necessary too spend to much time on it.

juhasch:
Great improvement!

src/jupyter_contrib_nbextensions/nbextensions/varInspector/var_list.py
```diff
@@ -0,0 +1,20 @@
...
+   if type(x).__name__ in ['ndarray', 'Series']:  return x.nbytes
+   elif type(x).__name__ == 'DataFrame': return x.memory_usage().sum()
+   else: return getsizeof(x)
+
+def var_dic_list():
+    types_to_exclude = ['module', 'function', 'builtin_function_or_method', 'instance', '_Feature']    
+    values = _nms.who_ls()
+    vardic = [{'varName':v, 'varType': type(eval(v)).__name__, 'varSize': _getsizeof(eval(v)), 'varContent': str(eval(v))[:200]} for v in 
+    values if (v not in ['_html', '_nms', 'NamespaceMagics', '_Jupyter']) & (type(eval(v)).__name__ not in types_to_exclude)]   
+    return json.dumps(vardic)
```
juhasch: Shouldn't there be a `import json` for this ?

jcb91:
Nice! Sorry I haven't had a chance to look at this myself recently, but good update :smile:
Looks like we can probably merge once the json import is added. Could you fix the linting errors while you're there :blush: (see travis [lint#L474](https://travis-ci.org/ipython-contrib/jupyter_contrib_nbextensions/jobs/218058709#L474) on).

jfbercher:
@jcb91 @juhasch 
I think that the extension is ready (but can always be improved of course). Ok to merge?

jfbercher:
@jcb91 I will update `nbtranslate.py` after to fix lot of lint errors.

jcb91:
:+1: I'm away from a computer so can't make a proper check, but I think we're fine to merge and then iterate later. Nice work @jfbercher :)

jcb91:
(ignore the currently-failing Travis conda build for now...)

juhasch:
👍 from me, too.

Issue #924, opened 2017-03-23
pip install problem


EBazarov:
When try to install lib, it gives error on installing `jupyter_nbextensions_configurator`.

```
Collecting jupyter_contrib_nbextensions
  Using cached jupyter_contrib_nbextensions-0.2.5-py2.py3-none-any.whl
Requirement already satisfied: tornado in ./.local/share/virtualenvs/kaidee/lib/python3.5/site-packages (from jupyter_contrib_nbextensions)
Collecting jupyter-highlight-selected-word>=0.0.10 (from jupyter_contrib_nbextensions)
  Using cached jupyter_highlight_selected_word-0.0.10-py2.py3-none-any.whl
Collecting jupyter-contrib-core>=0.3 (from jupyter_contrib_nbextensions)
  Using cached jupyter_contrib_core-0.3.0-py2.py3-none-any.whl
Collecting jupyter-nbextensions-configurator>=0.2.5 (from jupyter_contrib_nbextensions)
  Could not find a version that satisfies the requirement jupyter-nbextensions-configurator>=0.2.5 (from jupyter_contrib_nbextensions) (from versions: 0.1.1, 0.2.0, 0.2.1, 0.2.2, 0.2.3, 0.2.4)
No matching distribution found for jupyter-nbextensions-configurator>=0.2.5 (from jupyter_contrib_nbextensions)
```

I have verified last version of `jupyter_nbextensions_configurator` available is **0.2.4**, please change in `setup.py` line with ` jupyter-nbextensions-configurator>=0.2.5` to ` jupyter-nbextensions-configurator>=0.2.4`

jcb91:
oh bother, that's my fault, I've mixed up the 0.2.4 of the configurator with the 0.2.5 from here :scream:

jcb91:
wait, it wasn't me! It was bumpversion, which I use to update the versions. Agh, that's annoying. It'll mean a new, 0.2.6 release, unfortunately, but there we are, I suppose.

jcb91:
Ok, 0.2.6 should be up on pypi now with the correct requirements. Apologies all!

jcb91:
I think this is fixed now, but please reopen if not...

PR #925, opened 2017-03-23
[dragAndDrop] use markdown syntax instead of html

xNok:
As discus in https://github.com/jupyter/nbconvert/issues/552 

> [nbconvert] don't support pdf export from general HTML markup in markdown cells

Thus I think it is a better Idea to use markdown syntax, so this extension stays coherent with the ecosystem.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 0.2.5
+current_version = 0.2.6
 commit = True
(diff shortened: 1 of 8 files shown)


jcb91:
Hi @xNok thanks for the PR, unfortunately due to a rather silly versioning error of mine earlier today (see #924), the master which you based this on (and hence the PR merge commit) won't build correctly. Could you rebase this onto the current master?

PR #928, opened 2017-03-27
made it work with 5.0.0.dev again

ZelphirKaltstahl:
In version `5.0.0.dev` of the notebook `Jupyter.notebook.config.loaded` is not a function anymore, but a promise. By taking away the `()` the extension works with 5.0.0.dev again =)


jcb91:
Actually, I think it always was a promise, and the error is just a result of me confusing `load` with `loaded` (d'oh!). Thanks for catching this @ZelphirKaltstahl

PR #929, opened 2017-03-27
[dragAndDrop] use markdown syntax instead of html

xNok:
same as https://github.com/ipython-contrib/jupyter_contrib_nbextensions/pull/925, without the merge commit into the history

As discus in jupyter/nbconvert#552

[nbconvert] don't support pdf export from general HTML markup in markdown cells
Thus I think it is a better Idea to use markdown syntax, so this extension stays coherent with the ecosystem.

### DIFF
--- a/src/jupyter_contrib_nbextensions/nbextensions/dragdrop/readme.md
+++ b/src/jupyter_contrib_nbextensions/nbextensions/dragdrop/readme.md
@@ -21,7 +21,7 @@ Internals
 The image will be uploaded to the server into the directory where your notebook resides. This means, the image is not copied into the notebook itself, it will only be linked to. The markdown cell in the notebook will contain this tag:
 
 ```markdown
-![](http://127.0.0.1:8888/notebooks/myimage.png)
+![](myimage.png)
 ```
 
 The name of the image will be kept, if the drag&drop operation originates from a file system.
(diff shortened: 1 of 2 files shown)


src/jupyter_contrib_nbextensions/nbextensions/dragdrop/readme.md
```diff
@@ -20,8 +20,8 @@ Internals
 
 The image will be uploaded to the server into the directory where your notebook resides. This means, the image is not copied into the notebook itself, it will only be linked to. The markdown cell in the notebook will contain this tag:
 
-```html
-<img  src="http://127.0.0.1:8888/notebooks/myimage.png"/>
+```markdown
+![](http://127.0.0.1:8888/notebooks/myimage.png)
```
juhasch: I think it would be better to shown an example using a local file, not a URL.

juhasch:
Thanks, looks good to me, I only have a minor note. 
As this change is needed for PDF support, it might be useful to change the example.

xNok:
You are right the tag generated contains only the file name. (the image is uploaded next to the notebook file)

mpacer:
Looks like the conda recipe is broken — it's using py3.6 even on what is ostensibly a py3.4 build. That may be related to the error.

jcb91:
It is broken, but as far as I can tell so far,  that's not the reason (the
3.4 is provided by Travis, which won't provide 3.6, but either way the
build will ignore Travis python and install the conda version anyway, which
is 3.6). I have a feeling​ it may be related to our use of tox, but can't
reproduce the error locally. I'm going to try not using tox for the conda
build, but at any rate, this issue can be ignored for now, and certainly on
this pr, which doesn't alter any of he installation mechanics.

On 4 Apr 2017 5:50 pm, "M Pacer" <notifications@github.com> wrote:

Looks like the conda recipe is broken — it's using py3.6 even on what is
ostensibly a py3.4 build. That may be related to the error.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/ipython-contrib/jupyter_contrib_nbextensions/pull/929#issuecomment-291562043>,
or mute the thread

PR #946, opened 2017-04-05
Use a scheduling list for running marked cells

juhasch:
Run each cell individually after the previous one has finished. This allows stopping execution of a following cell:
run_list

Note: Only works with Notebook >= 5.0


juhasch:
Allow interrupting scheduled execution. Second click on interrupt button stops execution immediately.
interrupt

gabyx:
Awesome feature! Really helpful

src/jupyter_contrib_nbextensions/nbextensions/runtools/main.js
```diff
@@ -1,15 +1,153 @@
...
+        run_all_cells : 'Alt-x',
+        run_all_cells_ignore_errors : 'Alt-f',
+        stop_execution : 'Ctrl-c',
+        marked_color : '#0f0',
+        scheduled_color : '#00f',
+        run_color : '#f5f5f5'
+    };
+
+    // updates default params with any specified in the provided config data
+    var update_params = function (config_data) {
```
jcb91: this function never actually gets called. Also, given parameters are all `runtools.x`, it would make more sense to just do 

```javascript
$.extend(true, params, config_data.runtools);
```

src/jupyter_contrib_nbextensions/nbextensions/runtools/main.js
```diff
@@ -1,15 +1,153 @@
...
     'base/js/events',
+    'services/config',
+    'base/js/utils',
     'codemirror/lib/codemirror'
-],   function(IPython, $, require, events, codemirror) {
+],   function(Jupyter, $, require, events, configmod, utils, CodeMirror) {
     "use strict";
 
+    var base_url = utils.get_body_data("baseUrl");
+    var config = new configmod.ConfigSection('notebook', {base_url: base_url});
```
jcb91: it's better I think to use `Jupyter.notebook.config`, which is loaded by the page already, and means we don't have to make a second request to the server (or load utils/config modules)

src/jupyter_contrib_nbextensions/nbextensions/runtools/main.js
```diff
@@ -1,15 +1,153 @@
...
+    // updates default params with any specified in the provided config data
+    var update_params = function (config_data) {
+        for (var key in params) {
+            if (config_data.hasOwnProperty(key)) {
+                params[key] = config_data[key];
+            }
+        }
+    };
+
+    config.loaded.then(function() {
```
jcb91: if using `Jupyter.notebook.config`, as suggested above, loaded has already resolved, so this would get executed immediately. Instead, this should be changed into a function (say `initialize`), then in load_ipython_extension you can call `Jupyter.notebook.config.loaded.then(initialize);`

src/jupyter_contrib_nbextensions/nbextensions/runtools/main.js
```diff
@@ -62,36 +62,83 @@ define([
...
+     */
+    var finished_execute_event = function(evt, data)
+    {
+        var cell = data.cell;
+        /* Reset gutter color no non-queued state */
+        if ((cell instanceof IPython.CodeCell)) {
+            if (cell.metadata.run_control != undefined) {
+                if (cell.metadata.run_control.marked == true ) {
+                    var g=cell.code_mirror.getGutterElement();
+                    $(g).css({"background-color": "#0f0"});
```
jcb91: Wait, I think maybe it's not a reset, and the comment needs altering?

src/jupyter_contrib_nbextensions/nbextensions/runtools/main.js
```diff
@@ -145,6 +328,23 @@ define([
...
 
     /**
+     *
+     * @param cell notebook cell instance
+     * @param state {string} state to be display [ '', 'locked', 'executed', 'modified' ]
+     */
+    function set_cell_state(cell, state) {
+        var icon = "";
+        if (state === 'locked')
+            { icon = '<i class="fa fa-lock" /i>' }
```
jcb91: should this also have `aria-hidden="true"`?

src/jupyter_contrib_nbextensions/nbextensions/runtools/main.js
```diff
@@ -521,57 +630,63 @@ define([
...
+        /* require our additional custom codefolding modes before initialising fully */
+        require(['./cellstate'], function () {
+            if (Jupyter.notebook._fully_loaded) {
+                initGutter();
+            }
+            else {
+                events.one('notebook_loaded.Notebook', initGutter);
+            }
+        });
+        config.load()
```
jcb91: this can be replaced with `Jupyter.notebook.config.loaded.then(initialize);` as discussed above

src/jupyter_contrib_nbextensions/nbextensions/runtools/runtools.yaml
```diff
@@ -4,4 +4,61 @@ Description: Runtools provide a number of additional functions for working with
 Link: readme.md
 Icon: icon.png
 Main: main.js
-Compatibility: 4.x
+Compatibility: 5.x
```
jcb91: does this no longer work on 4.x?

src/jupyter_contrib_nbextensions/nbextensions/runtools/runtools.yaml
```diff
@@ -4,4 +4,61 @@ Description: Runtools provide a number of additional functions for working with
...
+  input_type: hotkey
+  default: Alt-x
+- name: runtools.run_alls_cells_ignore_errors
+  description: run all cells - ignore errors
+  input_type: hotkey
+  default: Alt-f
+- name: runtools.marked_color
+  description: |
+    Color for marking a codecell
+  default: '#0f0'
```
jcb91: I have a feeling that the configurator fails to load correctly 3-character colour codes, so it's probably best to use the full 6-char code here

src/jupyter_contrib_nbextensions/nbextensions/runtools/runtools.yaml
```diff
@@ -4,4 +4,61 @@ Description: Runtools provide a number of additional functions for working with
...
+  input_type: color
+- name: runtools.scheduled_color
+  description: |
+    Color when a codecell is scheduled to be rund
+  default: '#00f'
```
jcb91: as above

src/jupyter_contrib_nbextensions/nbextensions/runtools/cellstate.js
```diff
@@ -0,0 +1,20 @@
...
+      cm.state.cellState = null;
+      cm.off("gutterClick", onGutterClick);
+      cm.off("change", onChange);
+      cm.off("viewportChange", onViewportChange);
+      cm.off("swapDoc", onChange);
+    }
+    if (val) {
+      cm.state.cellState = new State(parseOptions(val));
+      updateInViewport(cm);
+      cm.on("gutterClick", onGutterClick);
```
jcb91: are these defined in this scope? How does this file get loaded?

juhasch:
@jcb91: Thanks for the review

The reason I did not continue to work on this (apart from restricted time) is, that I am not sure what the right behavior should be. Or if there is the a 'right' behavior at all.

What happens if you reshuffle the cells during execution? Should they be executed in the original sequence or should the cells that are now further up be executed first?
Both approaches have their merits. A UI that makes this transparent is necessary for this and I have no idea how it should look right now. Suggestions are welcome :smile:

src/jupyter_contrib_nbextensions/nbextensions/runtools/cellstate.js
juhasch: It is loaded at `load_extension` time now.

juhasch:
OK, I have updated the extension, incorporated @jcb91's comments and updated the documentation.

PR #948, opened 2017-04-06
[tree-filter] Make search case insensitve and accept RegExp

wikiped:
Currently filtering entries in the file tree is Case Sensitive. This patch will make it case insensitive and it will also accept Regular Expressions as valid search strings.

### DIFF
--- a/src/jupyter_contrib_nbextensions/nbextensions/tree-filter/index.js
+++ b/src/jupyter_contrib_nbextensions/nbextensions/tree-filter/index.js
@@ -6,14 +6,12 @@ define([
     'require',
     'jqueryui',
     'base/js/namespace',
-    'base/js/events',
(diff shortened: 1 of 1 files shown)


jcb91:
I think there's an issue requesting case sensitivity on the upstream repo https://github.com/jdfreder/tree-filter so it might be worth adding this there too? It'd also be good to either make regex optional, or at least a little more obvious that it's happening (e.g. readme, tooltips)

juhasch:
I think making the regex configurable sounds good.

wikiped:
[The patch](https://github.com/wikiped/jupyter_contrib_nbextensions/pull/2) was added to make case sensitivity / RegExp optional.

juhasch:
Thanks @jcb91

PR #951, opened 2017-04-08
[nbTranslate] improved code formatting in nbTranslate.py

jfbercher:
As the title says. Should  remove a bunch of linting warnings.


jcb91:
:+1: it seems though that there's a second copy of nbTranslate.py in the nbextension directory - I guess this should be reomved?

jfbercher:
@jcb91 Yes of course!

jcb91:
:+1: thanks!

PR #952, opened 2017-04-08
[codefolding] in edit view

### LINKED ISSUE #938: Enable codefolding on regular python files
I guess the name pretty much says it all.
Instead of just the notebooks, allow regular python (or other languages) to support code folding.

Thank you!

jcb91:
fixes #938

### DIFF
--- a/src/jupyter_contrib_nbextensions/nbextensions/codefolding/codefolding_editor.yaml
+++ b/src/jupyter_contrib_nbextensions/nbextensions/codefolding/codefolding_editor.yaml
@@ -1,0 +1,12 @@
+Type: Jupyter Notebook Extension
+Name: Codefolding in Editor
+Description: |
+  Enables the CodeMirror feature to allow codefolding in the Jupyter file
+  editor view.
+  Note that this also uses the codefolding hotkey from the codefolding
+  nbextension for the notebook view.
+Link: readme.md
+Icon: icon.png
+Main: edit.js
+Compatibility: 4.x
+Section: edit
--- a/src/jupyter_contrib_nbextensions/nbextensions/codefolding/edit.js
+++ b/src/jupyter_contrib_nbextensions/nbextensions/codefolding/edit.js
@@ -1,0 +1,4 @@
+define(['./main'], function (codefolding) {
+    "use strict";
+    return codefolding;
+});
--- a/src/jupyter_contrib_nbextensions/nbextensions/codefolding/main.js
+++ b/src/jupyter_contrib_nbextensions/nbextensions/codefolding/main.js
@@ -40,25 +40,25 @@ define([
 
     var on_config_loaded = function () {
         if (Jupyter.notebook !== undefined) {
-        // register actions with ActionHandler instance
-        var prefix = 'auto';
-        var name = 'toggle-codefolding';
-        var action = {
-            icon: 'fa-comment-o',
-            help    : 'Toggle codefolding',
-            help_index : 'ec',
-            id : 'toggle_codefolding',
-            handler : toggleFolding
-        };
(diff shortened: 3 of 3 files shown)


src/jupyter_contrib_nbextensions/nbextensions/codefolding/codefolding_editor.yaml
```diff
@@ -0,0 +1,12 @@
+Type: Jupyter Notebook Extension
+Name: Codefolding in Editor
+Description: |
+  Enables the CodeMirror feature to allow codefolding in the Jupyter file
+  editor view.
+  Note that this also uses the codefolding hotkey from the codefolding
+  nbextension for the notebook view.
```
juhasch: I guess we could add the hotkey as parameter here, too.

src/jupyter_contrib_nbextensions/nbextensions/codefolding/codefolding_editor.yaml
```diff
@@ -0,0 +1,12 @@
...
+Link: readme.md
+Icon: icon.png
+Main: edit.js
+Compatibility: 4.x
```
juhasch: Also 5.x ?

juhasch:
Thanks for updating the code 👍 

Do you mean to include updates to `skip-traceback` in this PR?

src/jupyter_contrib_nbextensions/nbextensions/codefolding/codefolding_editor.yaml
jcb91: Yes,  although it'll be stored in edit config section rather than notebook, so will be independent of the notebook view one. Not a problem though, really.

src/jupyter_contrib_nbextensions/nbextensions/codefolding/codefolding_editor.yaml
jcb91: Could be, i haven't tried it...

jcb91:
> skip-traceback in this PR?

Oh, no I didn't, that was meant to be separate. I was clearly doing that too late at night! Hang fire on merging & I'll rebate this to remove the traceback & add that as another pr later...

src/jupyter_contrib_nbextensions/nbextensions/codefolding/codefolding_editor.yaml
juhasch: Didn't think of that. We should only have one hotkey. I don't think anyone would want different hotkeys in notebook and edit mode.

jcb91: Well, it's possible, since i found that using codemirror to bind hotkeys (as I had to do in edit mode because the keyboard manager isn't in use in edit view) means that many that work in notebook view aren't recognised in the edit view. Perhaps I should try adding a keyboard manager...

jcb91: Alternatively, i could add the configurator feature I made a while back to allow the parameters to go in a different config section to the enabled/disabled status...

juhasch: I think it is fine to leave as is.

juhasch:
For me, the hotkey does not work in editor mode. 
I added an icon in PR #954.

jcb91:
> For me, the hotkey does not work in editor mode.

Yeah, the `alt-f` hotkey didn't work for me either, though setting it to `ctrl-[` does, so I'm assuming for now that it's something to do with how codemirror handles hotkeys. I think we merge this, then fix the hotkey later

PR #953, opened 2017-04-09
Add option to indent comments to current indent level

juhasch:
Add option do enable indenting comments like in https://github.com/jupyter/notebook/issues/954#issuecomment-292800281
comment_uncomment

### DIFF
--- a/src/jupyter_contrib_nbextensions/nbextensions/comment-uncomment/readme.md
+++ b/src/jupyter_contrib_nbextensions/nbextensions/comment-uncomment/readme.md
@@ -6,4 +6,6 @@ Adds a new configurable hotkey binding to toggle comments on/off.
 Options
 -------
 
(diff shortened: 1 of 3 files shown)


src/jupyter_contrib_nbextensions/nbextensions/comment-uncomment/readme.md
```diff
@@ -2,3 +2,8 @@ Comment/Uncomment
 =================
 
 Adds a new configurable hotkey binding to toggle comments on/off.
+
+Options
+-------
+
+indent: add comment at current indent level instead at beginning of line
```
jcb91: I think this should be 

> instead of at the beginning of the line

?

Might also be worth givin the full name of `comment_uncomment_indent`.

juhasch: True.

PR #957, opened 2017-04-11
alter skip-traceback to make traceback collapsed, rather than just omitted entirely

### LINKED ISSUE #935: [feature request] collapsible traceback
I like the look of the skip_traceback extension.  However, it would be useful if the traceback information would not be lost altogether.  I think it would be great if there were a small `+` next to the traceback, with which one could (un)collapse the traceback information.  The default behaviour should be the condensed display of skip_traceback.

jcb91:
fixes #935 (I think!)


juhasch:
Looks nice. When testing this with Chrome on Mac:
 * I can't disable hiding tracebacks with the toolbar button
 * Copying to clipboard does not work, or copies the code cell contents

jcb91:
> * I can't disable hiding tracebacks with the toolbar button

This is presumably the same as mentioned in #970. I expect I've merged this with some error in :disappointed:. I'll have a look

> * Copying to clipboard does not work, or copies the code cell contents

This was definitely working for me in chrome on Ubuntu 16.04, but again, I may well have broken something somewhere, so I'll investigate

jcb91:
> * I can't disable hiding tracebacks with the toolbar button

So what exactly happens here? You have the tracebacks hidden, you click the toolbar button, they don't un-hide? Or newly-created cells still have theirs hidden? It seems to be working ok for me, thought there does seem to be a bug that doesn't correctly *hide* tracebacks the first time the toggle is activated (though it does work if you toggle off-on-off, weirdly).

> * Copying to clipboard does not work, or copies the code cell contents

The copying seems to work fine for me on notebook 4.3.1, Chrome 56.0.2924.87 and Firefox 53.0, on Ubuntu 16.04. What notebook version are you using? Do you have an example notebook I could try? If it comes to it, I guess I can fire up OSX, although it runs at a snails pace for me at the moment :laughing:

juhasch:
OK, I see you removed the toolbar button. And my visual memory remembered the wrong button :grimacing:

This is what happens with the clipoard:
traceback

juhasch:
Not very different with Firefox:
traceback-ff

Issue #940, opened 2017-04-04
Freeze not working


SteliosStefanatos:
Freeze does not work properly for me. In "read-only" mode I am able to edit (and execute) the cell and in "freeze" mode I am able to execute the cell. Actually, I can still edit the cell contents in freeze mode, however, the last "unrestricted" version of the cell will be executed.

I have installed ver. 0.2.6 of `jupyter_contrib_nbextensions` and I am using Linux, Firefox, and Anaconda (python 3.6) with the latest updates (`conda update --all`). Freeze behavior is the same when I disable all other extensions.

P.S.: Thank you to all the contributors of nbextensions for your amazing work!

jcb91:
Hmm, I'm afraid I can;t think of an obvious reason for this problem right now. Do you see any warnings or error message in the 
[javascript console](http://webmasters.stackexchange.com/questions/8525/how-to-open-the-javascript-console-in-different-browsers)? Ping @kukanya as freeze author...

SteliosStefanatos:
Thank you for the reply.
I am getting the following javascript warnings, don't know if they are relevant or not...

```
Using //@ to indicate sourceMappingURL pragmas is deprecated. Use //# instead[Learn More]  main.min.js:1

Use of Mutation Events is deprecated. Use MutationObserver instead.  main.min.js:5:7259

Use of getPreventDefault() is deprecated.  Use defaultPrevented instead.  main.min.js:5:12747

unreachable code after return statement[Learn More]  highlighter.js:98:4
```

BTW, I stated in my original post that I could execute the cell in "freeze" mode, which is not the case; I cannot execute a "freezed" cell, which is the expected behavior. So the problem is only that I can edit the cell (either in "read-only" or "freeze" mode), with the edits applied when I execute in "read-only" mode.

tnarik:
If it were modified to manipulate the 'editable' metadata property of the cell, then it would work correctly.
I'm not familiar with Jupyter Notebook extensions, so not sure how easy it would be to change that behaviour, or if it is already covered by a different extension (wasn't able to find one).

juhasch:
The reason for this behavior is that notebook 5.0 introduces it's own implementation of read-only cells.
This extension needs to be updated. For now, it would be sufficient to replace the line
```    
 cell.code_mirror.setOption('readOnly', cell.metadata.run_control.read_only);
```
with this line
```
cell.metadata.editable = !cell.metadata.run_control.read_only;
```

[...]

PR #958, opened 2017-04-13
Freeze read only feature fixed for Jupyter 5.x (and compatible with Jupyter 4.x)

tnarik:
Proposed fix for #940.

### DIFF
--- a/src/jupyter_contrib_nbextensions/nbextensions/freeze/main.js
+++ b/src/jupyter_contrib_nbextensions/nbextensions/freeze/main.js
@@ -48,8 +48,6 @@ define([
             return
         }
 
-        cell.metadata.run_control = {};
-
         state = state || 'normal';
         var editable;
         var new_run_control_values;
@@ -77,7 +75,7 @@ define([
                 bg = "#f0feff";
                 break;
         }
-        $.extend(cell.metadata.run_control, new_run_control_values);
+        $.data(cell.metadata, 'run_control', new_run_control_values);
         cell.metadata.editable = editable;
         cell.code_mirror.setOption('readOnly', !cell.metadata.editable);
         var prompt = cell.element.find('div.input_area');
--- a/src/jupyter_contrib_nbextensions/nbextensions/freeze/main.js
+++ b/src/jupyter_contrib_nbextensions/nbextensions/freeze/main.js
@@ -51,9 +51,6 @@ define([
         if (cell.metadata.run_control === undefined)
             cell.metadata.run_control = {};
 
-        if (cell.metadata.editable === undefined)
-            cell.metadata.editable = true;
-
         state = state || 'normal';
         var editable;
         var new_run_control_values;
@@ -82,7 +79,7 @@ define([
                 break;
         }
         $.extend(cell.metadata.run_control, new_run_control_values);
-        $.extend(cell.metadata.editable, editable);
+        cell.metadata.editable = editable;
         cell.code_mirror.setOption('readOnly', !cell.metadata.editable);
         var prompt = cell.element.find('div.input_area');
         prompt.css("background-color", bg);
--- a/src/jupyter_contrib_nbextensions/nbextensions/freeze/main.js
+++ b/src/jupyter_contrib_nbextensions/nbextensions/freeze/main.js
@@ -51,34 +51,39 @@ define([
         if (cell.metadata.run_control === undefined)
             cell.metadata.run_control = {};
 
+        if (cell.metadata.editable === undefined)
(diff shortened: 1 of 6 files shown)


jcb91:
Thanks for this, it looks good to me. As noted in https://github.com/ipython-contrib/jupyter_contrib_nbextensions/issues/885#issuecomment-280071268, there's also a `cell.metadata.deletable` attribute, though it seems that non-editable cells are implicitly non-deletable, so I guess we don't need to set that as well. However, it might be good to migrate & remove the old metadata values?

tnarik:
I wasn't sure about what you thought regarding that, but yes, a migration
of the metadata would be good to have for any notebook edited with an old
version of the plugin.

it's also pretty straightforward and I can do that in day or even sooner.

On 14 Apr 2017 8:36 p.m., "Josh Barnes" <notifications@github.com> wrote:

Thanks for this, it looks good to me. As noted in #885 (comment)
<https://github.com/ipython-contrib/jupyter_contrib_nbextensions/issues/885#issuecomment-280071268>,
there's also a cell.metadata.deletable attribute, though it seems that
non-editable cells are implicitly non-deletable, so I guess we don't need
to set that as well. However, it might be good to migrate & remove the old
metadata values?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/ipython-contrib/jupyter_contrib_nbextensions/pull/958#issuecomment-294221073>,
or mute the thread

tnarik:
Updated the notebook to read old format ('run_control>read_only') metadata.
Also fixed an issue by which, once touched by the plugin, the 'editable' metadata was set to 'false' (because in Jupyter 5.x `editable = true` means the flag won't be persisted to file).

src/jupyter_contrib_nbextensions/nbextensions/freeze/main.js
```diff
@@ -116,9 +118,14 @@ define([
                 continue;
             }
             var state = 'normal';
+            // Old metadata format
```
jcb91: Oh, I was thinking of something more like

```javascript
if (cell.metadata.run_control !== undefined && cell.metadata.run_control.read_only !== undefined) {
    if (cell.metadata.run_control.read_only) {
        cell.metadata.editable = false;
    }
    delete cell.metadata.run_control.read_only;
}
```

so that we adopt the new metadata just by loading the notebook - make sense?

tnarik:
I just didn't want to change much the existing code, even if refactoring
could make sense.

Opening the notebook will immediately take care of the "upgrade" because
loading the extension triggers the  initialization of the cell states and
the setting of the cell metadata (via `set_state`).

Maybe I'm missing your point, it's the first time I touch Jupyter extension
code.




On 18 Apr 2017 2:34 p.m., "Josh Barnes" <notifications@github.com> wrote:

*@jcb91* commented on this pull request.
------------------------------

In src/jupyter_contrib_nbextensions/nbextensions/freeze/main.js
<https://github.com/ipython-contrib/jupyter_contrib_nbextensions/pull/958#discussion_r111954163>
:

> @@ -116,9 +118,14 @@ define([
                 continue;
             }
             var state = 'normal';
+            // Old metadata format

Oh, I was thinking of something more like

if (cell.metadata.run_control !== undefined &&
cell.metadata.run_control.read_only !== undefined) {
    if (cell.metadata.run_control.read_only) {
        cell.metadata.editable = false;
    }
    delete cell.metadata.run_control.read_only;
}

so that we adopt the new metadata just by loading the notebook - make sense?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/ipython-contrib/jupyter_contrib_nbextensions/pull/958#pullrequestreview-33230990>,
or mute the thread

jcb91:
> I just didn't want to change much the existing code, even if refactoring could make sense.

Fair enough, that's a reasonable approach :smile: 

> Opening the notebook will immediately take care of the "upgrade" because
loading the extension triggers the  initialization of the cell states and
the setting of the cell metadata (via `set_state`).

Yes, it will work ok, in the sense that the new `metadata.editable` will be set correctly, __but__ it will also leave behind the old `metadata.run_control.read_only` value, which will override the `metadata.editable` each time the nbextension loads. I think it'd be neater to remove the old-style metadata attribute once it's been transferred to the new `editable` attribute. Does that make more sense?

tnarik:
Ah, sorry. Yes, it makes perfect sense. I thought I tested that but I didn't.

tnarik:
Created a new pull request, #968, instead of this one,

PR #966, opened 2017-04-21
[Freeze] make cell color configurable

### LINKED ISSUE #963: Freeze: cell color should be customizable
I'd like to customize color of frozen cell and read-only cell with `custom.css`.

Currently Freeze uses `style` attribute to change cell color, and color values are hardcoded in `main.js`. 
https://github.com/ipython-contrib/jupyter_contrib_nbextensions/blob/master/src/jupyter_contrib_nbextensions/nbextensions/freeze/main.js#L77

I'd like to change to use `class` attribute and CSS instead. Can I send PR?

shingo78:
Fixes #963

### DIFF
--- a/src/jupyter_contrib_nbextensions/nbextensions/freeze/config.yaml
+++ b/src/jupyter_contrib_nbextensions/nbextensions/freeze/config.yaml
@@ -5,3 +5,15 @@ Link: readme.md
 Icon: icon.png
 Main: main.js
 Compatibility: 4.x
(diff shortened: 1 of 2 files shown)


jcb91:
looks good to me, although the toolbar & patching can be done before the config loads, since they don't use them at the moment?

shingo78:
Is this good?

jcb91:
:+1: looks good to me!

PR #967, opened 2017-04-21
[ExecuteTime] Fix highlight.color option not working

shingo78:
Fix options key typo


jcb91:
haha, good spot, thanks! :+1:

Issue #942, opened 2017-04-04
All extensions marked as "possibly incompatible" for notebook 5.0.0


andyneff:
I'm guessing the compatibility in the yaml files need to be updated to include 5.x (for those that work in 5.x ;) )

jcb91:
Yes, good point. I think quite a few have been used successfully on 5.0 (see https://github.com/jupyter/notebook/issues/2151 for details), but we do still need to update all the yaml files...

jfbercher:
> but we do still need to update all the yaml files...

Oops :-(

jcb91:
So, here's my list of what we need to update. If anyon'es tried any of these in
5.0 and found them to work (or not!), please leave a comment below.
I'll start a PR to update the yaml files which you can also PR updates to
directly.

The following need compatibility updating:
(a filled checkbox means compatible, but not yet declared as such)

 [ ] addbefore/main
 [ ] autosavetime/main
 [ ] autoscroll/main
 [ ] chrome-clipboard/main
 [ ] code_font_size/code_font_size
 [ ] code_prettify/2to3
 [ ] code_prettify/autopep8
 [ ] code_prettify/code_prettify
 [ ] codefolding/edit
 [ ] codefolding/main
 [ ] codemirror_mode_extensions/main
 [ ] contrib_nbextensions_help_item/main
 [ ] datestamper/main
 [ ] dragdrop/main
 [ ] equation-numbering/main
 [ ] execute_time/ExecuteTime
 [x] exercise/main
 [x] exercise2/main
 [ ] gist_it/main
 [ ] help_panel/help_panel
 [ ] hide_header/main
 [ ] hide_input/main
 [ ] hide_input_all/main
 [x] highlighter/highlighter
 [x] hinterland/hinterland
 [ ] init_cell/main
 [ ] keyboard_shortcut_editor/main
 [ ] limit_output/main
 [ ] move_selected_cells/main
 [ ] navigation-hotkeys/main
 [ ] nbTranslate/main
 [ ] notify/notify
 [ ] printview/main
 [x] python-markdown/main
 [ ] qtconsole/qtconsole
 [ ] rubberband/main
 [ ] ruler/main
 [ ] runtools/main
 [ ] scratchpad/main
 [ ] scroll_down/main
 [ ] search-replace/main
 [x] select_keymap/main
 [ ] skill/main
 [ ] skip-traceback/main
 [ ] snippets/main
 [ ] snippets_menu/main
[...]

PR #972, opened 2017-04-24
update yaml compatibility fields for already-5.x-compatible nbextensions

jcb91:
see #942


juhasch:
Great, I totally forgot about this...

Issue #965, opened 2017-04-21
Freeze: freeze cells automatically after execution


yacchin1205:
Hi,
We are developing *run-through* extension to apply the notebook to cloud infrastructure operation, and implements function to freeze cells when they are executed.
https://github.com/NII-cloud-operation/Jupyter-LC_run_through

Currently we have implemented these functions in *run-through* (like https://github.com/NII-cloud-operation/Jupyter-LC_run_through/blob/master/lc_run_through/nbextension/main.js#L315), but we assume that it should be one of these:
- Make freeze extension export `set_state(cell, state)` function and *run-through* uses it
- Implement *freeze executed cells* function in freeze extension

Though we'd like to make PR... Which is better?

jcb91:
Given the change to how set_state will work in #958 (and any potential future changes), I'd be inclined to just export more functions from Freeze. However, in principal, I think we could _also_ have an optional _freeze cells on execution_ setting?

yacchin1205:
@jcb91 Thank you for your comment(and fixing title)!
I think... it's not strange that Freeze has *freeze cells after executing* setting, but also think we should keep it simple... either will do.

Now... I will send a PR to export set_state function from Freeze when #958 is closed.

jcb91:
You're welcome :)

On second thoughts, perhaps it makes sense to keep the freeze-after-executing code in your nbextension, as I can't think of an immediate use for it in regular usage. But, as you say, either will do :stuck_out_tongue_winking_eye: 

#958 is now closed, in favour of the rebased version #968, which is now merged. PR away :)

yacchin1205:
Thanks @jcb91, I will send the PR 😄

yacchin1205:
@jcb91 Thank you for accepting the PR! I'd like to keep the freeze-after-executing code in my extension, and close this issue. Thanks.

jcb91:
No problem, thanks for submitting it 😊

PR #974, opened 2017-04-24
[Freeze] Export set/get_state functions

yacchin1205:
Export *_state functions from Freeze, based on the discussion #965 .


jcb91:
Looks good to me!

PR #976, opened 2017-04-25
[Freeze] Fix property checking

### LINKED ISSUE #975: Freeze: frozen cells are always unlocked when notebook is loaded
`migrate_state` function always unlocks an editable attribute, because the code below is executed even if there are no `cell.metadata.run_control.read_only`.

```
cell.metadata.editable = !cell.metadata.run_control.read_only;
```
https://github.com/ipython-contrib/jupyter_contrib_nbextensions/blob/master/src/jupyter_contrib_nbextensions/nbextensions/freeze/main.js#L59

As a result, the editable metadata is set to true, and the get_state function returns `normal` for the cell after reloaded.

I assume that it is executed only when the metadata has `run_control.read_only` (with old metadata format).

yacchin1205:
Fix #975


jcb91:
Thanks!

PR #977, opened 2017-04-29
Fix codefolding gutter

juhasch:
WIP: The gutter won't be added otherwise using notebbok 5.0.0 / Mac / Chrome.

`[ gutters , "CodeMirror-foldgutter"]` results in: `[Array(0), "CodeMirror-foldgutter"]`
 
`gutters.push('CodeMirror-foldgutter');` results in `["CodeMirror-foldgutter"]`

### DIFF
--- a/src/jupyter_contrib_nbextensions/nbextensions/codefolding/main.js
+++ b/src/jupyter_contrib_nbextensions/nbextensions/codefolding/main.js
@@ -126,10 +126,11 @@ define([
      * @param cell {codecell.CodeCell} code cell to activate folding gutter
      */
     function activate_cm_folding (cm) {
-        var gutters = cm.getOption('gutters');
-        if ($.inArray("CodeMirror-foldgutter", gutters) < 0) {
-            cm.setOption('gutters', [ gutters , "CodeMirror-foldgutter"]);
-        }
+        var gutters = cm.getOption('gutters').slice();
+        if ( $.inArray("CodeMirror-foldgutter", gutters) < 0) {
+                gutters.push('CodeMirror-foldgutter')
+                cm.setOption('gutters', gutters);
+            }
 
         /* set indent or brace folding */
         var opts = true;
--- a/src/jupyter_contrib_nbextensions/nbextensions/codefolding/main.js
+++ b/src/jupyter_contrib_nbextensions/nbextensions/codefolding/main.js
@@ -128,8 +128,7 @@ define([
     function activate_cm_folding (cm) {
         var gutters = cm.getOption('gutters');
         if ($.inArray("CodeMirror-foldgutter", gutters) < 0) {
-            gutters.push('CodeMirror-foldgutter');
-            cm.setOption('gutters', gutters);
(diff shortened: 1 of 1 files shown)


jcb91:
This is my doing, I think, I noticed we were effectively setting a nested array, contrary to the spec, so altered it. However, I must be missing something here, why do we want a 0-length array as the first element? The [CodeMirror manual says](https://codemirror.net/doc/manual.html#option_gutters) it should be an array of strings, and [setting the option](https://github.com/codemirror/CodeMirror/blob/4f08a5c4f459f551d84703b61feaa0f479534d25/src/edit/options.js#L97-L100) calls `setGuttersForLineNumbers`, which does something comparable (I think?!) to my `gutters.push` operation to ensure the linenumbers gutter is/isn't present as appropriate in [src/display/gutters.js#L27-L33](https://github.com/codemirror/CodeMirror/blob/6992c534ad33ce60e7500fe04a99a481ef8f548e/src/display/gutters.js#L27-L33).
Also, do you mean that this works in 4.x, but breaks in 5.x? Is that maybe a CodeMirror version thing? :confused: :grey_question:

juhasch:
It is a strange thing. Actually, I think what you did was correct. 
I came across this when investigation the runtools gutter stuff.

I guess it could be a bug in CodeMirror. When you toggle line numbers from the menu (view->toggle line numbers), the gutter shows up correctly...

jcb91:
> When you toggle line numbers from the menu (view->toggle line numbers), both gutters shows up correctly...

Curiouser and Curiouser! Well, that sort of makes it sound like the `setGuttersForLineNumbers` isn't triggered by setting gutters, but it's right there, at least by CodeMirror 5.8.0 (which is what notebook has [required](https://github.com/jupyter/notebook/blob/4.1.0/bower.json#L8) since at least 4.1.0). However, I also note that 5.0.0 is now using [a patched version of CodeMirror](https://github.com/jupyter/notebook/pull/2068), which I think has also been backported to 4.3.2 or so, and which makes it rather more difficult to track anything down :worried:. Oh well...

juhasch:
Yes, it is not easy to follow what is happening.
Actually I want to disentangle runtools from using the codefolding gutter. In theory it should be easy, in practice no so much. Also, the arrangement of the gutters changes each time, as it is never clear what is being loaded first. The line number gutter however works OK.

I will try to investigate more into how the line numbering gutter implementation gets things right.

juhasch:
This makes adding gutters work. Without this, references instead of individual arrays to specify gutters are used. This seems to trigger strange behavior in CodeMirror.

jcb91:
It seems that the `setOption` call [does a quick equality check](https://github.com/codemirror/CodeMirror/blob/ead38e82dd9f6e00f4c3cf059579a87b39718a5f/src/edit/methods.js#L44), and doesn't call any handlers if `newvalue == oldvalue`, which I guess makes sense. Of course, if we just push into the same array, then the two variables end up referencing the same object, so they compare as equal. :+1:

PR #978, opened 2017-04-29
Remove stale extension

juhasch:
cellstate extension has not been updated for a long time. Won't work with recent notebook versions


jcb91:
looks like a good idea to me :+1:

PR #986, opened 2017-05-11
[nb5] update more nbextensions to list themselves as 5.x-compatible

jcb91:


jcb91:
just a few more I've been using with 5.x lately

Issue #993, opened 2017-05-18
[toc2] changes width of notebook cells


ostrokach:
Enabling "Table of Contents (2)", even with all options unchecked, still changes the witdth of my notebook cells.

With "Table of Contents (2)" disabled:

![selection_014](https://cloud.githubusercontent.com/assets/5614375/26227088/38bec304-3bfe-11e7-94d1-ed73f2c2cbfc.png)

With "Table of Contents (2)" enabled (all options unchecked):

![selection_015](https://cloud.githubusercontent.com/assets/5614375/26227089/3cf103e2-3bfe-11e7-89a9-656c94f5eb51.png)

I'm trying to get a floating table of contents without changing the default witdth of my notebook cells.

Thank you!

---

```
jupyter_contrib_nbextensions 0.2.8-py36_0 conda-forge
```

ostrokach:
Adding the following line to `~/.jupyter/custom/custom.css` and restarting seems to produce the desired result:

```
#notebook-container {
  width:1140px !important;
}
```

jcb91:
> Adding the following line to ~/.jupyter/custom/custom.css and restarting seems to produce the desired result:

Unfortunately this will only be true for your screen width and other in the same range. For substantially different widths, this will cause a similar problem to what you're encountering. Toc2 *should* simply not alter the width *at all* with the option unchecked, but it appears that that is not happening, so this is a bug.

jfbercher:
ok. I can reproduce and will have a look this we.

jfbercher:
Closed by #994

ostrokach:
I tried installing the latest master tarball:

```
pip install https://github.com/ipython-contrib/jupyter_contrib_nbextensions/tarball/master
jupyter contrib nbextension install --sys-prefix
jupyter contrib nbextension install --user
```
restarted the jupyter notebook and refreshed all the windows, but I am still getting the same problem. Am I doing something wrong?

jfbercher:
Can you try to open 
http://localhost:8888/nbextensions/toc2/toc2.js
(assuming that your server runs on port 8888)
and check if function setNotebookWidth(cfg, st) contains 
`$("#notebook-container").css({'width':''}) `?
This is just to check that you execute the last version.

ostrokach:
Sorry, false alarm. I should have uninstalled the conda version before installing the master branch tarball using pip:

```
conda uninstall jupyter_contrib_nbextensions
[...]

PR #994, opened 2017-05-20
[toc2] update for cells width ans sidebar height

jfbercher:
This address issue #993. In case nothing is selected (no widenNotebook and no Sidebar), then don't modify or restore cells width to the original width. 
<strike>In passing I see that I had a pending modification that should improve the transition effect when toggling the sidebar visibility. </strike>
Also recompute sidebar height on window resize event.

### DIFF
--- a/src/jupyter_contrib_nbextensions/nbextensions/toc2/toc2.js
+++ b/src/jupyter_contrib_nbextensions/nbextensions/toc2/toc2.js
@@ -190,6 +190,15 @@ var make_link = function(h, num_lbl) {
     }
 }
 
+  function setSideBarHeight(cfg, st) {
+      if (cfg.sideBar) {
+        var headerVisibleHeight = $('#header').is(':visible') ? $('#header').height() : 0
+          $('#toc-wrapper').css('top', liveNotebook ? headerVisibleHeight : 0)
+          $('#toc-wrapper').css('height', $('#site').height());
+          $('#toc').css('height', $('#toc-wrapper').height() - $('#toc-header').height())
+      }
+  }  
+
   var create_toc_div = function (cfg,st) {
     var toc_wrapper = $('<div id="toc-wrapper"/>')
     .append(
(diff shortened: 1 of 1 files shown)


src/jupyter_contrib_nbextensions/nbextensions/toc2/toc2.js
```diff
@@ -176,15 +177,15 @@ var make_link = function(h, num_lbl) {
                 $('#notebook-container').css('margin-left', 30);
                 $('#notebook-container').css('width', $('#notebook').width() - 30);
             } else { // original width
-              $("#notebook-container").css({'width':"82%", 'margin-left':'auto'})             
+              $("#notebook-container").css({'width':st.nbcontainer_width, 'margin-left':'auto'})
```
jcb91: I think this should probably set an empty string, to not set anything, otherwise it'll break when the window is resized.

jfbercher: Yes, you are right. thanks

src/jupyter_contrib_nbextensions/nbextensions/toc2/toc2.js
```diff
@@ -165,8 +165,9 @@ var make_link = function(h, num_lbl) {
       callback && callback();
   }
 
-  function setNotebookWidth(cfg, st) {
-    //cfg.widenNotebook  = false; 
+  function setNotebookWidth(cfg, st, sideBarIsToggling) {
+    //cfg.widenNotebook  = true; 
+    var sideBarIsToggling = typeof sideBarIsToggling !== 'undefined' ? sideBarIsToggling : false;
```
jcb91: this `sideBarIsToggling` variable doesn't actually seem to get used anywhere - what is it intended for?

src/jupyter_contrib_nbextensions/nbextensions/toc2/toc2.js
```diff
@@ -176,19 +177,28 @@ var make_link = function(h, num_lbl) {
                 $('#notebook-container').css('margin-left', 30);
                 $('#notebook-container').css('width', $('#notebook').width() - 30);
             } else { // original width
-              $("#notebook-container").css({'width':"82%", 'margin-left':'auto'})             
+              $("#notebook-container").css({'width':'', 'margin-left':'auto'})
```
jcb91: the margin-left should probably also be reset here (admittedly, it does end up as `auto` with the default css, but it's neater to just remove the override, in case anyone is trying to customize things)

src/jupyter_contrib_nbextensions/nbextensions/toc2/toc2.js
jfbercher: Nothing. Indeed it is a dead code from a previous attempt. I remove it.

src/jupyter_contrib_nbextensions/nbextensions/toc2/toc2.js
jfbercher: ok

jfbercher:
Thanks to @jcb91 for reviewing and directions for improvements.

PR #996, opened 2017-05-20
[NbTranslate] update

jfbercher:
When calling google translate, some return values have changed.  
The extension is modified accordingly. 
[Actually in the previous version, the google engine returned fail while the translation was done, so implementation was not so obvious]


jfbercher:
Arg. I would need to rebase before.
Ok.  I will push this one later

PR #1006, opened 2017-05-31
Fix cell removing after copying of freezed cell

aminought:
If you freeze cell, then copy it, paste, unfreeze and try to delete, you will be unable to do that. I don't know why, but after copying jupyter adds 'deletable' property and in this case it is 'false'. I fixed it with handling of 'deletable' property.


jcb91:
Thanks! I have a feeling that setting editable false with notebook also implicitly sets deletable false, and freeze isn't handling the un-freezing correctly, as you've found. For future-proofing, I guess we ought to use the notebook-provided functions when they're available...

Issue #1015, opened 2017-06-29
python2 kernel and variable inspector


xcix:
When using python2 kernel and variable inspector is enabled, print operator in jupyter does not work with SyntaxError exception. Output of %whos magic contains print_function from __future__, possibly because of import in the beggining of this file https://github.com/ipython-contrib/jupyter_contrib_nbextensions/blob/master/src/jupyter_contrib_nbextensions/nbextensions/varInspector/var_list.py

jfbercher:
Should be fixed by #1016

PR #1016, opened 2017-06-30
[varInspector] Fix #1014 and #1015 - reload on  %reset magic and fix print on python2

### LINKED ISSUE #1014: varInspector keeps variables persistent after using %reset magic
If you reset the kernel with the `%reset` magic command, all of the previous variables are still present with their old values in the variable explorer. This doesn't appear to happen when the variables are individually deleted with `del`

jfbercher:
- fix #1014 - Actually the `%reset` magic cleared some hidden variables and functions that are used behind the scene by varInspector. Testing the code_exec return message is used to detect this and then the extension is reinitialized (similarly to what happens when restarting the kernel)
- fix for #1015 - removed the import from __future__ solves this
- added some types hidden from listing for python kernels (`type`and `ufunc`)


jfbercher:
[varInspector]
- Reworked detection for the need to reload the extension. Actually in some circumstances, previous treatment led to an infinite loop of reloads
- Added support for pandas dataframes (json representation was not correctly loaded)

PR #1017, opened 2017-07-04
New optional feature to skip the first header of a notebook in the TOC

louisabraham:


louisabraham:
Be careful, I think there are some non intended changes in src/jupyter_contrib_nbextensions/nbextensions/toc2/toc2.js.

I edited the version I installed from pypi, maybe you want to revert those changes.

Only

    +      // skip title if necessary
    +      if (cfg.skipTitle &&!i){ return; }

are necessary in toc2.js.

jfbercher:
This is certainly a good idea. Thanks. 

- I think that it would be better if you can rebase this pull request on the last version (the one on the current repo). The version on Pypi is not up to date. Update from the version on github (just download the toc2.js if you want), and make a new commit. 
- It would be interesting to update the .yaml too, so that users can choose the default for skipTitle. 
- perhaps add a button like the "n" and "t" to allow defining the skipTitle per notebook? 
- update the history (end of README, last modifications are still missing)

jcb91:
> The version on Pypi is not up to date

Feel free to make a new release any time / request that I make one. There are only minimal downsides to having many closely-spaced releases, so I don't think there's any particular requirement that should be satisfied before making one other than that things don't seem broken 😋

jcb91:
I'm going to close this as it should be handled more completely by either #1027 or #1029

jcb91:
thanks for the input @louisabraham

PR #1019, opened 2017-07-08
[code_prettify] added support for IPython magics, cf #1018

### LINKED ISSUE #1018: Code_prettify does not recognize IPython magic commands
E.g., using `%%time` raises the following error, which is probably straight from PyYapf:

> [code_prettify]
> Error: SyntaxError
> invalid syntax (<unknown>, line 1)

Not sure if you feel this is a code_prettify issue or a yapf issue.

jfbercher:
To support IPython magics in python code cells, a function `yapf_reformat` is defined and automatically loaded at initialization as a library. This function takes care of commenting magics before reformatting and uncommenting them after. This should fix #1018.

### DIFF
--- a/src/jupyter_contrib_nbextensions/nbextensions/code_prettify/2to3.yaml
+++ b/src/jupyter_contrib_nbextensions/nbextensions/code_prettify/2to3.yaml
@@ -3,7 +3,7 @@ Name: 2to3 Converter
 Description: Converts python2 code in a notebook's code cell to python3 code
 Link: README_2to3.md
 Main: 2to3.js
-Compatibility: Jupyter (4.x)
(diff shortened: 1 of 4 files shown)


catch22:
Thanks, this PR works well for me.

EWouters:
I noticed that `2to3` has no problem with cell magics without `yapf_reformat` for some reason unknown to me, but in `autopep8` `%%something` will get returned as `% % something`.

Would it be a good idea to add this to the KernelExecOnCells library somehow? I guess we don't want kernel specific code in there, but handling cell magics is a problem any python extension will face when using the library.

I guess the best thing would be to write some js code that handles cell magics so it works for all kernels.

PR #1029, opened 2017-07-22
[toc2] add option toskip h1 headings for use as notebook title

### LINKED ISSUE #1027: [toc2] Distinguish between (unnumbered) title and (numbered) section headings
Jupyter notebooks support six levels of headings in MarkDown through `#`...`######`. The `toc2` extension supports numbering of these headings.

However, a document usually has only one **title** and this title is typically not numbered.

I typically use the level-1 heading (`# Title`) for the title, and would not want to waste a level of numbering on this.

It would be nice if `toc2` would offer an option to interpret the level-1 headings as an unnumbered top-level title, and then start with just one number on level 2. That is, `## Section` gives something like **1. Section**, and not _1.1. Section_.

Or is there another/better way to do notebook titles in MarkDown?

Of course, you could do them using HTML tags to use a larger bold face font. But that would be an  awkward solution.

jcb91:
fixes #1027, similar implementation to #1017

### DIFF
--- a/src/jupyter_contrib_nbextensions/nbextensions/toc2/README.md
+++ b/src/jupyter_contrib_nbextensions/nbextensions/toc2/README.md
@@ -29,7 +29,7 @@ The initial configuration can be given using the IPython-contrib nbextensions fa
 - Widening the display area to fit the browser window (may be useful with sidebar option; default: true)    
 - The numbering of headers (true by default)
 - Moving header title and menus on the left (default: true)
-- Skipping the first header (useful if you want to display a title): set `skipTitle` to true in the notebook's metadata (default: false)
+- Skipping h1 headers, useful if you want to use h1 as unnumbered notebook title (default: false)
 - Customization of highlighting the title of currently selected/running sections.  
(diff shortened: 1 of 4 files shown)


jcb91:
(edit: rebased to include @louisabraham's efforts, added to readme history)

louisabraham:
Thank you! Sorry I didn't reply on the other thread, I didn't have much time lately…

jfbercher:
This will lead to a small issue when exporting to LaTeX. For instance when exporting using the article template (default), the level-1 headings are converted to sections; so the exported document will have a single section but many sub[sub]sections.

jcb91:
> This will lead to a small issue when exporting to LaTeX. For instance when exporting using the article template (default), the level-1 headings are converted to sections; so the exported document will have a single section but many sub[sub]sections.

Hmm, yeah I guess so, but that's already true for any notebook where people are using a level-1 heading as a title. I'm not sure what can be done about that, as it essentially comes from conflicting ideas about what headings are used for (a top-level heading being a title, as opposed to a section). Any ideas welcome!

Issue #1023, opened 2017-07-19
[toc2] with github html preview


bianbian2010:
Hi~

It seems the section links in the floating/side table do not work in html preview mode [example here](https://htmlpreview.github.io/?https://github.com/bianbian2010/temp/blob/master/toc2-example.html) (links in the static table do work so I think it might be how the section IDs are appended to the link?)

I am very new to Jupyter and html/css so sorry for my ignorance. Please let me know if there is a quick fix I can apply on my own to resolve this issue.

Thanks very much!

jcb91:
Hmm, it seems like all the sidebar links (which are generated by javascript, and only specify a hash, like `#H11-1`, not a full uri) are computed by my browser to point to `raw.github.com`, (e.g. `#H11-1` takes you to https://raw.githubusercontent.com/bianbian2010/temp/master/toc2-example.html#H11-1) whereas those on the main page have been specifically filled in with a full uri (e.g. https://htmlpreview.github.io/?https://github.com/bianbian2010/temp/blob/master/toc2-example.html#H11).

Essentially, I think this could be considered a fault of htmlpreview.github.io, rather than toc2. 
My guess is that htmlpreview.github.io is doing some magic to process hrefs which are present on the page at load (see [htmlpreview.js#L25-L33](https://github.com/htmlpreview/htmlpreview.github.com/blob/296f172523a041f2452c60f31546f445169bb3da/htmlpreview.js#L25-L33)), but any links subsequently added by javascript (like those in the sidebar) don't get the same treatment. Perhaps it would make sense to file an issue at [htmlpreview/htmlpreview.github.com/issues](https://github.com/htmlpreview/htmlpreview.github.com/issues)?

bianbian2010:
Thanks very much for the detailed reply!

Is it possible to modify the javascript to generate a full uri? (I can try to make this change on my own if it's not too complicated) I've seen a working sidebar [here](https://htmlpreview.github.io/?https://github.com/ahambi/140824-TOC/blob/master/A%20floating%20table%20of%20contents.htm#What-you-need-to-do) and just want to try in case the href handling from htmlpreview was intended.

jfbercher:
yes, I agree with @jcb91's analysis. the bug is more on the htmppreview side. 
[...]

PR #1030, opened 2017-07-22
toc2 -- Full url in links | #skip tag to skip headers from being inserted

jfbercher:
- #1023 - add full URL in links

- #1027 - add a `#skip` tag for skipping header insertion in toc. Just put 
```
## This is a title #skip
```
and this will be rendered as 
<h2>
This is a title
</h2>
without numbering and will not be inserted in toc.

### DIFF
--- a/src/jupyter_contrib_nbextensions/nbextensions/toc2/README.md
+++ b/src/jupyter_contrib_nbextensions/nbextensions/toc2/README.md
@@ -10,7 +10,12 @@ The toc2 extension enables to collect all running headers and display them in a 
 #### Second demo:
 ![](demo2.gif)
 
-The table of contents is automatically updated when modifications occur in the notebook. The toc window can be moved and resized. It can be docked as a sidebar or dragged from the sidebar into a floating window. The table of contents can be collapsed or the window can be completely hidden. The navigation menu can be enabled/disabled via the nbextensions configuration utility. It can also be resized. The position, dimensions, and states (that is 'collapsed' and 'hidden' states) are remembered (actually stored in the notebook's metadata) and restored on the next session. The toc window also provides two links in its header for further functionalities:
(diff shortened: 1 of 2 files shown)


jcb91:
ah, I hadn't understood your `#skip` suggestion, then! I think I prefer the other implementation though, so as not to add weird tags for people opening the notebook without toc2 enabled...

jfbercher:
> think I prefer the other implementation though, so as not to add weird tags for people opening the notebook without toc2 enabled...

Yes, you are right with that. Let's go with the other solution.

--

jfbercher:
I pushed a variant where any html tag with class "tocSkip" will prevent the corresponding header to be numbered and inserted in toc. 
- This tag will not be displayed if the toc2 extension is not present,
- this can be employed at any heading level, as many times as necessary ,
- this will not affect conversion to eg LateX (html tags are removed).

An example is as follows
```
## A title <a class='tocSkip'>
```
which will be displayed as 
## A title

jcb91:
> I pushed a variant where any html tag with class "tocSkip" will prevent the corresponding header to be numbered and inserted in toc

Looks like a good plan to me. This can also work in parallel with the automatic solution I've implemented in #1029, I think?

jcb91:
Although, [line 537](https://github.com/jfbercher/jupyter_contrib_nbextensions/blob/27b3b4229652f43e8f5c80e5449dd3f52577eda9/src/jupyter_contrib_nbextensions/nbextensions/toc2/toc2.js#L537) should be amended to not include the skipped headers in calculating the numbering levels

jcb91:
ok, I'm going to merge this as is, then address line 537 in #1027, since both approaches should work in parallel

parthi2929:
Hi

Adding ```<a class='tocSkip'>``` kind of looks hacky and not many could be aware. Instead, some short hand like  ```#*``` could be used?

PR #1031, opened 2017-07-23
[toc2] make toc entries collapsible

### LINKED ISSUE #1024: toc2  feature request: Allow each section levels to be collapsed in the floating toc panel
Hi  there, 

   For a large notebook whose toc may be also very long, it is desirable that each nested sections in the toc can be collapsed individually (Currently they are all fully expanded).  Just see similar features of the navigation panel in MS word or Adobe pdf.   Thanks!

jcb91:
Fixes #1024 

Note, in order to make this work neatly, I have also:

* changed how the toc cell is rendered (it now uses the same DOM structure and classes as the sidebar, rather than its previous set of un-nested divs with classes for padding levels)
* altered how toc link ids are generated and handled:
   - for the semi-unique toc id, don't replace any special characters. By handling the search correctly, we can avoid jquery issues with special characters like `.`, `[`, etc. In addition, this prevents potential confusion between e.g. sections 1.13 and 11.3, which would previously have both have been translated to the id as `113`.source
   - keep the original id as the href for all links (previously, the sidebar links used modified id, which cell links used original ids in order not to break exporting), and store the semi-unique modified id as html attribute `data-toc-modified-id`. Use the toc.js-specified click handler to ensure that we actually select the modified id

I'm _reasonably_ confident this all works ok from my limited testing, but I'd definitely welcome others' testing, and in particular, anything @jfbercher might have to say :)


jfbercher:
@jcb91 
I just say WOW!!

I spent a part of the afternoon yesterday investigating how I could do that, but this is now done and very well done! 
Just a couple of minor remarks before testing (I am eventually in vacations and walking through France). 

- I removed the dots from ids because it was causing (or just was reporting -- don't remember well)  causing problems with jquery https://stackoverflow.com/questions/9930577/jquery-dot-in-id-selector May be this is no more the case with recent jquery
-  for `.lev` classes, I know there are better ways to do the things, but I remember that some people rely on it for displaying their (old) generated html documents. Perhaps shall we keep the definitions in the css (which is read from the `jupyter_contrib` repo); or the tpl 

- For collapsing working together with colapsible_headings, perhaps that both extensions can trig an event, like 'toc_collapsing' or 'header_collapsing' and pass the id of the header being clicked? The behaviour can be made configurable via a boolean 'If toc extension is prensent, also collapse headings in toc' and 'if collapsible_heading extension is present, also collapse sections in notebook'

Again. Wow!

jcb91:
> (I am eventually in vacations and walking through France)

Sounds nice, enjoy! :smile:

> I removed the dots from ids because it was causing (or just was reporting -- don't remember well) [causing problems with jquery](https://stackoverflow.com/questions/9930577/jquery-dot-in-id-selector) May be this is no more the case with recent jquery

Yes, I saw references to that in comments, and as far as I can tell, it's still an issue when using jquery's string selectors: Jquery will interpret a `.` to signify a class, a `[` to specify an attribute, etc. as in css rules. However, the browser's `document.getElementById` method has no such limitation, so we can leave special characters in the id as long as we use that. In addition, I'm now searching for links with a specified `data-toc-modified-id` attribute anyway, which I do with jquery's `filter` rather than a string selector for the id.

> for `.lev` classes, I know there are better ways to do the things, but I remember that some people rely on it for displaying their (old) generated html documents. Perhaps shall we keep the definitions in the css (which is read from the `jupyter_contrib` repo); or the tpl

Ah, right, I hadn't thought of that! Yes, in that case, I think we should leave them in the css for backward compatibility. They don't need to be in the template though, as they won't be needed for newly-exported versions.

> For collapsing working together with colapsible_headings, perhaps that both extensions can trig an event, like 'toc_collapsing' or 'header_collapsing' and pass the id of the header being clicked? The behaviour can be made configurable via a boolean 'If toc extension is present, also collapse headings in toc' and 'if collapsible_heading extension is present, also collapse sections in notebook'

Yes, that sounds like a good plan, to have a setting like "sync toc items' collapsed state with collapsible headings nbextension" in toc2 config. We don't really need a setting for each extension, as then we run into priority problems :wink:  I think the simplest way to get it working would be to have collapsible_headings fire an event when collapsing/uncollapsing a heading, something like `collapse.HeadingCell` and `uncollapse.HeadingCell`, passing the relevant cell as data. Then toc2 can (un)collapse the toc to match. For the opposite direction, we just add an optional bit in the toc callback to find the heading cell via the modified id as we do for a go-to click, and then call the toggle_heading method (which ch will export). I'll have a stab now.

jcb91:
> We don't really need a setting for each extension, as then we run into priority problems

Having thought about this more, I think I see what you were getting at, in that you might want the sync in one direction but not the other. I've added separate options for each direction.

Unfortunately, non-live notebooks don't have the jupyter events module defined, so I've had to add a shim whereby we create an equivalent object at `window.events` if necessary. I hope that makes sense. I *think* this now all works ok...

jfbercher:
@jcb91 All that seems really great! What a nice idea to have define your own `events` object. 
I will try to explore and test all that this week or so.

jfbercher:
I have finally reviewed and tested the PR. 
It is indeed very nice, I like it! :1st_place_medal: 

I just propose some small adaptations: 
- https://github.com/ipython-contrib/jupyter_contrib_nbextensions/pull/1031/commits/3e509cd2e83440b6445a61456f7a5a732aa5d35a Only add collapse controls to items with descendants
- I tested synchronization with collapsible headings in both ways. It seems that `collapse_by_id` function must be accessible from outside the main function for collapsible --> toc to work. Addressed in https://github.com/ipython-contrib/jupyter_contrib_nbextensions/pull/1031/commits/a3172d464a893d066ac8ae2f47d00f6a6a9a5607
- `collapse_to_match_collapsible_headings` value was not updated from the configurator because the value stored in notebook metadata was preferred. Changed in https://github.com/ipython-contrib/jupyter_contrib_nbextensions/pull/1031/commits/6e7e5627e679c327a04f4909aac271540074a247 but probably we must think of a better way to process parameters that must be always be taken from system and those that can be fixed per notebooks
- finally, I added a span tag to embed the full item in toc, which enable to highlight the full line instaed of just the html link. https://github.com/ipython-contrib/jupyter_contrib_nbextensions/pull/1031/commits/1d3a8202564f600e37d6a94a6780e52df4ae7666
I also tested the export to html and it seems to work as it should. 

Really nice @jcb91!. Thanks
I think we can merge if you agree with  the small modifications above.

jcb91:
> I have finally reviewed and tested the PR. 
> It is indeed very nice, I like it! :1st_place_medal: 

Haha great, thanks!

> I just propose some small adaptations: 

Ok, all of these make sense to me.

I've rebased the changes to simplify the now rather-tangled commit history, and made a couple of extra very small changes as part of the rebase:
 * removed some extra css rules about menu padding, which seem to have got added accidentally
 * and add a missing `a->span` alteration for the custom css
 * moved *all* function definitions outside the main `table_of_contents` function, for clarity.

but then I think we should be good to merge, assuming I haven't broken anything with my rebase :wink:

jfbercher:
Tested this morning. It is great!

I would have a further modification. I realized yesterday that the toc links, though working in Chrome/Chromium did not work anymore in Firefox (Linux/last version). This may be related to the point mentioned in some docs : 

> "However, bubbling of a click event will not cause an <a> element to initiate navigation as if a real mouse-click had been received." 

eg [here](http://mdn.beonex.com/en/DOM/element.click.html) or [here](https://stackoverflow.com/questions/7999806/jquery-how-to-trigger-click-event-on-href-element). A possible solution to simulate the click using the mouse event is mentioned [here](https://stackoverflow.com/questions/7999806/jquery-how-to-trigger-click-event-on-href-element). Since all the links are on the same page, I suggest to the [scrollIntoView](https://developer.mozilla.org/en/docs/Web/API/Element/scrollIntoView) method to scroll the element with the given id  into view.  
```
 document.getElementById(trg_id).scrollIntoView(true)
```
I have tested with this modification and it works at least on Chrome/Firefox/Gnome web browser, both in live notebook and after html export (just edited the html to load the current `toc2.js` on disk). In addition, I think that this has good chances to fix  #1023 

I have also updated the README. 
Can you check that @jcb91 and give the (hopefully last) "go" for merging?

jfbercher:
- updated `toc2.tpl` to include font-awesome; otherwise carets were not displayed in firefox when exporting to html
- updated again `main.css`. I think that there were things lost somewhere in the various commits/rebase. Check it @jcb91 , there are certainly things that can be improved.

src/jupyter_contrib_nbextensions/nbextensions/toc2/main.css
```diff
@@ -69,6 +72,7 @@ padding-left: 20px;
 
 .toc ul.toc-item {
     list-style-type: none;
+    display: block;
```
jcb91: this is redundant, display is block anyway

src/jupyter_contrib_nbextensions/nbextensions/toc2/main.css
```diff
@@ -4,15 +4,18 @@
 /*background color for links when you mouse over it 
 This style is used on export, but can be overwritted in the livenotebook 
 by selecting a color in the nbextension-configurator*/
-#toc-wrapper li > span:hover {
+
+
+
+#toc-wrapper ul li span:hover {
```
jcb91: this isn't needed `li span` is sufficient & more efficient, since every `li` is in a `ul` anyway, while the `>` prevents nested child spans from being selected (fewer css redraws, more efficient, hopefully)

src/jupyter_contrib_nbextensions/nbextensions/toc2/main.css
```diff
@@ -4,15 +4,18 @@
...
   background-color: #DAA520;
 }
 
 #toc-level0  a {
   /*color: #333333;*/  /* now specified in js */ 
   text-decoration: none;
 }
-#navigate_menu li > span:hover {background-color: #f1f1f1}
+#navigate_menu ul li span:hover {background-color: #f1f1f1}
```
jcb91: see above

src/jupyter_contrib_nbextensions/nbextensions/toc2/main.css
```diff
@@ -145,9 +149,18 @@ padding-left: 20px;
...
 }
+*/
+.toc-item ul, li { margin:0; padding:0 }
+.toc-item li > span { display:block }
+.toc-item li > span { padding-left:0em }
+.toc-item li li > span { padding-left:1em }
+.toc-item li li li > span { padding-left:2em }
+.toc-item li li li li > span { padding-left:3em }
+.toc-item li li li li li > span { padding-left:4em }
+.toc-item li li li li li li > span { padding-left:5em }
```
jcb91: Why are we adding extra padding, and why separately for various nesting levels?! If desired (is the regular indenting not sufficient?), could this not be handled by adjusting the `ul`'s padding?

Also note that this has set margin & padding for *all* `li` elements to 0 - I guess that was unintentional, and should have been `.toc-item li`?

src/jupyter_contrib_nbextensions/nbextensions/toc2/main.css
```diff
@@ -38,13 +41,13 @@ padding-left: 20px;
...
-  padding-left: 2px;
+  padding-left: 0px;
   clear: both;
   list-style-type: none;
 }
 
-#navigate_menu-level0  {padding-left: 10px;}
-#navigate_menu-level0 ul  {padding-left: 10px;}
+#navigate_menu-level0  {padding-left: 0px;}
+#navigate_menu-level0 ul  {padding-left: 0px;}
```
jcb91: I'm not familiar with any of the navigate_menu stuff (haven't had it enabled), but now that I've tried it, it seems a bit weird, and is coming out resizable. Perhaps it deserves altering to make it look like all the other notebook menus?

jcb91:
> I would have a further modification. I realized yesterday that the toc links, though working in Chrome/Chromium did not work anymore in Firefox (Linux/last version). This may be related to the point mentioned in some docs

Good catch, good fix! :+1: 

> I have also updated the README.

:+1: :smile: 

> updated `toc2.tpl` to include font-awesome; otherwise carets were not displayed in firefox when exporting to html

Another good catch, good fix! :+1: 

> updated again `main.css`. I think that there were things lost somewhere in the various commits/rebase. Check it @jcb91 , there are certainly things that can be improved.

This, I'm less certain about, though likely it's partly just that I haven't understood. I've added a few comments with questions for clarifications

> Can you check that @jcb91 and give the (hopefully last) "go" for merging?

Haha, we're so close :smile:!

src/jupyter_contrib_nbextensions/nbextensions/toc2/main.css
jcb91: although, any such modification should probably go into another, separate PR :sweat_smile:

jfbercher: Yes!

src/jupyter_contrib_nbextensions/nbextensions/toc2/main.css
jfbercher: ok corrected

src/jupyter_contrib_nbextensions/nbextensions/toc2/main.css
jfbercher: ok

src/jupyter_contrib_nbextensions/nbextensions/toc2/main.css
jfbercher: I wanted to highlight the entire line containning the section item on hover. 
I used the answer [here](https://stackoverflow.com/questions/9537444/highlight-entire-line-in-a-nested-list-on-hover) and [here](http://jsfiddle.net/dYdQB/). There might be a better way to do it but didn't figured it. 

- ok for `.toc-item li`

jfbercher: thanks!

jfbercher:
I think this is almost ready now. All local tests work fine. Let us merge and see if everything is also ok on the live version. Thanks to @jcb91 for his great work.

PR #1032, opened 2017-07-23
[toc2] use amd structure to avoid polluting the global namespace

jcb91:


jcb91:
I think this should work ok, but haven't tried html export - any chance you could give it a try at some point @jfbercher? I'm thinking I'll try to move toc2 over to using the inliner class which collapsible_headings is currently using, so that we can more easily export using both of them, so this would be useful to have before attempting that :)

jfbercher:
@jcb91 I have tested conversion today and apart a require error that appears in the console, it works! 
That is very nice to have considered that point; Indeed I was annoyed with this pollution of the namespace. I'll recycle the idea for `jupyter_latex_envs` .  I have pushed a small PR to your repo for that. 

There is also a small backward compatibility issue to address. In old html conversions, the js tries to load directly `table_of_contents()`, not `toc2.table_of_contents()`...

jfbercher:
Commit https://github.com/ipython-contrib/jupyter_contrib_nbextensions/pull/1032/commits/466a78b5f251eac032cef55345631715ecc0506f does not address the backward compatibility issue. If we still want to support html files converted before the conversion of toc2 to a module -- which is a nice thing, I am afraid that we will have to define this module with some other name (eg `toc2_lib` or whatever) and keep a `toc2.js` that would read this module and export a `table_of-contents` function. 
The livenotebook could work only with the module, so that namespace will not be  polluted. Thoughts?

jcb91:
> If we still want to support html files converted before the conversion of toc2 to a module -- which is a nice thing, I am afraid that we will have to define this module with some other name (eg `toc2_lib` or whatever) and keep a `toc2.js` that would read this module and export a table_of-contents function.
The livenotebook could work only with the module, so that namespace will not be polluted. Thoughts?

I agree, backwards compatibility is a good thing to keep. However, I don't think we need any extra modules or anything - I think we can just fix this in `toc2.js` using something like:

```javascript
if (!liveNotebook) {
    // export main function for backwards compatibility
    window.table_of_contents = table_of_contents;
}
```

After all, adding just the single name `table_of_contents` to the global namespace in non-live notebooks isn't *so* bad :wink:

jfbercher:
@jcb91 Nice idea. I tested it but it did not worked, probably because of a race condition -- `table_of_contents`   was not defined when reached. 
I have pushed a commit where the `require` ensures that the library is fully loaded before using `table_of_contents`. In my tests, this works. 
```
if (!liveNotebook) {
    // export main function for backwards compatibility
    function table_of_contents(cfg, st) {
        return require(['nbextensions/toc2/toc2'], function(toc2) {
            return toc2.table_of_contents(cfg, st);
        });
    }
}
```

jfbercher:
@jcb91 For me this is working. Have you had a look to the mast commits? Do you think we can  merge?

jcb91:
Apologies @jfbercher, I've been away from a computer for a few days. I'll likely get a chance to look at things wednesday. This looks to me like it'll work, but it's a bit convoluted trying to figure out what gets defined in what order, so I'd like to simplify it a little if that's ok? Anyway, feel free to merge away if it's working ok, I can always make another pr later to satisfy my ocd ;)

jfbercher:
@jcb91 Sure. I have a little time myself. Simplifications will be welcome. We will merge when ready!

jcb91:
ok, I've attempted to rebase this onto the current master, and also made a stab at simplifying the config loading. Currently I've just replicated the existing behaviour in terms of what's per-notebook and what's not, but I have a feeling some (e.g. threshold?) may have been miscatergorized. Not sure. I've also removed some unused items: the `liveNotebook` variable in `main.js` (which only gets loaded in live notebooks anyway), plus the settings `config_loaded`, `extension_initialized`, `nbcontainer_marginleft`, `nbcontainer_marginright` and `nbcontainer_width`, which were all unused...

src/jupyter_contrib_nbextensions/nbextensions/toc2/main.js
```diff
@@ -47,74 +42,49 @@ define([
...
               },
               collapse_to_match_collapsible_headings: false,
-        skip_h1_title: false,
 }
+    // default values for per-notebook configurable parameters
+    var metadata_settings = {
+        nav_menu: {},
+        number_sections: true,
+        sideBar: true,
+        skip_h1_title: false,
```
jfbercher: I think that skip_1_title is a system-wide parameter.

jcb91: To me, it's per-notebook, as whether it makes sense depends on the notebook content?

jcb91:
> Will have to rebase and merge your version :-)

Ah sorry, it always gets messy here! :laughing:

jfbercher:
@jcb91 Just tested the live-version. 
 [This](https://rawgit.com/jfbercher/jupyter_latex_envs/master/src/latex_envs/static/doc/latex_env_doc.html) works. But it seems that an error has appear [here](https://htmlpreview.github.io/?https://github.com/bianbian2010/temp/blob/master/toc2-example.html) which prevents the toc to display?

jcb91:
> But it seems that an error has appear [here](https://htmlpreview.github.io/?https://github.com/bianbian2010/temp/blob/master/toc2-example.html) which prevents the toc to display?

Hmm, curious. It seems that `draggable` isn't present, from which I deduce that at toc's execution, jquery-ui hasn't yet been loaded, and indeed it doesn't appear in the page source, so I can only assume that something else (mathjax? bootstrap?) loads it later. I suppose the answer to this in a live notebook would be to require `jquery-ui`, but it doesn't seem to get registered correctly by the require library in the htmlpreview page, so that wouldn't work there. The only other alternative I can think of would be to just set an interval, and poll for its presence, but that seems a bit ridiculous. Any thoughts?

jfbercher:
I was investigated too. Either it is jquery-ui delay for loading or a delay in inserting toc-wrapper in the DOM. In such case we would have to [wait for insertion](https://stackoverflow.com/questions/5525071/how-to-wait-until-an-element-exists), but seems a bit complicated; or add a timeout but seems a bit ridiculous too.

src/jupyter_contrib_nbextensions/nbextensions/toc2/main.js
jfbercher: Yes, but we have nothing to change the behavior per notebook, except directly editing the metadata.

jfbercher:
It it were just a DOM inserting delay, we could add the `.draggable()` when creating the element (before insertion).

jcb91:
>  or a delay in inserting toc-wrapper in the DOM. In such case we would have to wait for insertion, but seems a bit complicated;

> It it were just a DOM inserting delay, we could add the .draggable() when creating the element (before insertion).

Not sure I understand you here?

jfbercher:
I meant that things happens as 
a) var toc_wrapper = $('\<div id="toc-wrapper"\/\>')
    .append( ....
b)  $("body").append(toc_wrapper);
c)  $('#toc-wrapper').draggable({...
Perhaps that in c),  $('#toc-wrapper') doesn't exist yet? 
Since it worked previously and still works in other situations than the example reported [above](https://htmlpreview.github.io/?https://github.com/bianbian2010/temp/blob/master/toc2-example.html) leaves the hypothesis opened.

jcb91:
Ah, ok. Well, from a breakpoint set at the draggable call, I see that when it happens, `$.ui` is still undefined, so I think at the very least we have 2 problems! 😅

jfbercher:
You should be right. Indeed, the fact that the target is not defined does not rise the error `$(...).draggable is not a function`.

Issue #1041, opened 2017-08-10
html_embed doesn't work for attached images


waynebliu:
!jupyter nbconvert --to html_embed embedhtmltest.ipynb

[NbConvertApp] Converting notebook embedhtmltest.ipynb to html_embed
Traceback (most recent call last):
  File "/home/wayne/miniconda3/envs/mypy/bin/jupyter-nbconvert", line 11, in <module>
    load_entry_point('nbconvert==5.2.1', 'console_scripts', 'jupyter-nbconvert')()
  File "/home/wayne/miniconda3/envs/mypy/lib/python3.6/site-packages/jupyter_core/application.py", line 267, in launch_instance
    return super(JupyterApp, cls).launch_instance(argv=argv, **kwargs)
  File "/home/wayne/miniconda3/envs/mypy/lib/python3.6/site-packages/traitlets/config/application.py", line 658, in launch_instance
    app.start()
  File "/home/wayne/miniconda3/envs/mypy/lib/python3.6/site-packages/nbconvert/nbconvertapp.py", line 313, in start
    self.convert_notebooks()
  File "/home/wayne/miniconda3/envs/mypy/lib/python3.6/site-packages/nbconvert/nbconvertapp.py", line 481, in convert_notebooks
    self.convert_single_notebook(notebook_filename)
  File "/home/wayne/miniconda3/envs/mypy/lib/python3.6/site-packages/nbconvert/nbconvertapp.py", line 452, in convert_single_notebook
    output, resources = self.export_single_notebook(notebook_filename, resources, input_buffer=input_buffer)
  File "/home/wayne/miniconda3/envs/mypy/lib/python3.6/site-packages/nbconvert/nbconvertapp.py", line 381, in export_single_notebook
    output, resources = self.exporter.from_filename(notebook_filename, resources=resources)
  File "/home/wayne/miniconda3/envs/mypy/lib/python3.6/site-packages/nbconvert/exporters/exporter.py", line 172, in from_filename
    return self.from_file(f, resources=resources, **kw)
  File "/home/wayne/miniconda3/envs/mypy/lib/python3.6/site-packages/nbconvert/exporters/exporter.py", line 190, in from_file
    return self.from_notebook_node(nbformat.read(file_stream, as_version=4), resources=resources, **kw)
[...]

PR #1044, opened 2017-08-12
Fix html_embed exporter

juhasch:
Fixes issue #1041:
Correctly embed images already embedded in the notebook itself.

### DIFF
--- a/src/jupyter_contrib_nbextensions/nbconvert_support/embedhtml.py
+++ b/src/jupyter_contrib_nbextensions/nbconvert_support/embedhtml.py
@@ -4,6 +4,7 @@ import base64
 import re
 from nbconvert.exporters.html import HTMLExporter
 from ipython_genutils.ipstruct import Struct
+import os
 
 try:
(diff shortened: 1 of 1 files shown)


gabyx:
was the path already in the metadata?
I thought there is not `path` variable in the resource metadata :-)
if so, this is a cool fix :-) fixes my export button top

juhasch:
This works at when calling nbconvert from command line.
I don't have notebook 5.1 running, so this might not work with your extension. However, I would assume this is a bug in the notebook then.

gabyx:
ah, ok, hm..., so this merge will 
make my extension not working before the notebook bug is not resolved...
maybe I have some time

juhasch:
Hmm, looks like the notebook calls `from_notebook_node()` and does not provide the path. In my eyes it should. This means I need to add a check in the exporter, and your problem with the extension is still not solved.

gabyx:
Created a pull request to fix that:
 https://github.com/jupyter/notebook/pull/2753

PR #1050, opened 2017-08-16
[codefolding] fix restoration of nested folds

### LINKED ISSUE #1046: Codefolding extension does not restore folding with nested expression
Consider the following MWE:

```
# testsets
a = 2
def foo():
    pass
```
which appears, when the Codefolding extension is loaded, as:
image 1

- If I fold the code at the line `def foo():`, save and reload, I get the folding back as expected.
- If I fold the code at the first line, save and reload, I still get the folding back as expected.
- However, if I fold at the third line and then at the first line, save and reload, the folding is *not* restored.

Executing `IPython.notebook.get_cells()[0].metadata.code_folding` from the console, I do get what to me looks like what I should get when the code is folded at both positions:
image 2
I still get this when I reload the page and the code does not look folded, so I guess there is something wrong in the code that should restore the folding?

I see this on both Chrome and Opera under Windows 10, with the jupyter server run from bash on ubuntu on windows.

jcb91:
by processing them in reverse order

fixes #1046

### DIFF
--- a/src/jupyter_contrib_nbextensions/nbextensions/codefolding/main.js
+++ b/src/jupyter_contrib_nbextensions/nbextensions/codefolding/main.js
@@ -148,6 +148,31 @@ define([
     }
 
     /**
+     * Restore folding status from metadata
+     * @param cell
+     */
+    var restoreFolding = function (cell) {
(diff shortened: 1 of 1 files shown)


juhasch:
Cool! Thanks for fixing :+1:

PR #1052, opened 2017-08-19
Changed regex parsing in embedhtml.py to XML parsing with lxml, regex…

gabyx:
regex can catch all sorts of stupid things.

+ made the logging more consistently.


gabyx:
It seems the test do not work since lxml >= 3.8.0 is introduced?
How should we change travis? or this pull request?

soamaven:
Hi thanks for working this. I think I brought something similar up about a year ago, glad to see some work on it. I actually took a similar approach and stalled.

Rather than using the lxml library is there a reason not to just use the python standard HTML library? I was able to not have to treat the different file formats as cases (enabling .gif export for example).

gabyx:
you mean: xml.etree.ElementTree.XMLParser, had no luck to parse the HTML
I tried the https://docs.python.org/3.5/library/html.parser.html?highlight=htmlparser
with tracking the position, extracting all img tags and then afterwards replace it,
but the getpos() function returns weird positions for end tags, and somekind. I think going with the lxml version is the best. Or is there another parser which jupyter and other preprocessors use??

soamaven:
That is where I got stalled too! My code comments from back then indicate the end tags were acting weird!

Curious, will your implementation preserve `img` attributes? I vaguely remember errors in the previous embedhtml exporter when attribute such as size, etc were given.

One last comment, do you think an embedhtml.py preprocessor should follow easily from this? I'd like to be able to embed images into slides, and suspect the most straight forward was to do that would be to preprocess and then export to slides in one command if possible.

gabyx:
I think lxml preserves XML tags, because I only load it, and manipulate it, I mean it should, otherwise lxml would be a awful implementation

gabyx:
Cool, good work, I will look over :-) sounds reasonable to make also a preprocessor ! :-)

src/jupyter_contrib_nbextensions/nbconvert_support/embedhtml.py
```diff
@@ -49,29 +53,37 @@ def replfunc(self, match):
             with open(filename, 'rb') as f:
                 data = f.read()
 
-        self.log.info("embedding url: %s, format: %s" % (url, imgformat))
```
juhasch: I think it is helpful to have a log of the embedded images. Better move it to `self.log.debug`

src/jupyter_contrib_nbextensions/nbconvert_support/embedhtml.py
```diff
@@ -49,29 +53,37 @@ def replfunc(self, match):
...
-            img = '<img src="data:application/pdf;base64,' + \
-                b64_data + '"'
+            prefix= "data:application/pdf;base64,"
         else:
-            img = '<img src="data:image/' + imgformat + \
-                ';base64,' + b64_data + '"'
-        return img
+            prefix = "data:image/" + imgformat + ';base64,'
 
+        node.attrib["src"] = prefix + b64_data
```
juhasch: Can you preserve the original attributes of the node and only change the `src` attribute ?

juhasch:
Thanks for working on this. It is much welcome.

Regarding a preprocessor: With the new ability of the notebook to embed images as attachment, maybe that would be the better approach  for a preprocessor ?
Also, you will want to support `img` tags and markdown there.

soamaven:
Thanks @juhasch for the suggestion. I am working on it on @gabyx 's fork [here](https://github.com/gabyx/jupyter_contrib_nbextensions/pull/1) now; it should handle `img` tags already.

I agree it should handle markdown image references also. I will try to incorporate that.

gabyx:
Cool, I think this imrpovements will be much appreciated, thanks for the effort. I merged the pull request

gabyx:
what shall we do with the lxml stuff? Is there a better alternative? I mean it gives clean and concise code :-)...

soamaven:
@juhasch I didn't before understand what you meant by attachments. I do now. Yes that *may* be a better approach, specifically for MD image references. However, I don't know if attaching html `img` tags is the best approach, as all the attribute data would be lost -- I don't think jupyter MD can't accept as many attrs as native html. I personally use html to control look often.

gabyx:
what are attachments? =)?

soamaven:
@gabyx Having just learned about them myself, it looks like attachments enable user familiar actions like drag-and-drop for images into MD cells, and the "insert  image" from the edit menu. It adds the image as base64 data to the metadata of the cell itself.

Anyone feel free to correct me.

juhasch:
@soamaven  Exactly. This is great if you want to have everything in one document, but might lead to large documents.

I personally would love a preprocessor that reads images that are located locally on disk, scales them to a decent size, and embeds them into a new notebook. This would make distributing notebooks so much easier.

juhasch:
@gabyx Btw. have you considered using BeautifulSoup ?
```
from bs4 import BeautifulSoup
import re
soup = BeautifulSoup(html_source, 'html.parser')
for tag in soup.find_all(re.compile("^img")):
    print(tag.name)
```

soamaven:
@juhasch 

>You actually could also add your own attributes to md image references, like size, since the preprocessor works before the conversion of md->html.

Ah, but since I can't, for example, set the image size with md, how will the preprocessor know what size to make the embedded html? I would aim to preserve look as much as possible. Looks like we could store attr data in the`attachmnents` JSON data? However it seems pretty hard-coded how attachments are [rendered in a textcell](https://github.com/jupyter/notebook/pull/621/files#diff-70e05bf8611faf49d3654451cf6dd648R361) no? js is not my strength yet.

Once a new parser is settled on, we can pursue scaling and metadata embedding. I have in mind something more user-friendly -- everyone will have their own opinions on what size is acceptable, so that variable should be made easily accessed by user, no?

I'll quickly suggest eventually collecting all this functionality into the newer nbextension `export embedded`, and letting various options be set via the nbconfigurator.

I think it's time to move some of this to a new issue? Here is a list so we can keep it all straight:

**Current Implementation**
- [ ] Decide which html parser to use for `exporthtml.py`
- [ ] Ensure preservation of original html `img` attributes in current `exporthtml.py`
- [ ] Move `exporthtml.py` logging to `self.log.debug`

**Future Implementation** _Date Unknown_
- [ ] Make use of attachments for markdown image references when pre-processing
  - Can all base64 blobs be put into attachments?
- [ ] Be able to choose scaling/compression of images when embedding (even maybe just max / normal / lossless ?). 
   - This could be cell specific using cell metadata.

juhasch:
@soamaven Yes, better to have a separate issue. I created issue #1064. Let's continue the preprocessor discussion there. Maybe you a different functionality in mind ?

gabyx:
Could we agree on using lxml? If so, we should make the conda build passing, since there is a problem with the lxml (travis config problem?)

jcb91:
So I don't follow why we can't use the standard lib html parser? I get that it doesn't cope with invalid html, but is it possible to get invalid html from the md->html exporter? I'd rather not add any extra dependencies if possible, but if anything, I think it's better to opt for lxml over bs, as @soamaven describes.

jcb91:
Also, from https://travis-ci.org/ipython-contrib/jupyter_contrib_nbextensions/jobs/270165235#L778-L810
it seems that some merge lines have got included by accident?

src/jupyter_contrib_nbextensions/nbconvert_support/embedhtml.py
```diff
@@ -1,9 +1,14 @@
...
 
 import base64
+<<<<<<< HEAD
+import lxml.etree as et
+from nbconvert.exporters.html import HTMLExporter
+=======
 import os
 import re
 
+>>>>>>> currentMaster
```
jcb91: this is a merge which hasn't been resolved correctly!

src/jupyter_contrib_nbextensions/nbconvert_support/embedhtml.py
```diff
@@ -25,56 +30,74 @@ class EmbedHTMLExporter(HTMLExporter):
...
+            raise ValueError("""Could not find attachment for image '%s'
+                                 in notebook""" % imgname)
+=======
                     img = '<img src="data:' + imgformat + \
                           ';base64,' + b64_data + '"'
                     return img
             raise ValueError(
                 'Could not find attachment for image "%s" in notebook' %
                 imgname)
+>>>>>>> currentMaster
```
jcb91: another unresolved merge

gabyx: ups, men, I thought sourcetree does it right... (chose take mine) ... dont know what happend.
scusi

gabyx:
I tried the https://docs.python.org/3.5/library/html.parser.html?highlight=htmlparser
with tracking the position, extracting all img tags and then afterwards replace it,
but the getpos() function returns weird positions for end tags ... (its not about invalid HTML together with javascript) @soamaven  got stuck with the same problems.

I mean its certainly possible to write this parser without lxml, but it gets so ugly, especially when trying to preserve the tags, the only way I saw was to store the whole html string internally and for all  tags do the replacement, going back to regex is probably no solution, since its not the right tool for xml...

If we use PIL for image converting in #1067, the why cant we use lxml, its fast and reliable...
If we insist on not having lxml, I will have a look at the python internal html parser again,

src/jupyter_contrib_nbextensions/nbconvert_support/embedhtml.py
jcb91: No worries, it happens to all of us from time to time 😜

jcb91:
> the only way I saw was to store the whole html string internally and for all  tags do the replacement

Yeah, that sounds like the right approach to me?  Do you have an example implementation attempt (even non-working!)?

As for the conda build, the reason it's currently failing is that you haven't specified `lxml` as a requirement in the `conda.recipe/meta.yaml` (conda obeys installs the things specified by its recipe, regardless of `setup.py` content).

gabyx:
I had one, which I deleted, I can give it another try, an report back :)

------ Originalnachricht ------
Von: "Josh Barnes" <notifications@github.com>
An: "ipython-contrib/jupyter_contrib_nbextensions" 
<jupyter_contrib_nbextensions@noreply.github.com>
Cc: "Gabriel Nützi" <gnuetzi@gmail.com>; "Mention" 
<mention@noreply.github.com>
Gesendet: 31.08.2017 15:21:47
Betreff: Re: [ipython-contrib/jupyter_contrib_nbextensions] Changed 
regex parsing in embedhtml.py to XML parsing with lxml, regex… (#1052)

juhasch:
@gabyx Any progress on this ?
It would be nice if we could merge the exporter part of your PR.

gabyx:
Try to give it another shot today  :-)

gabyx:
Here is my shot at it:
https://github.com/gabyx/jupyter_contrib_nbextensions/pull/2
the problem is, that this implementation, which concats all HTML strings while performing transformations on  tags,  is really slow!

For me this is no really good alternative if it cannot be made faster.
(remember get_pos() does not help, because it is acting strange, so caching the positions and afterwards transform  tags does not really work)

juhasch:
Thanks. I looked at it and lxml is much nicer (and faster) IMO. Let's see what @jcb91 thinks.

As for this PR:
 * lint is currently broken, so one check will fail. There is #1097 to fix it.
 * you will need to run `isort` to survive lint (or wait for the error messages from Travis and fix them...)
 * please remove the preprocessor. We can discuss it on #1067 or you can make a new PR
 * The conda recipe file `meta.yaml` requires the additional dependency on lxml

jcb91:
> I looked at it and lxml is much nicer (and faster) IMO. Let's see what @jcb91 thinks.

Fair enough, if it makes a big difference, I don't think it's a big problem to add lxml, as it's fairly lightweight and well-supported. Let's run with it.

gabyx:
How should we install lxml on appveyor, -> currently not working
setup.py contains lxml >= 3.8.0 which works on linux
how to add the appropriate stuff in the tox.ini?
can we somehow install from a wheel,  suppose we need a different wheel for different pythons?

jcb91:
> How should we install lxml on appveyor, -> currently not working
> setup.py contains lxml >= 3.8.0 which works on linux
> how to add the appropriate stuff in the tox.ini?
> can we somehow install from a wheel, suppose we need a different wheel for different pythons?

Ugh, so essentially we want a painless Windows install for lxml, but there isn't one. This is exactly the kind of thing it's best to avoid in dependencies :disappointed: I'm not immediately sure how to proceed, will have a read...

gabyx:
travis seems ok, on windows, the test_embed asserts where the two notebooks are not the same 
`assert statinfo_e.st_size > statinfo.st_size`
hm...

jcb91:
Oh dear. That's weird. I have a Windows machine I can dig out to investigate, but it'll take a while. I guess this is why we have tests though 😲

juhasch:
Tried it today on a 64 bit Windows 10 machine. The test went through without problems.

jcb91:
This gets weirder and weirder. I tried on a 64bit win 8.1, and got the same failure as on appveyor, hough from all I've been able to gather, the process works ok when I call it outside of tests 😕

juhasch:
Strange indeed. I have been using `nosetests`. 

Can you check what the file sizes are, i.e. is the error real (no embedding) or is it a side effect ?

jcb91:
File size stats were definitely wrong (smaller in 'embedded' version), though I didn't get to compare actual text. I won't be able to look again for a couple of days though.

gabyx:
thanks for testing, could it be that the file size stats are buggy under windows. maybe some file locks or  not properly closed files? strange 

Von meinem iPhone gesendet

src/jupyter_contrib_nbextensions/nbconvert_support/embedhtml.py
```diff
@@ -25,56 +25,65 @@ class EmbedHTMLExporter(HTMLExporter):
...
-        embedded_output = regex.sub(self.replfunc, output)
+        # Parse HTML and replace <img> tags with the embedded data
+        parser = et.HTMLParser()
+        root = et.fromstring(output, parser=parser)
+        nodes = root.findall(".//img")
+        for n in nodes:
+            self.replfunc(n)
+
+        # Convert back to HTML
+        embedded_output = et.tostring(root, method="html")
```
jcb91: This should use `root.getroottree()` to avoid losing the doctype tag.

juhasch:
It is our old friend `CRLF`. LXML under Windows somehow adds this and therefore fools the test.
I have a new `test_exporters.py` in Pr #1107 . Please try this one.

juhasch:
Sorry bout that. @jcb91 convinced me to look for the real solution 👍

gabyx:
ok, easy, we can also revert that again!
am I correct: on windows, the html and embed_html make both CRLF endings?

juhasch:
`et.tostring` returns bytes instead of an unicode str if no encoding is given.
str -> CRLF
bytes -> LF only

So the HTML exporter gives you CRLF on Windows, and the html_embed exporter should too by returning an unicode str instead of bytes.

jcb91:
Lxml prefers to deal with byte strings in general, so it may be worth encoding the input as a byte string in addition to unicoding the output. Also could we specify utf-8 rather than just unicode, since that's what's in the meta charset tag?

gabyx:
why not `encoding="utf-8"`
where did you find the documentation on that ? lxml docs are not so good at all hm..

gabyx:
strange behaviour:
```python
s = "<root>asd\r\n</root>"
print(s)
t=ET.fromstring("<root>asd\r\n</root>")
ET.tostring(t,encoding="utf-8")
```

```
<root>asd
</root>
b'<root>asd\n</root>'
```

jcb91:
😮

juhasch:
The lxml documentation states:
You can also serialise to a Unicode string without declaration by passing the unicode function as encoding (or str in Py3), or the name 'unicode'. This changes the return value from a byte string to an unencoded unicode string.

gabyx:
```
s = b"<root>asd\r\n</root>"
print(s)
t=ET.fromstring(s)
ET.tostring(t,encoding="unicode")
```
dont get why it removes `\r\n` ....??

gabyx:
Aha `&#13;`hm... lol

juhasch:
On Windows nbconvert produces CRLF line endings. On Unix/Mac you get LF line endings.
However, if you return bytes instead of str to nbconvert, it looks like nbconvert does not add CRLF and only writes LF line endings on Windows.

gabyx:
but as I understand: the lxml parsing, and writing should not change the lineendings, what ever it was right?

gabyx:
When I input a CRLF file:
```python
s = open("test.txt","rb")
print(s)
p = ET.HTMLParser()
t=ET.parse(s, parser=p)
with open("out.txt","wb") as f:
    f.write(ET.tostring(t,encoding="utf-8"))
```
I get this
```
<!DOCTYPE html PUBLIC "-//W3C//DTD HTML 4.0 Transitional//EN" "http://www.w3.org/TR/REC-html40/loose.dtd">
<html><body><root>&#13;
asd&#13;
</root>&#13;
</body></html>
```
This is not what we want? We want to keep the `\r` ....

gabyx:
Maybe we can do the following: in `embed_html.py`

```python
parser = et.HTMLParser()
root = et.fromstring(output.replace(b"\r",b""), parser=parser)
....
embedded_output = ET.tostring(root.getroottree(), method="html",  encoding="utf-8").replace(b"\n",os.linesep.encode("utf-8"))
```

juhasch:
`encoding="utf-8"` is not a valid parameter according to documentation, see http://lxml.de/api/lxml.etree-module.html#tostring

And if you pass `encoding="unicode"` you get a `str` returned which is what nbconvert assumes.

gabyx:
ok, lets wait for the tests, to see if the "unicode" does the job
lxml html parsing converts now CRLF to `&#13;\n` (probably only on linux)
anyway, there are more bytes... for the assert

jcb91:
I'm guessing it's a variant of the classic unicode/bytestring problem. See https://ci.appveyor.com/project/jcb91/jupyter-contrib-nbextensions/build/lxml-27/job/f35fm40m2t42kvp6#L139 for a diff, showing a bunch of whitespace changes

juhasch:
Let's change the test. The generated html files are valid, I think, so we should check these. 
How about testing the src attribute itself:
```Python
        with open(nb_dst_filename, 'r') as f:
            output = f.read()
        parser = et.HTMLParser()
        root = et.fromstring(output, parser=parser)
        nodes = root.findall(".//img")
        for n in nodes:
            url = n.attrib["src"]
            assert url.startswith('data')
```

gabyx:
Rewrote the tests. HTML is always parsed to verify correctness...

juhasch:
OK, we are getting close :+1:. Only lint/isort is complaining anymore.

gabyx:
finally, it comes to an end ;-)
hopefully, maybe review the test_exporter.py

jcb91:
So is this finally ready to merge? Seems like it to me :+1:

juhasch:
I am a little busy currently, but what I have seen LGTM. So :+1:

PR #1054, opened 2017-08-20
Codefold preprocessor improvements

gabyx:
This improvements make the preprocessor much more stable:
There were several bugs, such that outer levels did not get collapsed when exporting.
Now the coldfolding during export is closer to the experience in the notebook (if not totally the same)
**Input:**
Notebook
**Output:**
HTML

### DIFF
--- a/src/jupyter_contrib_nbextensions/nbconvert_support/pre_codefolding.py
+++ b/src/jupyter_contrib_nbextensions/nbconvert_support/pre_codefolding.py
@@ -32,13 +32,13 @@ class CodeFoldingPreprocessor(Preprocessor):
         """
         Remove folded lines and add a '<->' at the parent line
         """
-        self.log.debug("CodeFoldingPreprocessor:: folding at: %s" % folded)
+        # self.log.debug("CodeFoldingPreprocessor:: folding at: %s" % folded)
         lines = cell.splitlines(True)
 
         if folded[0] == 0 and (lines[0][0] == '#' or lines[0][0] == '%'):
             # fold whole cell when first line is a comment or magic
-            p = lines[0].rstrip('\n') + self.fold_mark + '\n'
-            return p
+            # self.log.debug("fold whole cell")
+            return lines[0].rstrip('\n') + self.fold_mark + '\n'
         
         foldIndent = 0
         fold = False
@@ -50,7 +50,7 @@ class CodeFoldingPreprocessor(Preprocessor):
             # fold indent level
             lstrip    = l.lstrip(r' \t') # strip tabs and spaces
             indent    = len(l) - len(lstrip)
-            isSkipLine = lstrip[0] == "#" or lstrip in ["\n" , "\r\n"]
+            isSkipLine = lstrip[0] == "#" or lstrip in ["\n"] # is it a comment or an empty line
 
             if indent <= foldIndent and not isSkipLine:
                 # folding finished, when we reached no skip line on an upper 
@@ -76,7 +76,7 @@ class CodeFoldingPreprocessor(Preprocessor):
                 else:
                     skipped += l
 
-            self.log.debug("%02i, %02i < %02i, %i, %i,'%s' ", i, indent, foldIndent, updateIndent, fold, l[0:indent+10].strip("\r\n"),)
+            #self.log.debug("%02i, %02i < %02i, %i,'%s' ", i, indent, foldIndent, fold, l[0:indent+10].strip("\r\n"),)
 
         return fcell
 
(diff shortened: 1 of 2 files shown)


gabyx:
The state of the art coldfolding produces wrong results:
**Input**
bildschirmfoto 2017-08-21 um 23 39 10
**Output**:
bildschirmfoto 2017-08-21 um 23 39 15

juhasch:
Thanks. I will take a look at it.

src/jupyter_contrib_nbextensions/nbconvert_support/pre_codefolding.py
```diff
@@ -32,26 +32,52 @@ def fold_cell(self, cell, folded):
...
+        isSkipLine  = False
+        skipped = ""
+
         for i, l in enumerate(lines):
             # fold indent level
-            indent = len(l) - len(l.lstrip(' '))
-            if indent <= fold_indent:
+            lstrip    = l.lstrip(r' \t') # strip tabs and spaces
+            indent    = len(l) - len(lstrip)
+            isSkipLine = lstrip[0] == "#" or lstrip in ["\n"] # is it a comment or an empty line
```
jcb91: the `lstrip[0]` in this line is causing tests to fail. It should be something more like

```python
# skip comments and empty lines, except at the end of the fold
# in which case they'll be replaced when the fold finishes
is_skip_line = lstrip.startswith('#') or lstrip == '\n'
```

src/jupyter_contrib_nbextensions/nbconvert_support/pre_codefolding.py
```diff
@@ -32,26 +32,52 @@ def fold_cell(self, cell, folded):
         """
         Remove folded lines and add a '<->' at the parent line
         """
+        # self.log.debug("CodeFoldingPreprocessor:: folding at: %s" % folded)
```
jcb91: I think debug logs can (should) stay in, since the default log level is info...

src/jupyter_contrib_nbextensions/nbconvert_support/pre_codefolding.py
```diff
@@ -32,26 +32,52 @@ def fold_cell(self, cell, folded):
...
         lines = cell.splitlines(True)
 
         if folded[0] == 0 and (lines[0][0] == '#' or lines[0][0] == '%'):
             # fold whole cell when first line is a comment or magic
+            # self.log.debug("fold whole cell")
```
jcb91: ditto above, we can keep this at debug level

src/jupyter_contrib_nbextensions/nbconvert_support/pre_codefolding.py
gabyx: sure :-), will take a look at the mistakes =)

gabyx: by the way, how do you reproduce the linter errors locally?
I am using atom, but flake8 is giving me not the same errors? Do you have a solution for this or you just look at travis?

jcb91:
So the output on Travis comes from flake8, run by tox. I just run flake8 locally in a terminal, and that's enough, but I guess it's possible that the atom version isn't looking for config in the same places as my standalone setup?  Not sure at the moment sorry...

gabyx:
Ah now I got the errors with atom, was on the wrong branch shit :-)

Ehm -> Can line ending be '\r\n'  like on Windows? in python notebooks?
Hm... so we should probably make that robust for this too

gabyx:
Hm, it seems its alway '\n'  which is good

jcb91:
> Can line ending be '\r\n' like on Windows? in python notebooks?

I think notebook cells are stored as an array of lines, which makes them ending-agnostic, although the file itself can presumably have carriage returns... always best to assume the most difficult :wink:

gabyx:
I checked on a notebook from windows and it seems its always '\n'. Having a json file which is not lineendings-agnostic would be a design mistake =), so notebook went the right path here I think :)

gabyx:
Could we merge these changes =)? Thanks :-)

jcb91:
Staring at https://travis-ci.org/ipython-contrib/jupyter_contrib_nbextensions/jobs/267700939#L762, there are still a bunch of linting errors?

jcb91:
*starting

gabyx:
ah ok :), they are only warning, thx that you are so persistent :)! makes the code better

Von meinem iPhone gesendet

jcb91:
Ok, seems good to me, thanks @gabyx!

PR #1055, opened 2017-08-20
Update yaml files for 5.x compatible nbextensions

jfbercher:
I have gone though the nbextensions this evening and marked those which seems working for me. I haven't tested everything but at least the basic functionality is present in these extensions. Most of the nbextensions work. 
I have a doubt concerning `split cells notebook`, `scrolldown` and haven't tested `skill syntax`.

### DIFF
--- a/src/jupyter_contrib_nbextensions/nbextensions/autosavetime/autosavetime.yaml
+++ b/src/jupyter_contrib_nbextensions/nbextensions/autosavetime/autosavetime.yaml
@@ -1,5 +1,5 @@
 Type: IPython Notebook Extension
-Compatibility: 3.x, 4.x
+Compatibility: 3.x, 4.x, 5.x
 Name: AutoSaveTime
 Main: main.js
 Icon: icon.png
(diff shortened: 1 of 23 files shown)


jcb91:
Thanks, good work @jfbercher.

Skill syntax is pretty much just a codemirror mode, which means it relies little on notebook & should be fine on 5.x. I think we can update it & wait for bug reports :wink:. What were your doubts about split cell & scrolldown?

jfbercher:
if I remember well, split cell was splitting horizontally the cell with code on the left and output on the right. This is no more the case: code + output are on the 50% left. You are the author @jcb91 ? 
For scolldown, it seems that it doesn't scroll at all @kukanya ?

jcb91:
> This is no more the case: code + output are on the 50% left. You are the author @jcb91 ?

No, I'm not the original author, and haven't used it, so I'm not sure, but from memory I think the idea is that a split cell _should_ be 50% left, so that the next cell can be 50% right, and the two end up side by side. Maybe I've got that confused though...

jfbercher:
No, you are right. It is what it does. So it works. 
I think that I have seen the other extension (split input/output) elsewhere else. I will mark skill and splitcell as 5.x compatible.

jcb91:
thanks for this @jfbercher

Issue #1047, opened 2017-08-15
[NbConvertApp] WARNING | Config option `codefolding` not recognized by `NbConvertApp`


lucainnocenti:
I'm trying to export a notebook containing some `code_folding` into html, following [the relevant section in the guide](http://jupyter-contrib-nbextensions.readthedocs.io/en/latest/exporting.html#retaining-codefolding).

However, when running `jupyter nbconvert --to html --NbConvertApp.codefolding=True mynotebook.ipynb` the `codefolding` option seems to not be recognized, as per the title (I get 4 identical warnings about this), and the output html does not contain any folding.

I haven't tried the other option mentioned in the guide, that is, to modify the `jupyter_nbconvert_config.py` file, mainly because that file does not already exist in my system, while the guide seems to assume that it does and already contains some other code.
I have the `json` version of the file though, with the following content:
```
{
  "version": 1,
  "Exporter": {
    "template_path": [
      ".",
      "/home/lk/anaconda3/lib/python3.6/site-packages/jupyter_contrib_nbextensions/templates"
    ],
    "preprocessors": [
      "jupyter_contrib_nbextensions.nbconvert_support.CodeFoldingPreprocessor",
      "jupyter_contrib_nbextensions.nbconvert_support.PyMarkdownPreprocessor"
    ]
  }
}
```
So it looks like the exporting of the code folding should be already enabled, but for some reason it doesn't work.

jcb91:
> `[NbConvertApp] WARNING | Config option 'codefolding' not recognized by 'NbConvertApp'

This happens because we're using a value that isn't provided by the `NbConvertApp` class. I guess really we should use a config value on the preprocessor class, so something like `CodeFoldingPreprocessor.codefolding`. I guess this is the source of the problem, since it seems that if the warning has been triggered, the value doesn't actually get set (see [traitlets/config/configurable.py#L162-L186](https://github.com/ipython/traitlets/blob/e9acdb51268ec3637d4054cfa1c261e3802e9455/traitlets/config/configurable.py#L162-L186)), which becomes a problem (see below).

> I haven't tried the other option mentioned in the guide, that is, to modify the `jupyter_nbconvert_config.py` file, mainly because that file does not already exist in my system, while the guide seems to assume that it does and already contains some other code.
I have the `json` version of the file though, with the following content:

[...]

PR #1058, opened 2017-08-22
Fix codefolding preprocessor configuration

juhasch:
Adresses #1047:

Adds configurable traitlets ` CodeFoldingPreprocessor.remove_folded_code`and 
`CodeFoldingPreprocessor.fold_mark`.

### DIFF
--- a/src/jupyter_contrib_nbextensions/nbconvert_support/pre_codefolding.py
+++ b/src/jupyter_contrib_nbextensions/nbconvert_support/pre_codefolding.py
@@ -59,6 +59,13 @@ class CodeFoldingPreprocessor(Preprocessor):
                 fcell += l
         return fcell
 
+    def preprocess(self, nb, resources):
+        """Skip preprocessor if not enabled"""
+        if self.remove_folded_code:
+            return super(CodeFoldingPreprocessor, self).preprocess(nb, resources)
+        else:
+            return nb, resources
+
     def preprocess_cell(self, cell, resources, index):
         """
         Read cell metadata and remove lines marked as `folded`.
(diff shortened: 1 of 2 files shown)


jfbercher:
Perhaps adding a `preprocess` method would prevent calling `preprocess_cell` for each cell. 

```
def preprocess(self, nb, resources):

        if self.remove_folded_code: 
            return super(CodeFoldingPreprocessor, self).preprocess(nb, resources)
        else:
            #print("Applying CodeFoldingPreprocessor: disabled")
            return nb, resources 
```

juhasch:
Thanks, I guess this doesn't hurt.

Maybe I should change the unicode symbol '↔' for folding back to an ascii one '<->'.
The pygment highlighter (rightly I guess) marks it as an arrow.

src/jupyter_contrib_nbextensions/nbconvert_support/pre_codefolding.py
```diff
@@ -69,10 +81,11 @@ def preprocess_cell(self, cell, resources, index):
         index : int
             Index of the cell being processed (see base.py)
         """
-        dofolding = self.config.NbConvertApp.get('codefolding', False) is True
-        if hasattr(cell, 'source') and cell.cell_type == 'code' and dofolding:
-            if hasattr(cell['metadata'], 'code_folding'):
-                folded = cell['metadata']['code_folding']
-                if len(folded) > 0:
-                    cell.source = self.fold_cell(cell.source, folded)
+        if self.remove_folded_code:
```
jcb91: is this check necessary, given the check in the `preprocess` method above?

juhasch: True.

PR #1061, opened 2017-08-23
[various] use Jupyter.notebook.config instance, where appropriate

jcb91:
to avoid reloading (from server!) the same config file for each nbextension.
In most cases this also means we don't need to require the 'services/config' module any longer.


jcb91:
This may break a few things if I've not understood the loading mechanisms for all affected nbextensions, but I'm going to merge now, then wait for bugs :P

PR #1066, opened 2017-08-25
[toc2] tentative fix for #1065 -- update jqueryui cdn and version

### LINKED ISSUE #1065: toc2 export to html issue
I have difficulties with "Save as html (with toc"): export to html itself works perfectly, but I can't see TOC afterwards in html file.  

In javascript  console I can see the following: 

toc2

[toc2] working in non-live notebook: Error: Module name "base/js/namespace" has not been loaded yet for context: _
http://requirejs.org/docs/errors.html#notloaded
    at C (require.min.js:8)
    at g (require.min.js:26)
    at toc2.js:15
    at Object.execCb (require.min.js:29)
    at $.check (require.min.js:18)
    at $.enable (require.min.js:23)
    at Object.enable (require.min.js:27)
    at $.<anonymous> (require.min.js:23)
    at require.min.js:8
    at v (require.min.js:7)

and then 

Uncaught TypeError: $(...).draggable is not a function
    at create_toc_div (toc2.js:284)
    at Object.table_of_contents (toc2.js:535)
    at Untitled22.html:14676
    at Object.execCb (require.min.js:29)
    at $.check (require.min.js:18)
    at $.enable (require.min.js:23)
    at $.init (require.min.js:17)
    at require.min.js:26

Mac OS 10.11.6, in both Chrome and Safari. Fresh installation of Anaconda
jupyter notebook --version
5.0.0

python --version
Python 2.7.13
Notebook extensions were installed with conda: 
conda install -c condo-forge jupyter_contrib_nbextensions

any help would be highly appreciated.

jfbercher:
Updating jquery ui CDN seems to fix #1065. 

I also updated stylesheet version and removed a redundant line (jquery itself is already loaded in the base html)

### DIFF
--- a/src/jupyter_contrib_nbextensions/templates/toc2.tpl
+++ b/src/jupyter_contrib_nbextensions/templates/toc2.tpl
@@ -4,10 +4,9 @@
 {%- block header -%}
 {{ super() }}
 
- <link rel="stylesheet" href="https://code.jquery.com/ui/1.11.4/themes/smoothness/jquery-ui.css">
+ <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/jqueryui/1.12.1/jquery-ui.css">
 
-<script type="text/javascript" src="https://ajax.googleapis.com/ajax/libs/jquery/1.9.1/jquery.min.js"></script>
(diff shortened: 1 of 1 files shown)


jcb91:
Looks good to me, makes sense for the issue we're seeing, though it might be good to also make TOC work even if jqueryui doesn't load correctly in future. Would that be possible? Not sure.

PR #1078, opened 2017-09-03
New release

jcb91:
Mainly prompted by new versions of jupyter_contrib_core, so the correct config key gets set on enabling.

Also move changelog into its own file, so it can be updated by each PR author :)


jcb91:
ugh apologies for the numerous rebases here 😟

Issue #1083, opened 2017-09-08
toc2 navigation issue


knowitallism:
I recently had to reinstall nbextensions. Since then, when I use toc2 it creates links of the form "href=\"http://localhost:8888/notebooks/..." This means that when I convert my notebook the resulting html file will link me back to the original ipynb file instead of the new html file. This happens even if I use the toc2 exporter to convert to html. I have no idea why toc2 is now creating links as an absolute file path instead of a relative one, but I would really appreciate any help anyone could give me in fixing this.

jfbercher:
ok. 
Can reproduce. This affects the `toc cell` but not the sidebar or window navigation. The bug was introduced in an attempt to fix #1023 which should have been fixed anyway by using `scrollIntoView` instead if following the click by the native browser method. Thus I think that we can return to relative links for the toc cell. 
Try to replace line 70 in `toc2.js` by 
```
            'href': h.find('.anchor-link').attr('href'),
```
if ok, I will push a fix later.

knowitallism:
Thanks, your fix seems to have fixed the problem perfectly.

jcb91:
@jfbercher, it seems you've pushed commit ecf4f3d fixing this directly to the repo master instead of making a PR? Although in this case the fix is simple enough that it's not a problem, I think in general it would be better if we could avoid pushing directly to the master branch on the main repo! I'm opening a PR with the relevant commit in :smile:

jfbercher:
@jcb91 Yep. I thought git will propose to make a PR of this commit but it didn't and applied it directly to master...

jcb91:
Ok, I'm going to close this as fixed :+1:

PR #1084, opened 2017-09-10
[toc2] relative links in "toc2 cell"

jcb91:
fix for #1083. Revert full url in links to relative to the current page.


jcb91:
Oh, also, it makes sense to add things to the changelog as part of the PR, so I'm doing that here :smile:

PR #1091, opened 2017-09-13
Fix wrong width when cfg.widenNotebook is disabled

fkjogu:
Additionally, if `cfg.widenNotebook` is enabled, it inserts dynamically
some margin if the TOC sidebar needs more space.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,11 +24,15 @@ which makes it much easier to fill in each release's changelog :)
 New features and bugfixes:
 
 - `toc2`
-  * [#1084](https://github.com/ipython-contrib/pulls/1084)
+  * [#1084](https://github.com/ipython-contrib/jupyter_contrib_nbextensions/pull/1084)
     [@jfbercher](https://github.com/jfbercher)
     fix for
     [#1083](https://github.com/ipython-contrib/issue/1083):
     Revert full url in links to relative to the current page.
+  * [#1091](https://github.com/ipython-contrib/jupyter_contrib_nbextensions/pull/1091)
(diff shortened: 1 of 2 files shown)


jcb91:
looks good to me once the log is altered. Could you also add a note to the changelog to describe the alteration please? Thanks :smile:

jfbercher:
See my comment in #871. 
The behaviour is not what I intended to do, but if people finds that useful, let's go.

jcb91:
> See my comment in #871.
> The behaviour is not what I intended to do, but if people finds that useful, let's go.

Ah, ok, I see what you mean. However, widening the notebook is potentially an unwanted side-effect. Offering an option to turn this off is a good compromise, but at the moment, the option appears from its description that it should turn this off completely, but actually in some circumstances it doesn't, which leads to extra confusion.

I think if we offer an option 'widen notebook', then it should apply whether the the sidebar is in place or not. If it's useful to have different behaviours depending on sidebar status, I think there should be a second option along the lines of 'widen notebook only when sidebar is open'. Thoughts?

fkjogu:
I've updated the commit. Sorry, this log was never meant to be inclued.

fkjogu:
I will rebase to current master.

fkjogu:
Another possible solution would be, to widen independent of `cfg.widenNotebook` if the sidebar would overlap the notebook. My PR just makes the (unwidened) notebook smaller.

fkjogu:
@jcb91 I think, it is fit for merging. Are there still concerns?

jcb91:
> Are there still concerns?

No, apologies I've just been a bit slow!

Issue #1092, opened 2017-09-13
Beautify the (parts) of the repository


fkjogu:
How do you think about beautifying (parts of) the code in this repository?

I was working on a pull request on `toc2`, specifically `toc2.js` and noted that the formatting was nonuniform. Code readability would gain tremendously if we would just run the code through some beautifers available to make the formatting more streamlined. One could use https://github.com/beautify-web/js-beautify for example for JavaScript.

Would you accept partial pull request tuning the formatting? Also making this part of a contributers policy would be great.

jcb91:
Haha yes, this has certainly crossed my mind, particularly when working on toc2 :wink:. I even started running things through jscs, but never got round to making a pr because the changes always end up conflicting with other PRs made in the interim.

Do we have a contributor policy @juhasch? I have a vague idea that maybe the organisation says we follow the ipython one, which if memory serves me correctly is something like "formatting PRs are noise" but I may be confused, or just plain wrong.

In brief, :+1: for me, though obviously we'd need to decide on what style we actually enforce...

juhasch:
I guess we should add a contributors policy, thank you for bringing this up.

I believe most projects have a policy of don't fix what ain't broken, i.e. if you work on something, cleaning up/beautifying is fine.  Running scripts automatically reformatting otherwise untouched code is not.

I think it is a wise policy to follow, especially considering that we have no tests for the extensions. So any breakage will only be detected by users later on.

So feel free to beautify the parts of toc2 you are working on, but not for the rest of the Javascript stuff, at least before we have proper tests.

jfbercher:
Yes, of course, you can submit a beautified version as a single PR. This can be useful. A difficulty is indeed as indicated by @jcb91 that "but never got round to making a pr because the changes always end up conflicting with other PRs made in the interim". 
For toc2, the extension has growth with many additions and has become a bit complex. Perhaps that a refactoring will be useful. I don't think that beautifying will help a lot, but is certainly welcome.

fkjogu:
Cool, I've created a PR. Can update it, when required. Nicer formatting does not replace proper refactoring, but at least, its not beautiful mess :smile:.

PR #1095, opened 2017-09-14
Beautify toc2.js using js-beautify

fkjogu:
This PR applies [js-beautify](https://github.com/beautify-web/js-beautify) to `toc2.js`, fixing heterogenous formatting.

I can updated this PR if there is some collision with another PR.

#1092


jfbercher:
Ok. Not so much changes after all :-))
Merging now before other PR comes in.

PR #1097, opened 2017-09-17
Make Travis happy again

juhasch:
* increase allowed line length (hard to achieve 80 characters with the long jupyter command lines)
* change imports so isort is happy


juhasch:
@jcb91 Do you agree with allowing longer line lengths ?

jcb91:
> @jcb91 Do you agree with allowing longer line lengths ?

Meh, I prefer 80 char, but yeah, it's not a big deal. I think I recall once seeing that IPython used 100-char lines, but I forget where I saw that (or even if I imagined it), so I'm happy to settle for 120.

Issue #1102, opened 2017-09-19
Jupyter 5.1 changes add_buttons_group


Sukneet:
Juypter commit https://github.com/jupyter/notebook/commit/af244a63b82c03d01783d29d0f821d1dcade8709 adds an html span with the label value when calling add_buttons_group. This causes a long name in the toolbar for a lot of extensions. For example, the freeze extension:

![image](https://user-images.githubusercontent.com/359625/30615813-5890d810-9d4d-11e7-981e-0c0f5c77f3cd.png)

jcb91:
D'oh!

Sukneet:
The following seems to work
https://github.com/Sukneet/jupyter_contrib_nbextensions/commit/1547b3e9a5e38329dbbc2b06f07ac6b2f5a964e2

Is the style ok?

jcb91:
Looks good to me, and the actions have the added bonus that they can be used in the command palette :)

jcb91:
Opened a PR #1103. There are quite a few other nbextensions that will need similar updates, though. Thanks @Sukneet for reporting & fixing freeze 😄 🙇

Sukneet:
I fixed the other ones last night. Just need to verify they are all working.

juhasch:
Thanks for working on this. I am updating n the `runtools` extension and fixed it there already.

Sukneet:
I think I got them all (except runtools)
https://github.com/Sukneet/jupyter_contrib_nbextensions/commit/b7aa1bd8160cd132c0e8ff877d5c0515b67a2df9

takluyver:
Sorry for this - I thought I had been more careful in making the change than I actually was. Thanks for working on adapting jupyter_contrib_nbextensions to deal with this change.

Once PR #1103 has cleared up the problematically long labels, I hope that some nbextensions will be able to make judicious use of shorter labels on the buttons to make their purpose clearer. E.g. this code in cite2c:

https://github.com/takluyver/cite2c/blob/f310fb00af829664c756404c343d948ab556a2e0/cite2c/nbext/main.js#L296-L302

Produces buttons like this:

![screenshot from 2017-10-05 11-30-15](https://user-images.githubusercontent.com/327925/31222954-c13ebf60-a9c0-11e7-9a41-08cbaf4964f5.png)

jcb91:
[...]

PR #1103, opened 2017-09-20
[various] update add_buttons_group to use action instead of button

jcb91:
As discussed in #1102

### DIFF
--- a/src/jupyter_contrib_nbextensions/nbextensions/addbefore/main.js
+++ b/src/jupyter_contrib_nbextensions/nbextensions/addbefore/main.js
@@ -9,24 +9,24 @@ define([
 
     var load_extension = function() {
              Jupyter.toolbar.add_buttons_group([
-                 {
-                      'label'   : 'Insert Cell Above',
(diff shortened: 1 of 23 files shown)


jcb91:
This looks mainly ok, but a few changes are needed:

 1. rebase onto current master, as there are now some merge conflicts
 2. accessing `Jupyter.actions` raises warnings, so it seems better to do it through `Jupyter.keyboard_manager.actions`.
 3. whenever we call `ActionHandler.register`, we should provide both an action name _and_ a prefix (the prefix should be the nbextension filename, similar to that used in logs for lots of examples)
4. various buttons were being created with `id`s, which the new formulation seems to have used as action names. This is ok, but they (the actions) need to have an `id` key set __as well__

jcb91:
Apologies, you're right, the action doesn't have an id, so instead we need to pass an object with keys `id` and `action` to the toolbar registration function 😲

Sukneet:
Sorry for not getting to this! Been pretty busy. I was going to do it tomorrow, but looks like you already did it. :)

PR #1107, opened 2017-09-23
Test/embeddingtests

juhasch:
This is just for testing the appveyor test failures.


jcb91:
I've something similar in my lxml branch...

jcb91:
It seems that part of the problem is that the lxml conversion converts the `&#39;` codes into regular apostrophe characters, and eats some whitespace (at least in 2.7)...

juhasch:
I can produce it locally now (wrong exporter was called before).
Stdout no embedding:
```
[NbConvertApp] Converting notebook notebook.ipynb to html
[NbConvertApp] Writing 249492 bytes to notebook.html
```
Stdout embedding:
```
[NbConvertApp] Converting notebook notebook.ipynb to html_embed
[NbConvertApp] try embedding url: D:\src\jupyter_contrib_nbextensions\tests\data\icon.png, format: png
[NbConvertApp] Writing 249544 bytes to notebook.html
```

In test_exporter:
```
statinfo_e.st_size = 249544
statinfo.st_size = 261295
```
Why is the not-embedded filesize larger than nbconvert reports it to be ?

juhasch:
Not embedded HTML file start:
```
<!DOCTYPE html>
<html>
<head><meta charset="utf-8" />
```

Embedded HTML file start:
```
<html>
<head><meta charset="utf-8">
```

OK, our test is not good enough.

jcb91:
> Why is the not-embedded filesize larger than nbconvert reports it to be ?

😕 not sure... maybe this is a timing issue?

juhasch:
The size difference is due to LF vs. CR+LF.

jcb91:
> The size difference is due to LF vs. CR+LF.

Oh man! That's still a problem 😓

jcb91:
Perhaps we ought to read the files and strip CRs then

juhasch:
We could parse nbconvert stdout. However, I don't understand why nbconvert gives back a different file size in the embedding case. Strange.

jcb91:
> However, I don't understand why nbconvert gives back a different file size in the embedding case. Strange.

Possibly lxml has already added line endings?

juhasch:
The thing is, that nbconvert reports the correct number. So it seems to not count the additional CR.
I guess the best way is to create a helper function that calculates the file size minus CR.

juhasch:
😄 Works now!

jcb91:
> The thing is, that nbconvert reports the correct number. So it seems to not count the additional CR.

maybe it counts each line separately?

It might be simpler (better?!) To do a `'\n'.join(output.splitlines())` in the exporter, rather than hack the test. It seems to me better to avoid creating the CAR in the first place, no?

juhasch:
Good point.

OK, I guess the right thing to do is call
`embedded_output = et.tostring(root, method="html", encoding='unicode')`
instead of 
`embedded_output = et.tostring(root, method="html")`

in the exporter.

gabyx:
I added: embedded_output = et.tostring(root.getroottree(), method="html", encoding='unicode')

PR #1117, opened 2017-10-06
[tests] test yaml files using jupyter_nbextensions_configurator

jcb91:
to avoid typos which may prevent nbextensions getting installed


jcb91:
this also updates us to configurator 0.2.8, which includes fixes for bugs in notebook 5.1.0

PR #1122, opened 2017-10-12
[python-markdown] avoid js error when marked doesn't wrap things in <p>

### LINKED ISSUE #1119: Python Markdown: Print lists
Thanks for this very useful extension!

I'm wondering if it's possible to make it work with lists as well, maybe automatically adding a new line for each element?

In code, I want to be able to:
```python
l = [1,2,3]
```
```markdown
# Markdown
Current list values: {{l}}
```
but also:
```markdown
# Markdown
Beginning of list: {{l[:2]}}
```

Would that be possible? At least currently, it doesn't seem to work (mac/conda-forge up-to-date).

jcb91:
Fixes #1119


juhasch:
Thanks :+1:

PR #1123, opened 2017-10-12
 [various] ensure actions have names and prefixes, and buttons keep correct ids

jcb91:
This is essentially an update on top of #1103, adding changes mentioned in https://github.com/ipython-contrib/jupyter_contrib_nbextensions/pull/1103#issuecomment-334240369


jcb91:
I'm going to go ahead and merge this, hoping that it doesn't break anything (on the basis that it adds no **extra** jshint errors :rofl:). As always, open issues as required if it turns out I _have_ broken something!

Issue #1088, opened 2017-09-12
[toc2] changed settings not taking effect


lzkelley:
When I change the settings for TOC2 they don't take effect, even after restarting the entire notebook.  For example, I turned off the skip-h1-level setting, but the h1 headers are still being ignored...  Not sure what other details would be helpful to provide.

Settings do not change even after disabling and re-enabling the extension.

I'm on jupyter 4.3.0

jfbercher:
I think that is due to the fact that the corresponding parameter is fixed from system configuration but saved in notebook's metadat and reread from notebook's metada if present. The point is that we can not (yet) change it per notebook as the "n" or "t" links. I mentioned that before. For now, you should just edit the notebook metadata from Edit/Edit Notebook metadata and delete the relevant key.  I do not have the time from now to think to a fix. Alternatively, you can use the `title <a class='tocSkip'>` trick in #1030.

lzkelley:
Thanks, that works.

1) `edit > edit notebook metadata`
2) manually changed the values I wanted (in this case `"skip_h1_title"` & `"toc_cell"`)
3) saved, reloaded browser window, and the changes took effect.

jcb91:
> It seems like this is an important point that should go on the Readme, if not on the overall configurator page (if this is something that applies to many extensions

Yeah, this should have been clarified, it's my fault really for not properly thinking through  the consequences of having a per-notebook setting which doesn't have a button on the TOC panel. I guess really, we should either make the setting global, or add a button. Any thoughts as to which would be the better approach welcome! I think this should be limited to this extension, and possibly just this setting, though.

> I'll leave it open since the underlying issue still exists.

Good plan, it's best to have it open until we actually fix it, particularly as it helps people find the temporary workaround :+1:

jcb91:
this should hopefully be fixed by #1124, if anyone gets a chance to try that PR, feedback could be very helpful :)

jcb91:
I'm going to close this as fixed by #1124, but as always, reopen as necessary if it's not fixed!

Issue #1070, opened 2017-08-31
How to enable skip h1 header config?


xxf1995:
I can't find this option in the parameters under nb extensions config.
![wx20170830-185414](https://user-images.githubusercontent.com/8936233/29902707-a4232628-8db4-11e7-9069-65643edc73f9.png)

jcb91:
It looks as though you don't have a new enough version installed. This is probably (at a guess, given no further information) because you're using conda, and I haven't put together a new conda release yet 😅

jcb91:
this should hopefully be fixed in the most-recent releases

PR #1124, opened 2017-10-13
[toc2] configurable titles; dialog-alterable metadata settings

### LINKED ISSUE #1098: EXT: TOC title customization
I use this extension a lot and it is great! I am trying to replace the "Table of Contents" text with a text that would match the language of the text. Having an English title while the rest of the document is not English does not look nice:)

jcb91:
* fixes #1098 by adding titles which can be customised
 * adds a dialog to edit settings in notebook metadata, which should hopefully avoid confusion as in ~#1070~, #1088

### DIFF
--- a/src/jupyter_contrib_nbextensions/nbextensions/toc2/README.md
+++ b/src/jupyter_contrib_nbextensions/nbextensions/toc2/README.md
@@ -4,10 +4,10 @@
 
 The toc2 extension enables to collect all running headers and display them in a floating window, as a sidebar or with a navigation menu. The extension is also draggable, resizable, collapsable, dockable and features automatic numerotation with unique links ids, and an optional toc cell. Sections of currently selected/edited or running cells are highlighted in the toc. Some minor diplay tweaks are also available (moving header tile/menus, widening cells); Finally, the toc can preserved when exporting to html.
 
-#### First demo: Floating toc window and SideBar, toc auto-update, section numbering
+### First demo: Floating toc window and SideBar, toc auto-update, section numbering
 ![](demo.gif)
 
-##### Second demo: Save as html with toc / Navigation menu
+### Second demo: Save as html with toc / Navigation menu
 ![](demo2.gif)
 
 ### Third demo: Notebook scrolling and Collapsing sections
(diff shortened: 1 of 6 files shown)


jfbercher:
Very nice and very useful; Thanks @jcb91 !

I just tried it. A couple of remarks:

-  It would be great if modifications to the toc cell title/sidebar title were taken into account immediately (for now it seems that one needs to save and reload, though the notebook metadata are correctly updated and even when reloading via the reload button)
- perhaps that these two parameters can be also added system wise in the yaml config file 
- I think the reload button is, for a moment, mostly useless, since toc is updated every time a md cell is rendered

jfbercher:
A small not really important issue. 
Begin with toc2 as a sidebar. Drag it as a floating window. Then collapse it. Drag it back to the sidebar. Then uncollapse it. It doesn't show up. It seems that the height is zero. Drag it back as a floating widow, and try to collape/uncollapse; the height seems still zero. One must resize the window to get the toc contents. 

Sorry I am really overloaded and don't really have time to look in the code;

jcb91:
Thanks for taking time to look @jfbercher,much appreciated :)

> It would be great if modifications to the toc cell title/sidebar title were taken into account immediately

Ok, good spot. I have more modifications locally which aren't in this pr (mainly for simplicity/clarity of each pr), so I thought this was already the case, but at any rate, it should be easy to make it so.

> perhaps that these two parameters can be also added system wise in the yaml config file

Good point, I'd forgotten this!

> I think the reload button is, for a moment, mostly useless

True, I just figured it was easier to keep than try to catch every edge case for when to update - would you rather remove it completely?

> perhaps that the collapse/uncollapse button should be desactivated/grayed when the toc window is transformed into the sidebar, and the toc automatically shown/uncollapsed

That's the obvious/intuitive behaviour for me, but I wasn't sure whether others were using it as both sidebar and collapsed, so kept it. I can easily adjust it to work this way though, if you think that's ok?

jcb91:
> Sorry I am really overloaded and don't really have time to look in the code

No problem, thanks for looking at all! Bonne chance with all the other stuff :)

jfbercher:
>>I think the reload button is, for a moment, mostly useless

>True, I just figured it was easier to keep than try to catch every edge case for when to update - would you rather remove it completely?

>> perhaps that the collapse/uncollapse button should be desactivated/grayed when the toc window is transformed into the sidebar, and the toc automatically shown/uncollapsed

I don't see any problem to remove it, but perhaps in another PR, indeed. 

> That's the obvious/intuitive behaviour for me, but I wasn't sure whether others were using it as both sidebar and collapsed, so kept it. I can easily adjust it to work this way though, if you think that's ok?

yes, I think it is something to do. 

>> Sorry I am really overloaded and don't really have time to look in the code

> No problem, thanks for looking at all! Bonne chance with all the other stuff :)

Thanks!

jfbercher:
Perhaps also update the README/history?

jcb91:
Ok, I think this should be working now. Main differences include:

 * The panel is now added to the `#site` div in live notebooks, as it allows much simpler logic for the sidebar - just setting top & bottom rather than height.
 * The minimise button is now hidden in sidebar mode
 * Simplified setting & saving of size & position in metadata
 * add `display: hidden` to `tocSkip` class

jfbercher:
Hi @jcb91 

Just downloaded and begun testing. Very nice! :+1:

jfbercher:
>  Very nice!

No. Great!

jcb91:
:) glad you like it!

jankoslavic:
This looks good; hope it will be merged to master soon! Good job!

jcb91:
Ok, let's merge this, and fix anything it's missed as it turns up :)

PR #1128, opened 2017-10-18
[various] bugfix toolbar buttons with id

jcb91:
the ability to specify an action and an id was only added in notebook 5.2.0 (in https://github.com/jupyter/notebook/commit/af244a63b82c03d01783d29d0f821d1dcade8709), so any release before that fails to create buttons using the `{id: 'btn-id', action: }` syntax. This pr fixes the relevant bugs introduced in #1123


juhasch:
👍 Thanks!

PR #1131, opened 2017-10-19
Improve hide input by keeping the cell toolbar visible

scholi:
Hi,
Here is the pull request from the suggestion to hide only the input_area while keeping the cell toolbar visible while hiding inputs with the "Hide_input" extension.

Cheers and Enjoy!

Olivier

### DIFF
--- a/src/jupyter_contrib_nbextensions/nbextensions/hide_input/main.js
+++ b/src/jupyter_contrib_nbextensions/nbextensions/hide_input/main.js
@@ -15,14 +15,14 @@ define([
         // Find the selected cell
         var cell = Jupyter.notebook.get_selected_cell();
         // Toggle visibility of the input div
-        cell.element.find("div.input").toggle('slow');
+        cell.element.find("div.input_area").toggle('slow');
         cell.metadata.hide_input = ! cell.metadata.hide_input;
     };
 
     var update_input_visibility = function () {
         Jupyter.notebook.get_cells().forEach(function(cell) {
             if (cell.metadata.hide_input) {
-                cell.element.find("div.input").hide();
+                cell.element.find("div.input_area").hide();
             }
         })
     };


jcb91:
looks good to me :+1: 
perhaps it would be worth making a similar change to `hide_input_all`?

scholi:
Ok, done

scholi:
Wait before merging. I think that it's buggy with the mark-down cells for the hide_input_all extensions...

scholi:
Ok. I have tested it and seems to work fine on my side.

jcb91:
>  I think that it's buggy with the mark-down cells for the hide_input_all extensions

The same problem sort-of applies for `hide_input` - if someone runs this on a markdown cell, it'll show/hide the input area.

It occurs to me that the new implementation leaves the input prompts visible, which may be undesirable for many users. Perhaps we ought to be acting on something like

```javascript
$('.input .input_area, .input .input_prompt')
```

and finally, it may also be that some people don't want the cell toolbars to show on cells with hidden inputs, so perhaps we should make the whole thing optional :thinking:

scholi:
Ok, now it's a bit more mature. I have added a checkbox in the view menu which change the metadata of the notebook "hide_cellprompt". When set to true, the cell_toolbar will be hidden otherwise it will stay visible when toggling the visibility of the input.

jcb91:
Thanks @scholi for the update, apologies for the delay looking at it. Seems like a neat solution, though it might be nice in future to also be able to set a default value in notebook config. You're right, the menu wording is a bit unwieldy, but I can't really think of a more succinct way to express the concept -
the best I could think of was "show/hide celltoobar for hidden inputs" - it basically needs explaining 😂 btw there's a couple of typos in titles, where `hidding` should be `hiding`.

scholi:
Ok. Done. I have changed the checkbox in an eye which is more intuitive. I have also improved the wording as you suggested and now the script will look for the config value if not set for the notebook. The config is in hide_input.json as follow:

```
{
  "hide_input": {
    "hide_cellprompt": true
  }
}
```

jcb91:
Thanks @scholi for addressing my earlier comments :)

> The config is in `hide_input.json`

Apologies, I should have explained better. If you put this into a `hide_input` key of `notebook.json`, then you can use the `Jupyter.notebook.config` instance, and also the configurator for setting the default - see e.g. [nbextensions/notify/notify.js#L192-L193](https://github.com/ipython-contrib/jupyter_contrib_nbextensions/blob/8f2c1d24ac297e3a1a7c129a9b989390911841a4/src/jupyter_contrib_nbextensions/nbextensions/notify/notify.js#L192-L193) and [nbextensions/notify/notify.yaml#L10-L18](https://github.com/ipython-contrib/jupyter_contrib_nbextensions/blob/8f2c1d24ac297e3a1a7c129a9b989390911841a4/src/jupyter_contrib_nbextensions/nbextensions/notify/notify.yaml#L10-L18) for how to do this kind of thing.

Anyway, I think we can almost merge as-is for now, it can always be updated later. The one thing I'd like addressed before merging is that the default value should be to hide the toolbar, as it was before (this seems the more common use case, and it's usually better not to alter defaults without good reason!)

scholi:
Hi @jcb91,
Sorry for the late reply. Yes I had trouble to find the non-existing doc of the javascript API of jupyter so I try to figure things out by myself. So it works, but it's probably not the cleanest way of doing it.
I have now set the default value to hide the celltoolbar. I have also tried to use the `Jupyter.notebook.config` with the yaml file, but somehow I did not succeed. Somehow for me `Jupyter.notebook.config.data`only returns  a dictionnary of `load_extensions` and nothing more. For this reason I have kept the  json config file hide_input.json as a way to store the default value. If none is found in this file the value true will be taken and the celltoolbar will be hidden by default.

src/jupyter_contrib_nbextensions/nbextensions/hide_input/main.js
```diff
@@ -3,32 +3,85 @@
...
         // Find the selected cell
         var cell = Jupyter.notebook.get_selected_cell();
         // Toggle visibility of the input div
-        cell.element.find("div.input").toggle('slow');
-        cell.metadata.hide_input = ! cell.metadata.hide_input;
+        if(Jupyter.notebook.metadata.hide_cellprompt){
+                cell.element.find("div.input").toggle('slow')
+        }else{
+            cell.element.find("div.input > div.inner_cell > div.input_area").toggle('slow')
+            cell.metadata.hide_input = ! cell.metadata.hide_input;
```
jcb91: I think this should be done regardless of the value of `Jupyter.notebook.metadata.hide_cellprompt`, no?

src/jupyter_contrib_nbextensions/nbextensions/hide_input/main.js
```diff
@@ -40,12 +93,18 @@ define([
                 }
             }, 'toggle-cell-input-display', 'hide_input')
         ])).find('.btn').attr('id', 'btn-hide-input');
-        // Collapse all cells that are marked as hidden
-        if (Jupyter.notebook !== undefined && Jupyter.notebook._fully_loaded) {
-            // notebook already loaded. Update directly
-            update_input_visibility();
-        }
-        events.on("notebook_loaded.Notebook", update_input_visibility);
```
jcb91: This is still needed. Consider the case in which the notebook finishes loading *before* the config loads, but `Jupyter.notebook.metadata.hide_cellprompt !== undefined`. In this case, the input visibility is updated immediately, but won;t be updated again one the notebook finishes loading...

jcb91:
No worries :) Looks mainly good, with a couple of notes I've added in review.

> find the non-existing doc of the javascript API of jupyter

:laughing: yeah, that can't have helped! They don't exactly consider it an API, in that at least parts of it change with little warning. I just read the source by now! You seem to have figured it out pretty well though :)

>  I have also tried to use the `Jupyter.notebook.config` with the yaml file, but somehow I did not succeed. Somehow for me `Jupyter.notebook.config.data` only returns a dictionary of `load_extensions` and nothing more.

That (only having `load_extensions`) is to be expected, until you (or some script etc) sets some value(s). The `Jupyter.notebook.config` object represents the `notebook.json` config file on disk (the `notebook` 'section' in terms of the `ConfigSection` constructor). The `load_extensions` dict is how the notebook decides which nbextensions to load - any with value `true`. This is the config file that gets edited by `jupyter nbextension enable <require-uri>` commands. It can also be used to store other values in under other keys, however. It's more convenient than a sepearate config file, because the notebook already loads that config file anyway. The purpose of the yaml file and configurator is essentially just to provide a way to edit that json config file without having to do it directly.

It's not important to use this method in this particular PR, but in case it's useful in future, the general model for config loading that I use for nbextensions is:

 1. I expect options to be stored in the `notebook` config section (allowing the use of `Jupyter.notebook.config`), in an object named after the nbextension. 
 2. I store an options object containing default config values in the top-level closure of the nbextension ([notify/notify.js#L18-L21](https://github.com/ipython-contrib/jupyter_contrib_nbextensions/blob/8f2c1d24ac297e3a1a7c129a9b989390911841a4/src/jupyter_contrib_nbextensions/nbextensions/notify/notify.js#L18-L21))
 2. Define what options are available in the yaml file, so that the end-user can use the [jupyter_nbextensions_configurator](https://github.com/Jupyter-contrib/jupyter_nbextensions_configurator) to set the appropriate values in the server's `notebook.json` file. See [here for info on the yaml files](https://github.com/Jupyter-contrib/jupyter_nbextensions_configurator#yaml-file-format).
 3. Update the default-values object with any that might be found in `Jupyter.notebook.config.data` using [jquery's extend method](https://api.jquery.com/jquery.extend/). The value pulled from there may be `undefined` if no options have been explicitly set in the config file, or might only contain some of the keys, but jquery extend copes with that ok - see the [jquery docs](https://api.jquery.com/jquery.extend/) for details.
 4. erm... that's usually it. For a few extensions which store some values in _notebook_ metadata (for per-notebook storage as opposed to per-user in server config), the values are also updated from notebook metadata, but this gets more complicated as then you need a neat way to set the per-notebook data also, which the configurator doesn't provide.

Hope that makes sense...

src/jupyter_contrib_nbextensions/nbextensions/hide_input/main.js
scholi: The div.input contains the header and with that the celltoolbar. So line 23 should be only performed if _hide_cellprompt_ is true. What I should probably change is to force to show _div.input_ when _hide_cellprompt_ is false and to show _div.input_prompt_ as well as _div_input_area_

src/jupyter_contrib_nbextensions/nbextensions/hide_input/main.js
scholi: Agree

scholi:
Hi @jcb91,
Thank you for your feedback and sorry for the late reply.
I'll dig into the details of jupyter.notebook.config soon. I like the idea to have the default value in the configurator instead as in the json. What is not so clear to me yet is why the default values are set in the yaml file [notify/notify.yaml#L14](https://github.com/ipython-contrib/jupyter_contrib_nbextensions/blob/8f2c1d24ac297e3a1a7c129a9b989390911841a4/src/jupyter_contrib_nbextensions/nbextensions/notify/notify.yaml#L14) as well in the js file [notify/notify.js#L18-L21](https://github.com/ipython-contrib/jupyter_contrib_nbextensions/blob/8f2c1d24ac297e3a1a7c129a9b989390911841a4/src/jupyter_contrib_nbextensions/nbextensions/notify/notify.js#L18-L21) ?

jcb91:
> sorry for the late reply.

No need to apologise at all! As with everyone here, this isn't your job, so we completely understand that other stuff gets priority! You've taken time and effort to contribute, for which I (and doubtless many others) are grateful :)

> What is not so clear to me yet is why the default values are set in the yaml file [notify/notify.yaml#L14](https://github.com/ipython-contrib/jupyter_contrib_nbextensions/blob/8f2c1d24ac297e3a1a7c129a9b989390911841a4/src/jupyter_contrib_nbextensions/nbextensions/notify/notify.yaml#L14) as well in the js file [notify/notify.js#L18-L21](https://github.com/ipython-contrib/jupyter_contrib_nbextensions/blob/8f2c1d24ac297e3a1a7c129a9b989390911841a4/src/jupyter_contrib_nbextensions/nbextensions/notify/notify.js#L18-L21) ?

True, the reason for that's not obvious. Essentially we started using yaml files to declare stuff to the configurator because yaml's a nice format for people to write. However, there's no good javascript libraries for loading yaml, so they're not a suitable format for the nbextensions themselves to read. As a a result, __we store the _actual_ defaults that the nbextension _uses_ in the js file__. But, for the configurator to correctly display what the defaults are, we _also_ need them in the yaml file. __Those in the yaml file are _only_ used by the configurator, to try to convey to the end-user what the default is__. Essentially, they're used to populate the controls on the configurator page for editing the settings. They're never actually used by the extension, or written into config, so they're not even necessarily the same. Usually, they should be the same, although in fact it can _occasionally_ be more helpful to the user to display what is _effectively_ the default value, rather than the actual default as it is stored.

It's far from an ideal system, I admit - I would rather use a json file that both the configurator and the nbextension itself could load, but for backwards-compatibility, we're still using separate files.

I hope that makes (some!) sense?

scholi:
Yes that make sens. Thank you for the detailed explanations.

Issue #1134, opened 2017-10-20
DEP: traitlets 4.1 metadata should be set using the .tag


jakirkham:
Ran into this warning recently when starting up a notebook server. Included environment details below.

<details>
<summary>Error:</summary>

```python
/opt/conda/lib/python3.6/site-packages/jupyter_contrib_nbextensions/nbconvert_support/nbTranslate.py:214: DeprecationWarning: metadata {'shortname': 'rh'} was set from the constructor. With traitlets 4.1, metadata should be set using the .tag() method, e.g., Int().tag(key1='value1', key2='value2')
  help="Selected language").tag(config=True)
```

</details>

<details>
<summary>Environment:</summary>

```yaml
name: root
channels:
- conda-forge
- defaults
dependencies:
- anaconda-client=1.6.5=py_0
- apr=1.5.2=1
- asn1crypto=0.22.0=py36_0
- backports=1.0=py36_1
- backports.functools_lru_cache=1.4=py36_1
- beautifulsoup4=4.6.0=py36_0
- blas=1.1=openblas
- bleach=2.0.0=py36_0
- bokeh=0.12.9=py36_0
- boost=1.64.0=py36_4
- boost-cpp=1.64.0=1
- bottleneck=1.2.1=py36_1
- bzip2=1.0.6=1
- ca-certificates=2017.7.27.1=0
- certifi=2017.7.27.1=py36_0
- cffi=1.10.0=py36_0
- chardet=3.0.4=py36_0
- click=6.7=py36_0
- cloudpickle=0.4.0=py36_0
- clyent=1.2.2=py36_0
- conda=4.3.29=py36_0
- conda-build=2.1.17=py36_0
- conda-env=2.6.0=0
- conda-verify=2.0.0=py36_0
- cryptography=2.0.3=py36_0
- curl=7.54.1=0
- cycler=0.10.0=py36_0
- dask=0.15.4=py_0
- dask-core=0.15.4=py_0
- dask-imread=0.1.1=py36_0
- dask-ndfilters=0.1.2=py36_0
- dask-ndfourier=0.1.2=py36_0
- dbus=1.10.22=0
- decorator=4.1.2=py36_0
- distributed=1.19.3=py36_0
- drmaa=0.7.7=py36_0
- entrypoints=0.2.3=py36_1
- expat=2.2.1=0
- fasteners=0.14.1=py36_2
- fftw=3.3.6=3
- filelock=2.0.6=py36_0
[...]

PR #1138, opened 2017-10-23
[nbTranslate] use traitlets > 4.1 tag API

jcb91:
also use the 'alias' metadata tag over the deprecated 'shortname'
prompted by #1134


jcb91:
fixes #1134

PR #1155, opened 2017-11-04
[livemdpreview] add new nbextension!

jcb91:
idea prompted from https://github.com/jupyter/notebook/issues/2882

### DIFF
--- a/src/jupyter_contrib_nbextensions/nbextensions/livemdpreview/livemdpreview.js
+++ b/src/jupyter_contrib_nbextensions/nbextensions/livemdpreview/livemdpreview.js
@@ -35,7 +35,7 @@ define([
     };
 
     LiveMdPreviewer.prototype._default_options = {
-        show_side_by_side: true,
+        show_side_by_side: false,
         timeout :  500,
     };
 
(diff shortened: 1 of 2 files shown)


juhasch:
Very nice :smile:

One minor thing: The default for `show_side_by_side` is different in the yaml and in the js file.
I guess we should *really* try to read in the default values from the description file and not have it set twice.

jcb91:
Ah, good spot @juhasch, thanks! 

> I guess we should really try to read in the default values from the description file and not have it set twice.

Haha yeah, it could be simpler! I think we'd need to be using JSON though :wink:

juhasch:
If it works I will even swallow evil JSON :smirk:

jcb91:
Haha, well, getting the defaults from json should be easy enough, something like

```js
var utils = require('base/js/utils');
var pkg_json_url = require.toUrl('./package.json');
var params = {};

utils.promising_ajax(
    pkg_json_url, {processData : false, cache : false, type : "GET", dataType : "json"}
).then(function (pkg_json) {
    // e.g. if we use the config suggested in https://github.com/jupyter/notebook/issues/2824#issuecomment-334124949
    $.extend(true,
        params,
        pkg_json.org_jupyter_nbextension,
        Jupyter.notebook.config.data.mynbextensionname);
});
```

jcb91:
hmm, actually, I guess it's a bit more difficult, since there's also the stored parameter metadata. So you'd need something like

```js
var default_params = {};
pkg_json.org_jupyter_nbextension.parameters.forEach(
    function (param) { default_params[param.name] = param.default; });
```

to handle that

Issue #1034, opened 2017-07-28
toc2 sidebar navigation not working in Firefox or IE


aaros-web:
I'm using nbconvert to convert a .ipynb to .html with TOC nb extension and toc2 template but the resulting html file seems broken for the sidebar navigation part:
* with Firefox, the sidebar is present but clicking on a link doesn't work
* with IE, the sidebar is not even present
* with Chrome, everything seems fine

I'm thinking about sharing such html for analysis reports and it would be nice if the result was browser-agnostic.

If useful, my configuration is:
* jupyter 4.3.0
* jupyter nbconvert 5.1.1
* jupyter nbextension 5.0.0

jcb91:
Hmm, strange. Thanks for the report, could you post your nbconvert command line/script call, and anything you see in the 
[javascript console](http://webmasters.stackexchange.com/questions/8525/how-to-open-the-javascript-console-in-different-browsers) in IE/firefox?

chrisjsewell:
@jcb91  this may/may not be related but I have just noticed in your recent commit you change:

```javascript
a.attr("href", '#' + h.attr('id'));
```

to

```javascript
a.attr("href", window.location.origin + window.location.pathname + '#' + h.attr('id'));
```

For a local file, in Safari it works fine, but in Firefox the toc links now have a duplication, i.e.

`file:///path/to/folder/null/path/to/folder/filename.html#header-tag-1`

Was there a reason that you made this change, since it seemed to work fine before it?

Ta,
Chris

jcb91:
Hmm, that sounds like it's the cause. Essentially there's no necessary reason for the change - the first way was my solution, the second was @jfbercher's, and it looks like there's been some confusion in merging the two. Thanks for the diagnostic work!

jcb91:
Apologies, I got confused. The change was to fix #1023. My solution for that hasn't been merged yet, but will be in one of the TOC PRs I guess...

jfbercher:
I do not see why this don't work; perhaps @jcb91 's solution would. But perhaps this does not really matter since it seems that #1023 will not be solved anyway. As suggested by @jcb91, an issue should be opened at htmlpreview.github.io  
My proposal would be revert to the previous situation. Your opinion @jcb91? 
This would be probably the case when merging the current PR..

[In addition there was an additional issue in highlighting of selected cells that was introduced and that will be solved that way]

marianneke:
[...]

PR #1167, opened 2017-12-04
[toc2] move config into toc2.js

jcb91:
place default config & config loading into `toc2.js` (where they're used) to prevent duplicating stuff in both `main.js` and `toc2.tpl`, and avoid issues like that in https://github.com/ipython-contrib/jupyter_contrib_nbextensions/issues/1034#issuecomment-347188877


jcb91:
hopefully, this fixes #1034

jfbercher:
Good work @jcb91 
Thanks for fixing that.

Issue #1172, opened 2017-12-14
Does this package actually require a specific lxml version?


codypiersall:
In the setup.py, the version of lxml is tied to `>=3.8.0`.  I need to run jupyter_contrib_nbextensions on an embedded ARM computer where I built a kernel using [Yocto](https://yoctoproject.org/) and the version of Yocto I'm using only supplies lxml 3.6.4.  It's not trivial to build lxml 3.8+ on my PC.

Does this package really need lxml 3.8 or greater?  I cloned this repo and installed it after removing the version requirement, and everything installed correctly and all the tests passed.

I'm happy to submit a PR, but before I did I wanted to see if I was missing something.  If nbextensions really needs lxml >= 3.8, I'll just pin down the version of nbextensions in my own setup.py, but it'd be nice not to need to.

jcb91:
I must admit that I don't really know! The only thing in the repo using lxml is the embedhtml exporter, but from a cursory glance at the usage and changelog, I'm unsure which would be the minimal required version. I suspect https://github.com/lxml/lxml/pull/233 (added for 3.8) might be relevant, but as mentioned, I'm unsure. Perhaps @gabyx has a better idea?

jcb91:
Also curious @codypiersall, what's the usage case here, if that's something you're able to divulge? :)

codypiersall:
Thanks for the quick responses @jcb91!

From the PR you linked to in lxml it looks like the `fromstring()` method got modified to be able to accept `bytes` _or_ `str`, and it "does the right thing."  The only usage I found of lxml in nbextension is [when converting the notebook to html](https://github.com/ipython-contrib/jupyter_contrib_nbextensions/blob/master/src/jupyter_contrib_nbextensions/nbconvert_support/embedhtml.py#L82).  I'm not sure what the type of `output` is at this line.  Probably a `str` at this point?  I'm on a different computer so I can't check at this point.  If it is a `str`, then the `lxml` version being less than 3.8.  If it's a `str` or `bytes`, we could do the check before calling `et.fromstring()`.

After we get some more feedback, I'd be happy to submit a PR that makes the version dependency on lxml go away one way or another.

[...]

PR #1174, opened 2017-12-15
Remove lxml >=3.8.0 dependency.

codypiersall:
The version dependency makes it hard to install nbextensions on certain
platforms (e.g. ARM).  See Issue #1172.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -72,7 +72,7 @@ if you encounter any problems, and create a new issue if needed!
             'pyyaml',
             'tornado',
             'traitlets >=4.1',
-            'lxml >=3.8.0'
+            'lxml'
         ],
         extras_require={
             'test': [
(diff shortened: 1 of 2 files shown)


codypiersall:
PR attempting to fix #1174.

juhasch:
Travis is happy, so this looks good to me.

On a side note: I am using berryconda on Arm, so I did not run into this problem. It provides all nice scientific Python packages in a recent version. You get lxml version 4.1.1 there.

jcb91:
Well, Travis is happy because it installs lxml 4.1.1, so this PR doesn't change that :laughing:. However, I did test lxml down to 3.4 at least with tests running ok. I thought maybe we should keep major-version dependency for compatibility reasons (so `lxml >= 3`), but tbh lxml major versions don't appear to indicate backwards-incompatibility, so whatever, I think this is fine. I've added a small alteration to limit lxml imports to when it's actually used, which should hopefully help limit some of the problems people have had with incompatible depeny libs for lxml.

codypiersall:
@jcb91 Thanks for doing the extra work of installing older versions and checking that everything still works!

PR #1175, opened 2017-12-17
Undefined title_sidebar

jfbercher:
In some "old" exported html files, with toc2 embedded, the `cfg.title_sidebar` is not defined; This tests for the cfg key and define it if necessary (Otherwise the sidebar title appears as "undefined")

### DIFF
--- a/src/jupyter_contrib_nbextensions/nbextensions/toc2/toc2.js
+++ b/src/jupyter_contrib_nbextensions/nbextensions/toc2/toc2.js
@@ -529,6 +529,12 @@
             return
         }
 
+        // In a live notebook, read_config will have been called already, but
+        // in non-live notebooks, ensure that all config values are defined.
+        if (!liveNotebook) {
+            cfg = $.extend(true, {}, default_cfg, cfg);
+        }
+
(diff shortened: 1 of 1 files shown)


jcb91:
This (along with all metadata values) should already have a default value in the newest version? Is that not working correctly?

jfbercher:
Yes,  with current version and conversions, but already converted notebooks do not have all variables defined. See eg [this example](https://rawgit.com/jfbercher/jupyter_latex_envs/master/src/latex_envs/static/doc/latex_env_doc.html). 

Of course I can easily update my own files (I have a bunch of lectures in ipynb/html formats), but there can be some other floating in the web.

jcb91:
Oh, I'm all for having defaults! I just thought that this (and similar issues) would be handled by 197231361975370a51b5b7a078c9b9ea0ef0a496, but it seems I forgot that the `read_config` method which defines the defaults isn't called in non-live notebooks :disappointed:.

I've pushed a commit which should ensure defaults are set for *all* config values...

jcb91:
Seems to work for the title :)

jfbercher:
Well done!

Issue #1176, opened 2017-12-18
toc2 unable to save as HTML (with toc)


cqcn1991:
When I click this link, nothing happens. I want to know, what went wrong?

![default](https://user-images.githubusercontent.com/2425141/34093771-19ba9992-e405-11e7-86b0-5c4c2acaed12.png)

jfbercher:
You should have a log similar to 
`
[toc2] running: !jupyter nbconvert 'Visu_n.ipynb' --template toc2
`
in your console and find the converted file in the same directory as the notebook. 
Just checked and this still works. 

This "Save as HTML (with toc)" could/should be rewritten to take advantage of the new api entry which enable to call nbconvert with a template as an option of the "download as" menu (for Jupyter v. 5.1). I may have the time to do it shortly.

cqcn1991:
Got it, so it seems I can also run `jupyter nbconvert 'Visu_n.ipynb' --template toc2` manually through cmd?

jfbercher:
Yes indeed. 
Change the filename of course! (quotes only needed if there are spaces in the filename)

jfbercher:
From the command line, you can also issue 
```
jupyter nbconvert --to html_toc FILE.ipynb
```

jfbercher:
Explanations provided. 
Furthermore a new possibility "Download as HTML with toc" was added for Jupyter >= 5.1, cf #1182  Closing this issue.

PR #1182, opened 2017-12-20
[toc2] Add a 'download as HTML with toc' for Jupyter >= 5.1

jfbercher:
As mentioned in #1176,  added a menu item in `File/Download as` to download the converted HTML file with toc2 embedded in it. This works for Jupyter version greater than 5.1 and any kernel. 

For earlier versions, we still have/keep the "save as with toc" which is system call to `jupyter nbconvert --template toc2 CurrentNotebookFile.ipynb`, but this approach needs a python kernel.


jcb91:
looks good to me!

qingpeng:
since I am running jupyter on server, honestly I prefer the  "save as with toc" option, since that will save the html to the directory on the server rather than download the file to my local laptop...

PR #1190, opened 2018-01-06
Fix typo and revise text for better readability.

adyavanapalli:

PR #1191, opened 2018-01-06
Typos + readability improvements.

adyavanapalli:


adyavanapalli:
Some associated changes: #1190.

jcb91:
Thanks again @adyavanapalli. Reading this, is there any reason we don't install the python bits of nbtranslate by default @jfbercher? I think we _could_, as we do with other exporters?

Issue #1079, opened 2017-09-04
exercise2 extension cells showing up in some notebooks but not others


satra:
we have a set of notebooks with exercise2 extensions enabled. in a jupyter-hub setting or when running these through a docker container, these extension enabled cells show up in some of these notebooks but not others even though the metadata exists in the json file.

in fact in a git repo, i can use the rubberband + exercise to re-enable the view, but it doesn't make any change to the json file (since the metadata is already there). 

we are wondering if this intermittent behavior has been observed before.

dkirkby:
I am seeing the same behavior with exercise2, but I have noticed that it correlates with longer notebooks and that the "Show Solution" widgets reappear if I delete enough cells.

I haven't tried using exercise (as opposed to exercise2) but do you know if this behavior is unique to exercise2?

dkirkby:
@jcb91 Is this the right place to post exercise2 issues?

jcb91:
@satra apologies, I've only just noticed this issue! @dkirkby Sure, this is the right repo for exercise/2 issues. Both exercise and exercise2 use an almost identical codebase, the only real difference is in the controls they provide (checkbox vs toggle switch), so it's unlikely that the behaviour is unique to one of them. Essentially, I think that this issue stems form the fact that both extensions only set the initial visibility of exercise solutions once, when the nbextension loads. For small notebooks, this is after the notebook has loaded, so all is well. For larger notebooks, this can happen _before_ all cells in the notebook load, and any that load after the nbextension won't get their exercises displayed correctly. It should be easy to fix, so I'll try to patch this soon.

dkirkby:
Thanks for looking into this @jcb91!  I am using exercise2 for in-class activities in a course right now, so happy to test any patch.

jfbercher:
You are right @jcb91. I thought that the test for notebook_fully_loaded was already implemented but it is not indeed, so the observed behaviour! 
Indeed the fix is easy as we know. 
The [display existing solution](https://github.com/ipython-contrib/jupyter_contrib_nbextensions/blob/master/src/jupyter_contrib_nbextensions/nbextensions/exercise2/main.js#L162-L189) section should be transformed into a function and called at initialization after the "wait for notebook loaded" test. 

The source is quite old now, and looking at it, it could also undergo a small lifting.

Issue #1195, opened 2018-01-15
All extensions don't work with big notebooks. Load timeout issue


ODemidenko:
> The version of the notebook server is 5.0.0 and is running on:
> Python 3.6.2 |Anaconda custom (64-bit)| (default, Jul 20 2017, 12:30:02) [MSC v.1900 64 bit (AMD64)]
browser: Chrome

With small/new notebooks - all extensions work well.
With a big notebook - they don't start.
I've tried setTimeout and enabling/disabling Highlighter extension as was suggested in [issue 1009](https://github.com/ipython-contrib/jupyter_contrib_nbextensions/issues/1009)
But that helped only once (and I didn't get which advice actually helped). And now the problem continues.

In the JS console I see following error:

> main.min.js:12431 Failed to load extensions: Array(16) Error: Load timeout for modules: 
> custom/custom,nbextensions/nbextensions_configurator/config_menu/main,nbextensions/contrib_nbextensions_help_item/main,nbextensions/collapsible_headings/main,nbextensions/toc2/main,nbextensions/varInspector/main,nbextensions/snippets_menu/main,nbextensions/hide_input_all/main,nbextensions/highlighter/highlighter,nbextensions/highlight_selected_word/main,nbextensions/execute_time/ExecuteTime,nbextensions/skip-traceback/main,nbextensions/freeze/main,nbextensions/chrome-clipboard/main,nbextensions/move_selected_cells/main,nbextensions/init_cell/main
> http://requirejs.org/docs/errors.html#timeout
>     at makeError (require.js:165)
>     at checkLoaded (require.js:693)
>     at require.js:714

probably I used setTimeout option wrong? Am I right that it should be added to notebook metadata, like this:

> {
> "setTimeout": 60,
>   "hide_input": false,
>   "kernelspec": {
>     "name": "python3",
>     "display_name": "Python 3",
>     "language": "python"
>   },
>   "language_info": {...
?

jcb91:
> With small/new notebooks - all extensions work well.
> With a big notebook - they don't start.

Unfortunately, although this (timeout) can be caused by nbextensions which take a long time to load, it's also an issue with notebook itself. You can check https://github.com/jupyter/notebook/issues/2075 for details, but essentially as I understand it, what's happening is that

 1. notebook javascript fires off requests to load nbextensions
 2. notebook javascript starts loading all the notebook content (rendering markdown cells etc)
[...]

PR #1203, opened 2018-01-21
[Exercise2] Wait for notebook to be fully loaded before restoring solution states (hidden/dispayed)

jfbercher:
Tentative fix to @#1079 

The main modification is in the `load_ipython_extension()` function where we wait for the notobook to be fully loaded before trigering the (new) `display_existing_sols_at_startup()` function. 

Also I ran a code prettifying on all that. 

@satra, @dkirkby, can you download the new `main.js`, replace the old one in your nbextension/exercise2/ directory and test if it works for you?

### DIFF
--- a/src/jupyter_contrib_nbextensions/nbextensions/exercise2/main.js
+++ b/src/jupyter_contrib_nbextensions/nbextensions/exercise2/main.js
@@ -9,35 +9,35 @@
 // exercise2: built by @jfbercher from an earlier work by @junasch october 2015) - see readme.md
 
 
-    function show_solution() {
-        var cell=IPython.notebook.get_selected_cell();
-        var cell_index = IPython.notebook.get_selected_index();
-        var ncells = IPython.notebook.ncells();
-            cell.metadata.solution2 = "shown";
-            IPython.notebook.select_next();       
-            cell = IPython.notebook.get_selected_cell(); 
-            while (cell_index++<ncells & cell.metadata.solution2 !=undefined & cell.metadata.solution2_first !=true) {
-                cell.element.removeClass('hidden');
-                cell.metadata.solution2 = "shown";
-                IPython.notebook.select_next();       
-                cell = IPython.notebook.get_selected_cell();         
-                }
-        }
+function show_solution() {
+    var cell = IPython.notebook.get_selected_cell();
+    var cell_index = IPython.notebook.get_selected_index();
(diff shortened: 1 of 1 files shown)


dkirkby:
For the record, my conda env seems to have `exercise2/main.js` in two locations:
```
.../share/jupyter/nbextensions/exercise2/main.js
.../lib/python3.6/site-packages/jupyter_contrib_nbextensions/nbextensions/exercise2/main.js
```
so I will update both of these by hand and will let you know.

jcb91:
> my conda env seems to have `exercise2/main.js` in two locations

This is expected. The first is the one used by the notebook server, the second is the copy in the python package, from which the in-use one gets copied.

dkirkby:
I have been using the updated `main.js` for a few days now and haven't had any issues.  Old notebooks using the extension still work, and I have been able to create new ones also.  I can't really confirm that the issue is fixed since I don't know how to reproduce it, but it sounds like the changes address the problem.

satra:
@djarecka - when you get a chance can you check this out?

dkirkby:
@jcb91 Do you expect this will be a complete fix for this issue?  I got the impression from #1195 that the problem might be deeper.

jcb91:
> @jcb91 Do you expect this will be a complete fix for this issue? I got the impression from #1195 that the problem might be deeper.

Sort of. There are three possibilities:

1. the notebook is small enough that it loads completely before the nbextension loads. In this case, there's no problem, even without the updates
2. the notebook is large enough that the nbextension loads _before_ all the notebook has loaded, in which case any exercises which load _after_ the nbextension loads currently don't get their solutions hidden.
3. the notebook is _so_ large that no nbextensions get loaded at all, because requirejs assumes the calls to fetch them have timed out. This is the deeper issue addressed in #1195, which requires a fix/mitigation in `custom.js` (so that it gets executed before nbextension loading is attempted)

The updates discussed here and in #1206 should fix case two, but don't (can't!) help in case 3. Does that make sense?

jcb91:
Closing this, as it should hopefully be handled (assuming nbextension gets loaded at all, as discussed above) by #1206.

djarecka:
@jcb91 @jfbercher - thank you for your help. Unfortunately, I'm having problem with finding the notebooks that we had problems with loading last summer...
I will remember to report it if I see the problem again.

Issue #1202, opened 2018-01-20
Proposal Add isort extension


benjaminabel:
I created an [isort](https://github.com/timothycrosley/isort) extension https://github.com/benjaminabel/jupyter-isort, that I use daily. Currently it only sort imports in the current cell, but it could move all import in first code cell, and sort it. I based the implementation on  kenkoooo/jupyter-autopep8 it's my first extension, so any advice is welcome. Should I propose a PR?

jcb91:
Sure, PRs always welcome :)

benjaminabel:
Let's go:rocket:

jfbercher:
Hi, actually the kenkoooo/jupyter-autopep8 nbextension was an evolution of the original [code_prettify](https://github.com/jfbercher/code_prettify). 

Together with @jcb91 we have created a library that simplifies and stengthen the development of such extensions, see [here](https://github.com/ipython-contrib/jupyter_contrib_nbextensions/tree/master/src/jupyter_contrib_nbextensions/nbextensions/code_prettify), with documentation and examples for defining a new plugin. 

For now, `code_prettify`, `autopep8`,       `2to3` extensions are defined as plugins in the framework. It would probably a good idea to use it also for `jupyter_isort` (though it works as it is, of course)

jcb91:
Good point @jfbercher, thanks for that!

benjaminabel:
One problem with the current implementation of `code_prettify` is that it runs prettification one cell at a time. It would be nice to do an isort and move all imports in the first code cell and currently and don't see a a way to do that in code prettify.

jfbercher:
- I would be happy with an i-sort in each cell where imports occur (because I have use cases where it is better for the presentation that imports appear just before they are used). This and the processing of all cells in the notebook can be tackled with the library. 

- Moving all imports in the first cell is a different thing. What I would do
    - add a button and a keyboard shortcut that trig
   - a loop over all cells to grab the imports, then remove them from cells, create a cell at top, paste the imports in it and finally call isort on it. 

Something like that (for the loop)
```javascript
var imports = [];
var ind = 0;

var cells = IPython.notebook.get_cells();
cells.forEach(function(cell){
     var text = cell.get_text();
     // get imports
     var imp = text.match(/^import[\s\S]*?$/gm) 
     var from = text.match(/^from [\s\S]*? import[\s\S]*?$/gm) 
     // concat imports in current cell to the list of imports in the whole notebook
[...]

PR #1204, opened 2018-01-21
[isort] First implementation based on code_prettify

benjaminabel:
See #1202 for discussion. Actually I didn't add keyboard shortcuts because I can't find any
combination that works on my machine.

### DIFF
--- a/src/jupyter_contrib_nbextensions/nbextensions/code_prettify/README_isort.md
+++ b/src/jupyter_contrib_nbextensions/nbextensions/code_prettify/README_isort.md
@@ -1,0 +1,35 @@
+# Sort imports using isort
+
+This nbextension sorts imports in notebook code cells.
+
+Under the hood, it uses a call to the current notebook kernel to reformat the code. The conversion run by the kernel uses Python's package [isort](https://github.com/timothycrosley/isort) by [Timothy Edmund Crosley](https://github.com/timothycrosley).
+
+The nbextension provides
+
+- a toolbar button (configurable to be added or not)
+
+**pre-requisites:** of course, you must have the corresponding package installed:
+
+```
+pip install isort [--user]
+```
+
+## Options
+
+All options are provided by the [KerneExecOnCells library](kernel_exec_on_cell.js). There are a few nbextension-wide options, configurable using the [jupyter_nbextensions_configurator](https://github.com/Jupyter-contrib/jupyter_nbextensions_configurator) or by editing the `notebook` section config file directly. The options are as follows:
+
+- `isort.add_toolbar_button`: Whether to add a toolbar button to transform the selected cell(s). Defaults to `true`.
+
+- `isort.button_icon`: A font-awesome class defining the icon used for the toolbar button and actions. See [fontawesome.io/icons] for available icon classes. Defaults to `fa-sort`.
+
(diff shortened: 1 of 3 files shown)


jcb91:
> I can't find any combination that works on my machine

When you say 'works', in what sense have the ones youo've tried not worked? Or do you mean that you just can't find any more spare defaults?

src/jupyter_contrib_nbextensions/nbextensions/code_prettify/README_isort.md
```diff
@@ -0,0 +1,29 @@
...
+
+The nbextension provides
+
+- a toolbar button (configurable to be added or not)
+
+## Options
+
+All options are provided by the [KerneExecOnCells library](kernel_exec_on_cell.js). There are a few nbextension-wide options, configurable using the [jupyter_nbextensions_configurator](https://github.com/Jupyter-contrib/jupyter_nbextensions_configurator) or by editing the `notebook` section config file directly. The options are as follows:
+
+- `2to3.add_toolbar_button`: Whether to add a toolbar button to transform the selected cell(s). Defaults to `true`.
```
jcb91: these options in the readme need `2to3` replacing with `isort`

src/jupyter_contrib_nbextensions/nbextensions/code_prettify/isort.yaml
```diff
@@ -0,0 +1,45 @@
...
+    toolbar button. See http://fontawesome.io/icons/ for available icons.
+  input_type: text
+  default: 'fa-sort'
+
+- name: isort.button_label
+  description: Toolbar button label text
+  input_type: text
+  default: 'Sort imports with isort'
+
+- name: isort.show_alerts_for_errors
```
jcb91: since you've caught all exceptions in the library function, this option might be better of not listed, since it won't actually do anything

benjaminabel:
Thanks for your review @jcb91, I've amended the changes.

jcb91:
Great, looks good to me :+1: @benjaminabel Could you also submit an equivalent PR to the upstream [jfbercher/code_prettify](https://github.com/jfbercher/code_prettify)?

jcb91:
actually, no need, I'll do that...

benjaminabel:
I didn't know existence about this repository. Why don't you include it as a git submodule?

jcb91:
> Why don't you include it as a git submodule?

It more or less is already in practice, since they share history. I think you're probably right, it'd make sense to make this a submodule. Thoughts @jfbercher?

jfbercher:
@jcb91 Sorry for being late to reply. 
Actually, I think it could be easier to bring everything in `jupyter_contrib_nbextensions` ; then I would delete the upstream demo or mention that development goes on here.

jcb91:
> Sorry for being late to reply

No problem at all! This is nobody's job, after all ;)

> Actually, I think it could be easier to bring everything in `jupyter_contrib_nbextensions`; then I would delete the upstream demo or mention that development goes on here

Oh, ok! I don't really mind either way, so happy to go with whatever works for you :+1: Feel free to close #1208

jcb91:
Ok, I'm going to merge this in here, assuming this copy to be the definitive one, as suggested above.

benjaminabel:
Thanks for merging. Do we now add an action `process_all_cells_as_one` to the kernel_exec_on_cell object?

PR #1206, opened 2018-01-23
[exercise,exercise2]

### LINKED ISSUE #1079: exercise2 extension cells showing up in some notebooks but not others

jcb91:
fairly large reworking of exercise & exercise2:
* Don't use rubberband. Since multi-cell selection is available in 4.0+, so we can safely drop the rubberband for those not already using it
* align exercise & exercise2 to match each other
* separate html element creation/update/removal from metadata stuff
* simplify functions for creating/removing exercises, clicking controls by using metadata over thml elements
* clarify exercise (question + answer) vs solution (just answer) wording
* refresh exercises' solution visibility both when loading nbextension & on loading notebook (fixes #1079)

### DIFF
--- a/src/jupyter_contrib_nbextensions/nbextensions/exercise/main.js
+++ b/src/jupyter_contrib_nbextensions/nbextensions/exercise/main.js
@@ -27,8 +27,7 @@ define([
     'jquery',
     'require',
     'base/js/events',
-    'nbextensions/rubberband/main'
-], function(IPython, $, requirejs, events, rubberband) {
+], function(IPython, $, requirejs, events) {
     "use strict";
 
     /**
@@ -159,8 +158,6 @@ function load_ipython_extension(){
       };
 
     load_css('./main.css');
-    var exercise_wrapper = $('<div id="dragmask" class="highlight-drag"></div>');
-    $("#header").append(exercise_wrapper);
 
     // ***************** Keyboard shortcuts ******************************
     var add_cmd_shortcuts = {
@@ -198,8 +195,6 @@ function load_ipython_extension(){
         }
     }
 
-    console.log("Executing rubberband load_ipython")
-    rubberband.load_ipython_extension();
 }
 
 
--- a/src/jupyter_contrib_nbextensions/nbextensions/exercise/main.js
+++ b/src/jupyter_contrib_nbextensions/nbextensions/exercise/main.js
@@ -7,15 +7,15 @@
 December  6, 2017 @jcb91: use bootstrap 'hidden' class to play nicely with collapsible_headings
 December 30, 2015: update to 4.1
 Update december 22, 2015:
-  Added the metadata solution_first to mark the beginning of an exercise. It is now possible to have several consecutive exercises. 
-Update october 21-27,2015: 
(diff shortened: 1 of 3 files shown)


jcb91:
apologies @jfbercher, this is an alternative to your #1203, which I should have pushed up when I wrote it :disappointed:. It covers both ex & ex2

src/jupyter_contrib_nbextensions/nbextensions/exercise/main.js
```diff
@@ -34,169 +34,116 @@ define([
...
-}
+    function load_ipython_extension() {
+        // add css
+        $('<link rel="stylesheet" type="text/css">')
+            .attr('href', requirejs.toUrl('./main.css'))
+            .appendTo('head');
+
+        // Hide/display existing solutions at startup
+        events.on('notebook_loaded.Notebook', refresh_exercises);
+        if (IPython.notebook._fully_loaded) refresh_exercises();
```
jfbercher: Isn't necessary to check that `IPython.notebook != undefined`?

jfbercher:
No problem. Your version is largely better.

src/jupyter_contrib_nbextensions/nbextensions/exercise/main.js
jcb91: notebook is (at least in recent versions, maybe not in 3.x?) always _defined_ before any nbextensions are called, even if not fully loaded

dkirkby:
Thanks for the update @jcb91!

The students in my class originally installed exercise2 from 0.3.3 using:
```
conda install -c conda-forge libiconv jupyter_contrib_nbextensions
```
Do you have any idea when this update to master might show up on conda-forge so that they can use:
```
conda update jupyter_contrib_nbextensions
```
Or is there some other recipe to get this update sooner that students could reliably follow?

jcb91:
> Do you have any idea when this update to master might show up on conda-forge

Well, I released 0.4.0 on pypi a couple of days ago, but the windows builds for the conda-forge release have been failing (see https://github.com/conda-forge/jupyter_contrib_nbextensions-feedstock/pull/24 for details), so I haven't managed to push out 0.4.0 there yet. Unfortunately, even if I had, it still doesn't help in this case, as this was merged after 0.4.0.

So, currently the alternatives are

1. wait for 0.5.0, which I guess I could put out soon after 0.4.0, once I figure out how to get the windows builds rolling again
2. remove the conda package, then follow the main readme instructions to install the git-master version, using
   ```
   pip install https://github.com/ipython-contrib/jupyter_contrib_nbextensions/tarball/master
   jupyter contrib nbextension install --sys-prefix
   ```
   (this pip-install then contrib-install is what the conda recipe does anyway)

Issue #1210, opened 2018-01-27
Snippets extension causes notebooks (v5.3.1) to not stay in "Running" state


pybokeh:
Please see the thread at the jupyter notebook's closed at https://github.com/jupyter/notebook/issues/3245

juhasch:
It is hard to understand how the snippet extension could cause this.
Can you post the output of your Browser console when this happens ?
(on Chrome you can see this when enabling developer tools)

pybokeh:
I apologize, I am not sure I am providing the best output format. I copied the console output and pasted here:

```
menubar.js:304 actions jupyter-notebook:find-and-replace does not exist, still binding it in case it will be defined later...
MenuBar.bind_events @ menubar.js:304
MenuBar @ menubar.js:59
(anonymous) @ main.js:145
execCb @ require.js?v=6da8be361b9ee26c5e721e76c6d4afce:1690
check @ require.js?v=6da8be361b9ee26c5e721e76c6d4afce:865
(anonymous) @ require.js?v=6da8be361b9ee26c5e721e76c6d4afce:1140
(anonymous) @ require.js?v=6da8be361b9ee26c5e721e76c6d4afce:131
(anonymous) @ require.js?v=6da8be361b9ee26c5e721e76c6d4afce:1190
each @ require.js?v=6da8be361b9ee26c5e721e76c6d4afce:56
emit @ require.js?v=6da8be361b9ee26c5e721e76c6d4afce:1189
check @ require.js?v=6da8be361b9ee26c5e721e76c6d4afce:940
enable @ require.js?v=6da8be361b9ee26c5e721e76c6d4afce:1177
init @ require.js?v=6da8be361b9ee26c5e721e76c6d4afce:783
callGetModule @ require.js?v=6da8be361b9ee26c5e721e76c6d4afce:1204
completeLoad @ require.js?v=6da8be361b9ee26c5e721e76c6d4afce:1583
onScriptLoad @ require.js?v=6da8be361b9ee26c5e721e76c6d4afce:1711
[...]

PR #1212, opened 2018-01-28
Use base_url in snippets

juhasch:
Fixes issue #1210


jcb91:
Good job catching this @juhasch :)

For full generality (e.g. someone at some point renames snippets directory, puts it in a subdirectory, etc), we could add `'require'` to the requirements list, then use a relative path like `requirejs.toUrl('./snippets.json')`

But I think this version is fine here, since we don't add any instructions for separate installs anyway.

juhasch:
Thanks.

Issue #1193, opened 2018-01-12
Building a notebook extension that allows to register other cells as its execution dependencies


benelot:
Hello everyone,

I am thinking about writing an extension that allows for each cell to be annotated with cells on which its execution depends on. Working in a modular fashion could be simplified because it would allow to write short code segments that need to be run as initialization for others. For example, I can write a dataset download and extract section, which is then set as a dependency for a graph drawing cell. If I run the graph drawing cell, the dependent cells will be run before it to ensure its successful execution.

Can anybody give me some hints on where to start (what part of the doc I should read, not even sure if this is a front-end or back-end extension maybe both?) and how this could be done (as rough pseudocode like "you catch the event notification of the cell to be run and trigger all cells that are in its json tag called deps. The deps you can collect for each cell by inserting deps as special tags").

juhasch:
From the technical side this should not be too difficult to implement. You can take look at the runtools extensions for example, where you can mark cells and execute them.

It is not difficult to programmatically execute a given cell, or modify the execution behavior. This can be done using a custom execution command (called from a hotkey, menu entry or button), or you can change to logic by catching events, or overwriting the default execution function, implementing your own logic. Unfortunately there is no "official" API. Best start by looking at existing extensions in this or other repositories.

I think the more difficult part is to create a user interface that allows doing what you want in an intuitive and comprehensible manner. You might want to look at cell tags. You can add self-defined tags to a cell. Executing cells with this tag is a simple for loop.

benelot:
I think I would directly use the UI for the tags to represent the dependencies. Something like #A for the id and =>B for the dependency could be a simple representation of the dependency graph. Thanks for the hints, that is EXACTLY the kind of help I hoped for. I will keep you posted on the progress.

jcb91:
> Something like #A for the id and =>B for the dependency could be a simple representation of the dependency graph.

Ah yes, good plan! Nice, neat idea, and simple to implement :)

PR #1213, opened 2018-01-29
Execution dependency extension

benelot:
I am implementing the extension for custom execution dependencies as discussed in #1193. It seems to work and properly resolves the dependency tree (it even handles circular deps in a simple way). However, in the end it collects all cells in the order found by the topological sort and tries to run them, but then the java script crashes/and the script starts to run anew.

DISCLAIMER: I have limited experience with JavaScript, so there might be quirky stuff done in code which I comes from my experience as a Java developer. But the problem does not seem to be in that part, but seems to be somewhere else.

Any idea what is wrong? Second question I have is if there is a neat way I can tell the user that a circular dependency is preventing the dependencies to run. For now it just prints to the console.

To test it, simply create two python cells and add a #A tag to the first and a =>A tag to the second. The run the second and look into the developer console. There you can see that it restarts the script after it prints "Execute cells.."

### DIFF
--- a/src/jupyter_contrib_nbextensions/nbextensions/execution_dependencies/README.md
+++ b/src/jupyter_contrib_nbextensions/nbextensions/execution_dependencies/README.md
@@ -1,0 +1,43 @@
+execution_dependencies
+======================
+
+Writing extensive notebooks can become very complicated since many cells act as stepping stones to produce intermediate results for later cells. Thus, it becomes tedious to
+keep track of the cells that have to be run in order to run a certain cell. This extension simplifies handling the execution dependencies by introducing tag annotations to
+identify each cell and indicate a dependency on others. This improves on the current state which requires remembering all dependencies by heart or annotating the cells in the comments.
+
+If a cell with dependencies is run, the extension checks recursively for all dependencies of the cell, then executes them before executing the cell after all the dependencies have finished.
+Dependencies are definitely executed and not only once per kernel session.
+
+The two annotations are added to the tags of a cell and are as follows:
+
+ * add a hashmark (#) and an identification tag to the tags to identify a cell (e.g. #initializer-cell). The #identifiers must be unique among all cells.
+ * add an arrow (=>) and an identification tag to the tags to add a dependency on a certain cell (e.g. =>initializer-cell).
+
+Based on these dependencies, the kernel will now execute the dependencies before the cell that depends on them. If the cell's dependencies have further dependencies, these will in turn
+be executed before them. In conclusion, the kernel looks through the tree of dependencies of the cell executed by the user and executes its dependencies in their appropriate order, 
+then executes the cell.
+
+A more extensive example is described below:
+
+A cell A has the identifier #A.
+
+| Cell A [tags: #A]  |
+| ------------- | 
+| Content Cell  |
+| Content Cell  |
+
+
+A cell B has the identifier #B and depends on A (=>A).
+
+
+| Cell B [tags: #B, =>A]  |
+| ------------- | 
+| Content Cell  |
+| Content Cell  |
+
+If the user runs A, only A is executed, since it has no dependencies. On the other hand, if the user runs B, the kernel finds the dependency on A, and thus first runs A and then runs B.
+
+Running a cell C that is dependent on B and on A as well, the kernel then first runs A and then runs B before running C, avoiding to run cell A twice.
+
+
+If you are missing anything, open up an issue at the repository prepending [execute_dependencies] to the title.
--- a/src/jupyter_contrib_nbextensions/nbextensions/execution_dependencies/execution_dependencies.js
+++ b/src/jupyter_contrib_nbextensions/nbextensions/execution_dependencies/execution_dependencies.js
@@ -1,0 +1,138 @@
+/**
+ * execution_dependencies.js
+ * Introduce tag annotations to identify each cell and indicate a dependency on others.
+ * Upon running a cell, its dependencies are run first to prepare all dependencies. 
+ * Then the cell triggered by the user is run as soon as all its dependencies are met.
+ *
+ *
+ * @version 0.1.0
+ * @author  Benjamin Ellenberger, https://github.com/benelot
+ * @updated 2018-01-31
+ *
+ *
+ */
+define([
+    'jquery',
+    'base/js/dialog',
+    'base/js/namespace',
+    'notebook/js/codecell'
+], function (
+    $,
+    dialog,
+    Jupyter,
+    codecell
+) {
+    "use strict";
+
+    var CodeCell = codecell.CodeCell;
+
+    return {
+        load_ipython_extension: function () {
+            console.log('[execution_dependencies] patching CodeCell.execute');
+            var orig_execute = codecell.CodeCell.prototype.execute;                                        // keep original cell execute function
+            CodeCell.prototype.execute = function (stop_on_error) {
+                var root_tags = this.metadata.tags || [];                                                  // get tags of the cell executed by the user (root cell)
+                if(root_tags.some(tag => /=>.*/.test(tag))) {                                              // if the root cell contains any dependencies, resolve dependency tree
+                    var root_cell = this;
+                    var root_cell_id = root_cell.cell_id;
+                    var cells_with_id = Jupyter.notebook.get_cells().filter(function (cell, idx, cells) {  // ...get all cells which have at least one id (these are the only ones we could have in deps)
+                        var tags = cell.metadata.tags || [];                        
+                        return (cell === root_cell || tags.some(tag => /#.*/.test(tag)));
+                    });
+
+                    console.log('[execution_dependencies] collecting ids and dependencies...');
+                    var cell_map = {}  
+                    var dep_graph = {}
+                    cells_with_id.forEach(function (cell) {                                                // ...get all identified cells (the ones that have at least one #tag)
+                        var tags = cell.metadata.tags || [];
+                        var cell_ids = tags.filter(tag => /#.*/.test(tag)).map(tag => tag.substring(1));   // ...get all identifiers of the current cell and drop the #
+                        if(cell === root_cell){                        
+                            if(cell_ids.length < 1) {
+                                cell_ids.push(root_cell.cell_id);                                              // ...use internal root cell id for internal usage
+                            }
+                            else {
+                                root_cell_id = cell_ids[0];                                                    // get any of the root cell ids
+                            }
+                        }
+
+                        var dep_ids = tags.filter(tag => /=>.*/.test(tag)).map(tag => tag.substring(2));   // ...get all dependencies and drop the =>
+                        
+                        cell_ids.forEach(function (id) {
+                          //console.log('ID:', id, 'deps: ', dep_ids.toString())
+                            cell_map[id] = cell;
+                            dep_graph[id] = dep_ids;
+
+                        });
+                    });
+
+                    if(dep_graph[root_cell_id].length > 0) {
+                        console.log('[execution_dependencies] collecting depdendency graph in-degrees...');
+                        var processing_queue = [root_cell_id];
+                        var processed_nodes = 0;
+                        var in_degree = {};                                                                    // ...collect in-degrees of nodes
+                        while(processing_queue.length > 0 && processed_nodes < Object.keys(dep_graph).length) {// ...stay processing deps while the queue contains nodes and the processed nodes are below total node quantity
+                            var id = processing_queue.shift();                                                 // .....pop front of queue and front-push it to the processing order
+                          //console.log("ID: ", id);
+                            for(var i=0, dep_qty=dep_graph[id].length; i < dep_qty; i++) {
+                                var dep = dep_graph[id][i];
+                          //    console.log('      dep: ', dep);
+                                in_degree[id] = in_degree[id] || 0;
+                                in_degree[dep] = in_degree[dep] === undefined ? 1 : ++in_degree[dep];
+                                processing_queue.unshift(dep);
+                            }
+                        }
+
+                        console.log('[execution_dependencies] starting topological sort...');
+                        processing_queue = [root_cell_id];                                                     // ...add root node with in-degree 0 to queue (this excludes all disconnected subgraphs)
+                        processed_nodes = 0;                                                                   // ...number of processed nodes (to detect circular dependencies)
+                        var processing_order = [];
+                        while(processing_queue.length > 0 && processed_nodes < Object.keys(dep_graph).length) {// ...stay processing deps while the queue contains nodes and the processed nodes are below total node quantity
+                            var id = processing_queue.shift();                                                 // .....pop front of queue and front-push it to the processing order
+                            processing_order.unshift(id);
+                          //console.log("ID: ", id);
+                            for(var i=0, dep_qty=dep_graph[id].length; i < dep_qty; i++) {                     // ......iterate over dependent nodes of current id and decrease their in-degree by 1
+                                var dep = dep_graph[id][i];
+                          //    console.log('      dep: ', dep);
+                                in_degree[dep]--;
+                                if(in_degree[dep] == 0) {                                                      // ......queue dependency if in-degree is 0
+                                    processing_queue.unshift(dep);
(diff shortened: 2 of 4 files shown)


src/jupyter_contrib_nbextensions/nbextensions/execution_dependencies/execution_dependencies.js
```diff
@@ -0,0 +1,96 @@
...
+                    identified_cells.forEach(function (cell) {                                                      // ...get all identified cells (the ones that have at least one #tag)
+                        var tags = cell.metadata.tags || [];
+                        var identities = tags.filter(tag => /#.*/.test(tag)).map(tag => tag.substring(1));          // ...get all identities and drop the #
+                        if(cell === root_cell && !tags.some(tag => /#.*/.test(tag))) {
+                            identities.push("DD27AE1D138027D0D7AB824FD0DDDC61367D5CCA4AAB42CE50840762B053764D");    // ...generate an id for the root cell for internal usage
+                        }
+
+                        var deps = tags.filter(tag => /=>.*/.test(tag)).map(tag => tag.substring(2));               // ...get all dependencies and drop the =>
+                        identities.forEach(function (id) {
+                            cell_map[id] = cell;
```
jcb91: this line implicitly means only a single cell can have each `#some_id` tag. Is this what you want? It could make sense to have a few cells with the same tag, so do something like 

```javascript
cell_map[id] = cell_map[id] || [];
cell_map[id].push(cell);
```

Alternatively, if you don't intend to support this, at least let the user know that adding multiple cells with the same identifier tag won't work correctly in the readme.

src/jupyter_contrib_nbextensions/nbextensions/execution_dependencies/execution_dependencies.js
```diff
@@ -0,0 +1,96 @@
...
+                            in_degree[dep]--;
+                            if(in_degree[dep] == 0) {                                                               // ......queue dependency if in-degree is 0
+                                processing_queue.unshift(dep);
+                            }
+                        }
+                        processed_nodes++;
+                    }
+
+                    if(processed_nodes >= Object.keys(dep_graph).length) {
+                        console.error('There is a circular dependency in your execute dependencies!');
```
jcb91: you could do an `alert` here, or alternatively use the notebook dialog for something prettier. See https://github.com/ipython-contrib/jupyter_contrib_nbextensions/blob/a4544eca1182db2816523959002b5a31c263e267/src/jupyter_contrib_nbextensions/nbextensions/init_cell/main.js#L106-L115 for an example usage.

In fact, it probably makes sense for this extension to refuse to execute dependencies in untrusted notebooks in the same way that init_cell does...

src/jupyter_contrib_nbextensions/nbextensions/execution_dependencies/execution_dependencies.js
```diff
@@ -0,0 +1,96 @@
...
+                    }
+                    else{
+                        console.log("Map processing order to cells...", processing_order)
+                        var dependency_cells = processing_order.map(id =>cell_map[id]);                             // ...get dependent cells by their id
+                        console.log("Execute cells..", dependency_cells)
+                        dependency_cells.forEach(function (cell) { cell.execute(stop_on_error); });                 // ...execute all dependent cells in order
```
jcb91: I think this may be the main problem you're seeing. Since you've called cell.execute, this will call your patched version, so look for dependencies again, etc. Instead, you should almost certainly be doing

```javascript
dependency_cells.forEach(cell => orig_execute.call(cell, stop_on_error));
```

src/jupyter_contrib_nbextensions/nbextensions/execution_dependencies/execution_dependencies.js
```diff
@@ -0,0 +1,96 @@
...
+                    }
+                }
+                orig_execute.call(this, stop_on_error);                                                             // execute original cell execute function
+            };
+            console.log('[exec_deps] loaded');
```
jcb91: this should use the updated name of `execution_dependencies`

src/jupyter_contrib_nbextensions/nbextensions/execution_dependencies/README.md
```diff
@@ -0,0 +1,38 @@
+execution_dependencies
+======================
+
+Writing extensive notebooks can become very complicated since many cells act as stepping stones to produce intermediate results for later cells. Thus, it becomes tedious to
+keep track of the cells that have to be run in order to run a certain cell. This extension simplifies handling the execution dependencies by introducing tag annotations to
+identify each cell and indicate a dependency on others. This improves on the current state which requires remembering all dependencies by heart or annotating the cells in the comments.
```
jcb91: might be worth noting that dependencies are *definitely* executed, rather than say only being executed once per kernel session. This may be important for cells which take a long time to execute...

src/jupyter_contrib_nbextensions/nbextensions/execution_dependencies/execution_dependencies.js
benelot: Oh, that is how this works! Thanks for pointing it out to me!

src/jupyter_contrib_nbextensions/nbextensions/execution_dependencies/execution_dependencies.js
benelot: Sounds great, I will add it.

src/jupyter_contrib_nbextensions/nbextensions/execution_dependencies/execution_dependencies.js
jcb91: >  Is the cell_id you mentioned above unique within the notebook's context? Then I could use that for uniqueness and allow multiple cells to have the same hashtag id.

Yes, it should be unique for the current session. See [notebook/static/notebook/js/cell.js#L101](https://github.com/jupyter/notebook/blob/3de5e501f3f1eb16d89830b658d03b9c3e1a4c8b/notebook/static/notebook/js/cell.js#L101) for where it gets assigned, and [notebook/static/base/js/utils.js#L206-L220](https://github.com/jupyter/notebook/blob/3de5e501f3f1eb16d89830b658d03b9c3e1a4c8b/notebook/static/base/js/utils.js#L206-L220) for the implementation

src/jupyter_contrib_nbextensions/nbextensions/execution_dependencies/execution_dependencies.js
```diff
@@ -0,0 +1,96 @@
...
+                        return (cell === root_cell || tags.some(tag => /#.*/.test(tag)));
+                    });
+
+                    console.log('Collect ids and dependencies...');
+                    var cell_map = {}  
+                    var dep_graph = {}
+                    identified_cells.forEach(function (cell) {                                                      // ...get all identified cells (the ones that have at least one #tag)
+                        var tags = cell.metadata.tags || [];
+                        var identities = tags.filter(tag => /#.*/.test(tag)).map(tag => tag.substring(1));          // ...get all identities and drop the #
+                        if(cell === root_cell && !tags.some(tag => /#.*/.test(tag))) {
```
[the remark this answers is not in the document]
benelot: A single cell can have multiple identities, that is why it is called identities. I do not see why they would be required_ids. Below I just add an id, because the root cell does not need an id since it could be the one that is independent (the user is just working on that one and would not understand why it needs an id to run.).

src/jupyter_contrib_nbextensions/nbextensions/execution_dependencies/execution_dependencies.js
```diff
@@ -0,0 +1,96 @@
...
+
+    var CodeCell = codecell.CodeCell;
+
+    return {
+        load_ipython_extension: function () {
+            console.log('[exec_deps] patching CodeCell.execute');
+            var orig_execute = codecell.CodeCell.prototype.execute;                                                 // get original cell execute function
+            CodeCell.prototype.execute = function (stop_on_error) {
+                var root_tags = this.metadata.tags || [];
+                if(root_tags != [] && root_tags.some(tag => /=>.*/.test(tag))) {                                    // if the root cell contains any dependencies, resolve dependency tree...
```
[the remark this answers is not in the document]
benelot: I was not sure if this.metadata.tags would be [] or something like null/undefined. That is where my javascript knowledge ends. But I see that since I either have [...] or [], so it is defined and I can ask for its length.

src/jupyter_contrib_nbextensions/nbextensions/execution_dependencies/README.md
benelot: True, I should add that. Thanks for reviewing my code, I am very grateful for that! I really like jupyter notebooks and I am very interested in contributing more extensions in the future.

src/jupyter_contrib_nbextensions/nbextensions/execution_dependencies/execution_dependencies.js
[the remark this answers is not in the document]
jcb91: Exactly. `this.metadata.tags` may often be `undefined`, e.g. if no tags have been set. In fact, in a trickier world, it could be anything that'll fit into the notebook JSON, but in practice we can assume that it will be either an array of strings, or undefined (if it's not, something more significant has gone wrong with the notebook somewhere, and this extension breaking is likely to be the least of anyone's worries)

src/jupyter_contrib_nbextensions/nbextensions/execution_dependencies/execution_dependencies.js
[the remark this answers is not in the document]
jcb91: apologies, I misunderstood, for some reason I thought this was a list of dependencies' `id`s. No rename necessary :+1:

src/jupyter_contrib_nbextensions/nbextensions/execution_dependencies/README.md
jcb91: No problem, happy to help :smile: And of course, we'd be happy to have anything you think might be useful to include here :wink:

src/jupyter_contrib_nbextensions/nbextensions/execution_dependencies/execution_dependencies.js
benelot: Sounds both reasonable. I will look into this as soon as the rest works.

benelot:
One more thing, how do I know about compatibility? I currently have all compatibilities in, but all I know is that it is compatible with 5.x.

benelot:
Ok, I am done. You can merge it as soon as you are ready and there are no other issues to solve. I implemented everything we discussed except for the one #id for multiple cells. I will do that later or if there is a large request for that feature.

benelot:
It does not work yet on my larger example. This needs some more time.

jcb91:
For compatibility, I just check whether the functions I'm calling exist in the notebook of the given version (4.0.0 for 4.x, etc). You can do this quite conveniently on github by looking at the notebook repo's tags, e.g. https://github.com/jupyter/notebook/tree/4.0.0/notebook/static. In general though, this nbextension isn't using too many fancy notebook methods that have changed recently, like multiple selections, so I'd guess it's compatible with 4.x & 5.x. Probably the most likely thing to break 3.x might be the dialog but even that may be fine. However, I wouldn't worry much about 3.x because I haven't heard of anyone using it for quite some time now...


> It does not work yet on my larger example. This needs some more time.

Sure, will hold off merge till you're ready

benelot:
Fixed some problems with processing if the dependencies are only a subgraph within many completely unrelated cells. Thanks for all the help!

I will keep on my roadmap:
* One #tag for multiple cells
* Execution cells as dependencies only once per kernel session (option switchable to force reexecution of cells).

benelot:
Looks like it works now! I used it for the rest of the day and it did not break. Ready to be pulled!

jcb91:
Great, thanks. Could you add a note to the changelog under the github master section, just noting the new extension? Apologies, I should have mentioned this before :man_facepalming:

benelot:
Of course I can.

benelot:
There you go. Btw I am working on the improvement to make the dependent cell only if it has not been run before or if it has been modified since its last run. Are there any flags that tell me (if/the last time) a cell has been run and (if/the last time) it has been modified? Maybe I am looking in the wrong places because I can not find it.

jcb91:
Thanks :)

>  Are there any flags that tell me (if/the last time) a cell has been run and (if/the last time) it has been modified?

No, there aren't in the main notebook. There's the [ExecuteTime nbextension](https://github.com/ipython-contrib/jupyter_contrib_nbextensions/tree/master/src/jupyter_contrib_nbextensions/nbextensions/execute_time), but it doesn't record modifications, doesn't keep track of which kernel was running, and only takes into account the current browser frontend. The lack of such a feature is, I think at least in part, by design. The kernel is intended to be frontend-agnositc. That is to say, it should not care (or expect to be able to tell) whether it is being sent commands by a notebook, a terminal, a qt console, your own custom client, whatever. As a result, the kernel has no notion of cells (although I guess one could consider each ipython `In` element a cell, but that's language-dependent, and getting off-topic). In addition, since the frontend doesn't know what's happened on this kernel in the past, it will probably only be able to tell what has/hasn't been executed since the page loaded (which may be a long time after kernel creation, since the page can be refreshed). I suppose one could implement local storage or metadata structures to get around this, but that might be more effort than is really worth it.

However, in many (most?) cases, people are using kernels which are only connected to a single notebook frontend, and in such cases, we could get away with a relatively simple model where the nbextension could record cells' ids as they are executed, removes them from the executed list if they get modified or the kernel restarts, and therefore check whether they need re-executing, say if they use `->` to specify a dependency rather than `=>`. This would err on the side of re-executing if in doubt, since it would 'forget' cells had been executed whenever the page was refreshed. Make sense? Maybe it's not worth implementing imperfectly in the frontend, and better to just get people to do the checking in the cells' kernel code.

> Also, the Changelog's pull request links should be updated, they are all invalid except for mine. They are missing the repository name.

:man_facepalming: d'oh I just fixed this for the links for the 3.x releases, then clearly managed to make the same mistake again :( thanks for pointing this out :+1:, I'll add a PR to fix these.

> It seems that the build failed because my link is invalid even though it is the only one that works. Or is it only me?

Hmm. The lint build seems to fail because isort has changed its mind about ordering imports _again_. That can be safely ignored. The docs build seems to fail on a bunch of links which look like they should be fine, so I've restarted it to check. I don't think it's anything to do with your edit, which looks fine to me.

> P.S. I made a contribution page for you because I liked the way things were going with my first contribution. Maybe have a review on it to see if that is how you like your contributions:  [Create A new Notebook Extension](https://github.com/ipython-contrib/jupyter_contrib_nbextensions/wiki/Create-a-new-Notebook-Extension)

I guess you mean this to be a `CONTRIBUTING.md` file? Maybe replace "I" with your handle @benelot for clarity? Otherwise, sure, sounds like a good plan to me. I have a feeling we maybe once said we adhere to the ipython contributing guidelines, but I forget where I read that, or what they might be. At any rate, we can always update the file later as necessary...

src/jupyter_contrib_nbextensions/nbextensions/execution_dependencies/execution_dependencies.js
```diff
@@ -0,0 +1,138 @@
...
+
+                        });
+                    });
+
+                    if(dep_graph[root_cell_id].length > 0) {
+                        console.log('[execution_dependencies] collecting depdendency graph in-degrees...');
+                        var processing_queue = [root_cell_id];
+                        var processed_nodes = 0;
+                        var in_degree = {};                                                                    // ...collect in-degrees of nodes
+                        while(processing_queue.length > 0 && processed_nodes < Object.keys(dep_graph).length) {// ...stay processing deps while the queue contains nodes and the processed nodes are below total node quantity
```
jcb91: this loop never increments `processed_nodes`, so gets stuck for circular dependencies...

benelot: Added it to the code. Ready for merging.

benelot:
Yeah!

PR #1220, opened 2018-02-02
[toc2] add controls for all per-notebook metadata options

jcb91:
inspired in part by https://stackoverflow.com/questions/47726971, adds controls to edit notebook metadata for the options

* `sideBar` (display as sidebar rather than floating window)
* `toc_window_display` (Display ToC window/sidebar at startup)
* `toc_section_display` (Expand window/sidebar at startup)


jfbercher:
Sounds good to me

jfbercher:
Thanks @jcb91 . Perhaps add a comment at https://stackoverflow.com/questions/47726971?

PR #1226, opened 2018-02-04
FIXES #1223 and improves the user experience

### LINKED ISSUE #1223: init_cell - Unnecessary dialog in untrusted notebooks without initialization cells
Having the init_cell plugin enabled when opening an untrusted notebook will warn that it will not run the cells automatically, even though there are no initialization cells.

I provide my students with Jupyter through a JupyterHub and have added this extension by default as it could help me streamline distribution of material to them. My students will not know about this plugin, and I wan't to keep it that way unless there is a need for them to know about it.

I'll try make a PR fixing this as well as allowing them to run the code manually by clicking a button in the dialog.

consideRatio:
* FIXES #1223 - Only shows the dialog about untrusted notebooks if there was any initialization cells in it.
* UX improvement - Adding a button to do the same action as available in the menu "File -> Trust Notebook", it will require an additional confirmation.

init_code_ux

trust_notebook

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -118,6 +118,10 @@ New features and bugfixes:
  - `code_prettify` Update `code_prettify.yaml`
    [#1162](https://github.com/ipython-contrib/jupyter_contrib_nbextensions/pull/1162)
    [@fehiepsi](https://github.com/fehiepsi)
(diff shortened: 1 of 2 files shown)


jcb91:
👍 looks like a good plan to me.

PR #1229, opened 2018-02-07
Add "numpy shape" as size

HTLife:
I found that it's very convenient to know numpy shape in 'size' column instead of memory size.
Lots of people use jupyter to implement their deep learning code.  "shape" property is a more helpful info for debugging.
If this alternation is matched with the system design policy, it would be great to merge this feature into main repo.

### DIFF
--- a/src/jupyter_contrib_nbextensions/nbextensions/varInspector/var_list.py
+++ b/src/jupyter_contrib_nbextensions/nbextensions/varInspector/var_list.py
@@ -3,7 +3,7 @@ from sys import getsizeof
 
 from IPython import get_ipython
 from IPython.core.magics.namespace import NamespaceMagics
-
+import numpy as np
 _nms = NamespaceMagics()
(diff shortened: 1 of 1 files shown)


jcb91:
Thanks for this @HTLife. Neat idea, but I think it might be better to add an extra (optional?) column, or make this mode optional altogether, or at the very least add some explanation of what's going on with numpy arrays, rather than just make the column behaviour inconsistent without notifying the user. Also, the python needs a try/except to ensure it still works even if numpy is missing. I think something like this is a good enhancement to have in principle, though :+1:

jfbercher:
Yes, I agree that it would be better to add a new column and that this can be useful.  The shape should also be reported for pandas dataframe.

HTLife:
@jcb91  @jfbercher  Could you guys help me with that?   I'm not familiar with how to modify that part.  And I might learn from your commit ^_^

PR #1243, opened 2018-02-17
[kernel_exec_on_cell] fix shift-click handling

benjaminabel:
Fixed the label on buttons and correctly handle shift click on button based on data attribute.

BEFORE, the label was added to the button:

image

AFTER, the label is included correctly in the title of the button.

image

Also, the shiftKey event wasn't handled correctly and didn't work on my machine.


jcb91:
The label-added to button thing is a problem, yes. However, the shift-click behaviour works fine for me, as far as I can tell. What behaviour are you seeing?

Also, the new version appears to be a little confused. As it stands, the `handler` function (called to handle click events on the button) now just attaches _another_ new click handler to all button elements on the page, without actually doing what it's supposed to...

benjaminabel:
The shift-click didn't work on my desktop, my version did; but now I'm using `jupyterlab` .

PR #1244, opened 2018-02-17
Fix clicking on gutter handling

juhasch:
Don't select when clicking on codefolding gutter and fix codemirror change


juhasch:
Should fix #1192,too.

PR #1248, opened 2018-02-23
Fix/embedhtml2

juhasch:
Continuation of PR #1113 now that #1067 is merged.


jcb91:
Looks good to me :+1: A couple of questions:

* is there a reason to use `ntpath` over `os.path` for the basename in the exporter? I guess maybe because it will cut down windows-like paths generated on a windows machine, even if the current one is using posixpath (since actual path isn't important)? Might be worth adding a comment to explain.
* since `jupyter_contrib_nbextensions.nbconvert_support.embedhtml` has been renamed, should we add a backwards-compatibility stub module at the old location? I'm not sure whether anyone would be importing from there directly anyway though, I guess, since it's available from `nbconvert_support`. Sounds fine.

PR #1250, opened 2018-02-24
[toc2] Initialize toc2 button after cfg is loaded / fix #1249

### LINKED ISSUE #1249: toc2 - Left margin not restored when toggling on/off the sidebar
The margins are not restored when togging off the sidebar by pushing the toc button. 
This is due to the fact that function `toggle_toc`  in `toc2.js` fails [line 651](https://github.com/ipython-contrib/jupyter_contrib_nbextensions/blob/master/src/jupyter_contrib_nbextensions/nbextensions/toc2/toc2.js#L651) because `cfg` object is undefined.  Tentative fix follows.

jfbercher:
Tentative fix for #1249. 
Initialize `toc_button(cfg)` only after cfg has been defined (perhaps not read yet (?), but defined). Otherwise, the handler in `toc_button` points to an undefined object.

### DIFF
--- a/src/jupyter_contrib_nbextensions/nbextensions/toc2/main.js
+++ b/src/jupyter_contrib_nbextensions/nbextensions/toc2/main.js
@@ -63,6 +63,11 @@ define([
       }
     }
 
+    function toggleToc() {
+        toggle_toc(cfg)
+    }
+
+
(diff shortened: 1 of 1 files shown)


jfbercher:
Let's give a try to this

PR #1251, opened 2018-02-24
Fix some travis warnings and errors

juhasch:


juhasch:
@jfbercher :the link-checker fails here:
```
https://rawgit.com/jfbercher/jupyter_latex_envs/master/src/latex_envs/static/doc/latex_env_doc.html
(line   77) redirect  https://conda-forge.github.io/ - permanently to https://conda-forge.org/
(line  122) broken    https://github.com/benweet/stackedit/issue/187 - 404 Client Error: Not Found for url: https://github.com/benweet/stackedit/issue/187
```
Maybe you could take a look at it sometime ?

Not sure about the two OSX errors in travis.

jcb91:
@juhasch I _think_ I already made a pr to latex_envs to fix those links, so (if that's true), we just need a new pypi/conda release there to fix this

jcb91:
Yep: https://github.com/jfbercher/jupyter_latex_envs/pull/32

jfbercher:
> Yep: jfbercher/jupyter_latex_envs#32

Yes @jcb91, I have merged this one 4 days ago. Thanks. There was a incorrect link remaining in an html file. I was precisely working on a new version with some [corrections and improvements](https://rawgit.com/jfbercher/jupyter_latex_envs/master/src/latex_envs/static/doc/latex_env_doc.html#-What's-new-), and since I have regenerated the doc, this should be fixed :-)

jfbercher:
I have also made a Pypi release.

juhasch:
Great, so I will merge.

jcb91:
Thanks @juhasch. I'm afraid I have no extra info on the TLS issue. Travis issues list seems to have a few things with the same text, but they vary in their root cause from something Travis had to fix with their ruby install to certificate errors at the remote end, so it doesn't seem immediately obvious what the right course of action would be there...

juhasch:
What is the reason we need the MacOS tests anyway ?
They seem to take forever, and we don't have any binaries to test, so I would think testing on Linux would be enough ?

jcb91:
> They seem to take forever

Oh don't they just! 😂 So frustrating, particularly when Travis' macOS infrastructure falls over, which it seems to do a couple of times a year 🙄

>  I would think testing on Linux would be enough ?

Yeah, it probably is, I guess. I think it felt safer when first packaging everything for pypi to just test everything to be sure it would all work. But now, I think you're right, we can fairly safely assume that macOS is unix-y enough, and the builds can be dropped (or at least made allowed-failure). Worth keeping the appveyor windows ones though, as windows _is_ of course sometimes different, particularly with things like paths.

In fact, if we're going to drop the macOS builds, we could even switch to circleCI for the Linux ones (conda-forge use circle, and it seems pretty fast, though I've no direct comparison).

juhasch:
OK, so let's remove macOS at least for now. I agree keeping Windows tests is worthwhile.

Any idea what the remaining appveyorartifacts error is about ?

jcb91:
> Any idea what the remaining appveyorartifacts error is about ?

This build used to collect coverage files produced on appveyor, then submit them to coveralls or codecov (or maybe both? I forget). It hasn't worked fr a while, but I've not bothered to investigate why. To be honest, the coverage levels only relate to python code anyway, so are slightly misleading...

Issue #1245, opened 2018-02-22
TOC2: start h1 numbering at a predefined counter


fmaussion:
Sorry if this has been asked before: is there a way to start the numbering of the ``h1`` level at any other number than 1? 

I'd like to do so because my h1 headers are titles of "chapters", as defined by the [bookbook](https://github.com/takluyver/bookbook) tool.

jcb91:
Hmm. So if I understand correctly you want the h1 heading in a notebook to be say number 2, because it follows an h1 in another notebook? I don't think there is a way to do this so far, but I expect it would be relatively easy to implement as a feature the ability to manually set the starting number, if it'd be useful...

fmaussion:
Thanks for the quick reply! Yes, I would find it very useful indeed. But I'm not familiar at all with the development of nbextensions so I'm afraid I can't really help :(

fmaussion:
Actually, now that I think of it a bit more, the easiest way for this kind of structure might be to skip h1 headers altogether (which is already an option as I can see). 

I still think this would be a useful feature but maybe you don't have to put it on top of your priority list...

fmaussion:
For the context: I use toc2 essentially for it's ability to generate TOCs automatically, then use [bookbook](https://github.com/takluyver/bookbook) to generate html outputs ([example](http://fabienmaussion.info/scientific_programming/html/02-Installation.html))

The problems I see with the numbering is that since the titles are not automatically numbered in the html output, the toc is the only listing containing section numbers. Since I guess this is not something easily done on the TOC2 side, I might skip numbering altogether...

jfbercher:
Do you mean something like that?  

`<h1> something </h1>` --> no number something
`<h2> something </h2>` --> 1.1 something
`<h2> something </h2>` --> 1.2 something
`<h3> something </h3>` --> 1.2.1 something
`<h1> something else </h1>` --> no number something else
`<h2> something </h2>` --> 2.1 something else
`<h2> something </h2>` --> 2.2 something else
`<h3> something </h3>` --> 2.2.1 something else

fmaussion:
**In an ideal world**, I would like to have the following feature:

In the notebooks:
```
01-Notebook.ipynb
   # Notebook 1 long title
   ## Notebook 1 - Section 1
   ### Notebook 1 - Section 1 - Subsection 1

02-Notebook.ipynb
  <start counter notebook at 2>
   # Notebook 2 long title
[...]

PR #1252, opened 2018-02-24
[toc2] basenumber

jfbercher:
Enable to define a number to begin numbering with. 
Useful to continue numbering over several chapters defined in separated notebooks. This should be preserved while exporting with `jupyter nbconvert --to html_with_toc FILE` or using `toc2.tpl` template. This addresses #1245 of @fmaussion

image

image

### DIFF
--- a/src/jupyter_contrib_nbextensions/nbextensions/toc2/toc2.js
+++ b/src/jupyter_contrib_nbextensions/nbextensions/toc2/toc2.js
@@ -37,6 +37,7 @@
         number_sections: true,
         sideBar: true,
         skip_h1_title: false,
+        base_numbering: 1,
         title_cell: 'Table of Contents',
         title_sidebar: 'Contents',
         toc_cell: false,
@@ -564,7 +565,8 @@
                 break;
             }
         }
-        for (var i = min_lvl; i <= 6; i++) {
+        lbl_ary[0] = cfg.base_numbering-1 // begin numbering at base_numbering
+        for (var i = min_lvl+1; i <= 6; i++) {
             lbl_ary[i - min_lvl] = 0;
         }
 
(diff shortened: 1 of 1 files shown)


fmaussion:
Thanks so much for looking into this! 

I gave it a try, and it works partly for me:
- there is no new option in the extension configuration tab: however, I added a ``"base_numbering": 2,`` option in the notebook's ``toc`` section manually for each file **and this worked fine**. Altogether I'm not sure to understand if a general option is the good way to do this, since this numbering is very notebook specific.
- the rendering **in the notebook** is good:
selection_019
- however the rendering **in html** is strange: the ToC entry is fine, but the sidebar toc and the titles are still keeping their old numbering. See:
selection_020

Maybe it's only me though (I don't know if I have to update something else after changing the code as you show in your PR).

jfbercher:
Sorry... 
I mixed myself with the screenshots. First one was an attempt to change numbering by the nbextension configuration tab. 
Actually, I ended with updating the `base_number` *per notebook* (click on the ToC settings icon on the toc window/sidebar) -- see below. I also update the PR description above. 

image

jfbercher:
@fmaussion 
Rendering after conversion loads the `toc2.js` in the github repo. 
This will be only available after the PR is merged. This explains why the rendering in html in not fully updated (the `toc2_cell` is embedded in the notebook itself while the rest -- sidebar, numbering is computed by the `toc2.js` script)

fmaussion:
Perfect, thanks! Looking forward to see this merged then ;-)

fmaussion:
Any chance this can get merged before Sunday? If not it's OK, I'll just have to deal with it: my lecture starts next week and I can still use some workarounds for now.

jcb91:
Looks good to me :+1:

PR #1253, opened 2018-02-25
Remove OSX travis run

juhasch:


jcb91:
:+1:

josephcslater:
Does this mean there will no longer be OSX support?

jcb91:
> Does this mean there will no longer be OSX support?

No, OSX is similar enough to linux that the CI builds there are pretty much sufficient.

The main issue here was that our OSX Travis runs were slow to start, and were running into problems with the initial python setup, which was causing PRs to look like they had breaking changes all the time, when in reality they didn't. Since we don't have object libraries, and most of our file-location stuff is handled by `jupyter_core`, we don't have much in the way of platform-dependent code anyway. However, it would be nice to re-add OSX runs eventually, even if only for the added peace of mind.

Hope that makes sense?

josephcslater:
Perfect. As you know from my recent issue... I have one that I can't trace and am hoping it's figured out eventually (or my system inexplicably continues to work).

PR #1255, opened 2018-03-01
[toc2] fix css rule for overriding wrapper color

jcb91:
This is a fix for https://github.com/Jupyter-contrib/jupyter_nbextensions_configurator/issues/59, which turned out to actually be an issue with toc2


jcb91:
(essentially, the override was using the wrapper _class_, which is less specific than, and therefore overridden by, the wrapper _id_)

PR #1266, opened 2018-03-19
Toc2 history

jfbercher:
This enables to move in toc2 history of jumps in the notebook, using the forward-back browser buttons. 

It follows a discussion in https://github.com/ian-r-rose/jupyterlab-toc/issues/1 and an idea by @jcb91.


idoDavid:
@jfbercher 
Would love to test that, but isn't cloning the repo and installing only installs what's on the master branch? if not, how do I activate the extension? 

Bottom line didn't succeed to review locally your work :/

jfbercher:
Download [this file](https://raw.githubusercontent.com/jfbercher/jupyter_contrib_nbextensions/ee9eed57fd1623134cdde9d88f13c4f61b8d0132/src/jupyter_contrib_nbextensions/nbextensions/toc2/toc2.js) to your nbextensions/toc2 directory (exact location depends on your system/install) , see the data paths via `jupyter --paths`). You may save the previous version before overwritting, in case of... 
Then reload a notebook, the extension should be updated.

jfbercher:
It is probably  `/anaconda3/share/jupyter/nbextensions/toc2` (otherwise copy it everywhere - some of these paths might be soft links though). 
This only works in classical notebook, not jupyterlab, where the architecture is different.

idoDavid:
@jfbercher after reloading, the toc loads but no longer clickable, i.e. pressing the toc links don't take you anywhere.

jfbercher:
Reproducible. My fault. This happens when the history is empty, which is the cas when you load a notebook in a new tab. Corrected in the last commit. Can you check if it works @idoDavid ?

jfbercher:
@idoDavid have you checked if it works as intended? Any other comment?

idoDavid:
@jfbercher hi, not yet, would love to look it now. same drill? copy the file into the proper places and refresh?

idoDavid:
@jfbercher works like a charm!

PR #1268, opened 2018-03-20
Long labels in code_prettify/2to3

jfbercher:
And others derived from `kernel_exec_on_cells`. 

Remove the label from the `add_buttons_group` and add it back "by hand" to enable the bubble help. 
Fixes #1243 (shift-clik works for me) and last comment in #1102.


jcb91:
Thanks @jfbercher. Is the space trick really necessary? `label` can't be `undefined`, or an empty string?

jfbercher:
@jcb91. It seems necessary. Unfortunately, both `label: '`' and `label: undefined` fail with `kernel_exec_on_cell.js?v=20180320174230:344 [code_prettify] error loading: TypeError: Cannot read property 'help' of undefined
    at toolbar.js:97`. I haven't investigated what happens in `toolbar.js`

jcb91:
Ok, good enough for me :)

PR #1272, opened 2018-03-27
[toc2] add controls for all per-notebook metadata options

jcb91:
inspired in part by https://stackoverflow.com/questions/47726971


jfbercher:
Looks good to me. 
Thanks @jcb91!

PR #1290, opened 2018-05-03
pandas: Split-apply-combine (2 funcs: count, mean)

olisteadman:
A classic pandas routine which takes a Series/column, summarises it (counts up all the values of a certain type, or takes a mean of all the numeric values) and appends the summary as a new Series/column. See related answer on [StackOverflow](https://stackoverflow.com/questions/28353577/merging-and-sum-up-several-value-counts-series-in-pandas/49678890#49678890).


jcb91:
thanks!

PR #1296, opened 2018-05-23
Enable ruler extension in editor

juhasch:
Make ruler extension work in notebook and editor


sh0375:
Thank you very much! It solves my problem. It is working!
Small problem: css patch is turned off for editor - lines are displayed with offset.

PR #1301, opened 2018-06-05
updated readme - toc cell config note

laholmes:
added an explanatory note detailing how to configure placement of toc cell in notebook


juhasch:
Thank you.

PR #1307, opened 2018-06-12
Extension of toc2 in order to implement table of figures

ptitdoc:
Hello,

This is a rework of my previous pull request because the code base have since quite changed:
https://github.com/ipython-contrib/jupyter_contrib_nbextensions/pull/854

The goal of this pull request is to implement table of figures. The typical use case is to add a table on the beginning of the notebook highlightly important items that could be figures, issues, todos, recommendations ...

The current implementation add sufficient settings to replace the table of content with a table of figure. This include:
section_number_prefix: in order to add a prefix to numberized sections (eg: Figure1, Figure2 instead of 1, 2 ...)
analyse_level: in order to make the hierachy between section flat.
dom_search_pattern: in order to list arbitrary dom items instead of headings. My typical use cases are listing  `caption` for figure captions or to list tags based on a specific user defined class such as `.todo` for html items such as `<p class="todo">`.

### DIFF
--- a/src/jupyter_contrib_nbextensions/nbextensions/toc2/toc2.js
+++ b/src/jupyter_contrib_nbextensions/nbextensions/toc2/toc2.js
@@ -46,6 +46,7 @@
         toc_window_display: false,
         section_number_prefix: '',
         analyse_level: true,
+        dom_search_pattern: '[id]:header',
     };
     $.extend(true, default_cfg, metadata_settings);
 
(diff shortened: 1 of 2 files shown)


ptitdoc:
This pull request will be the base to make table of content event more generic:

Currently, it is only possible to change settings in order to use the table of content as a table of figures.

The goal is to be able to build more than one table of content in a single notebook. This however became slightly more complicated since the removal of `st`.

PR #1330, opened 2018-08-09
Updated annotations extension and added to master branch

pyython:
I really like the old usability/annotations extension on the 3.x branch.  I decided to update it and add in some enhanced functionality - support for custom colors for each commenter, the ability to 'resolve' a comment thread, and the ability to launch comment functionality just by clicking on the speech bubble icon.

I tested this on my 5.6.0 Jupyter server and seems to be working well.


pyython:
The test suite appears to be broken.  It's calling pip.main() which seems to be deprecated.

juhasch:
Thank you. Two things I noticed:
* there are two unrelated binary files (`.DS_store')
* The structure of the extension does look outdated, .i.e. you should return a public function for initializazion and using global variables like `IPython.notebook.metadata` is also deprecated. Please take a look at https://jupyter-contrib-nbextensions.readthedocs.io/en/latest/internals.html

PR #1337, opened 2018-08-22
Added New nbextension - provbook:Provenance of the Notebook

Sheeba-Samuel:
This PR implements an extension to Jupyter Notebook to capture and view its provenance over the course of time. This extension displays the provenance of each notebook cell execution. When the extension is enabled, the visualization of the provenance data is displayed below the input of every cell. The user can view the history over the course of time by moving the slider. As you move the slider, you could see the start and end time of each execution, how long it took to execute, the number of runs, the source, and output of each execution. It also allows the user to see the changes and the modified time of the text of a markdown cell whenever a notebook is saved. In this way, the user can compare the previous results with the current ones and see the difference that occurred. The user has the option to view the provenance of selected or all cells as well as clear them if needed.
The figure below shows the provenance of a code cell along with its input and output.
provbook_provenance


juhasch:
Thank you. Looks good to me but I honestly cannot test all extensions properly.
I will merge, unless you say you discovered a problem in the meantime.

Issue #1339, opened 2018-08-26
[toc2] Why is toc2 limited to input areas?


mpschr:
I quite often produce various outputs from a line of code which I structure with `<h>` elements for better navigation as follows:

```
from IPython.core.display import HTML, display
for x in somelist:
    t = *some code*
    display(HTML(f"result table {x}"))
    print(t)

```

I was surprised that the toc2 would not pick up on them.

Checking the code, I see that the header-lookup is limited to [$('.text_cell_render')](https://github.com/ipython-contrib/jupyter_contrib_nbextensions/blob/4906812149dcaa66250304ba541c24612dc74fa6/src/jupyter_contrib_nbextensions/nbextensions/toc2/toc2.js#L560). Is there a specific reason for that? Or could this extended easily to `output_subarea` elements ? 

I tried to play around a bit, but I don't understand the piece of code selecting the headers (`.find('[id]:header:not(:has(.tocSkip))');`)

gwerbin:
I think this is by design. My understanding of the notebook workflow is that you use a Markdown input cell to put a heading in the notebook, then you have the code generating that heading's plot underneath it. I can see why that's limiting for your case.

mpschr:
Yes, I used markdown in first place but thereafter assumed that all headers tags in a notebook would be captured. 

I could work out a solution for the report I was working on: for any one interested, it's in the referenced pull request, above. I imagine it could be of wider interest to incorporate the feature.

PR #1340, opened 2018-08-26
Update toc2.js

mpschr:
Adding `.output_subareas` so that user-made `<h>` elements with `id` are also being picked up. See #1339 for more detail


mpschr:
I just noticed that the user-created `<h>` have no `<a>` of `.anchor-link` class, which renders them unclickable when exported. They work in a live notebook!

I cannot figure out where the  `.anchor-link` class is coming from. Is it created by `toc2` or is it always included within `<h>` tag in a `ipynb`

mpschr:
Ok - I have adapted the function `make_link` in order to make it work. Hopefully somebody can test that all works fine and accept the pull request.

Is there anything needed for the documentation?

PR #1364, opened 2018-10-23
[gist-it] Making github endpoint configurable to support publishing gists to Github Enterprise

shiti-saxena:


shiti-saxena:
@jcb91 do you think this might be useful?

PR #1365, opened 2018-10-24
fix cell widen bug

kingbase:
The cell was not widen as there is an unnecessary check.

While sidebar is visible and widenNotebook is enabled, the `available_space` is `nb_wrap_w - 2 * margin - sidebar_w`, and I'm not sure why `available_space` was compared to `nb_inner_w + sidebar_w`. It seems that `sidebar_w` was calced twice. This check cause the width was not automatically widen based on the page **with the default installation of jupyter**.
And I'm not sure what does the check mean, so I just removed the check and it worked fine for me.


juhasch:
Thank you.

Issue #1334, opened 2018-08-19
navigation-hotkeys: please remove Esc toggle


stas00:
It took me a while to hunt this one down. This really useful navigation-hotkeys extension adds a problematic hotkey: `esc - toggle to edit mode`, which is already assigned to `switch to command mode`. This breaks a normal use flow where one would hit Esc to switch from any mode to Command mode followed by a custom command mode keyboard shortcut.

Once this extension is enabled `Esc` behavior becomes non-deterministic and if the current cell is not visible one can't tell the current mode - should I hit Esc or not? So for example I end up typing `ii` when I'm trying to interrupt the kernel, since hitting `Esc-ii` happened to be from an already 'command mode', and moving me into edit mode and entering 'ii' to the cell.

Please consider removing this shortcut or making it configurable on the user side. At the moment I have to disable this extension and manually re-insert all but one shortcut from it.

Thank you.

JSpenced:
You might want to checkout my pull request fyi. I've fixed this issue as I removed it in my pull since it can be set in the notebook.json file with the default command. Also, I updated it so the command palette works right. Let me if it doesn't work on your pc.

stas00:
Thank you, @JSpenced.

I don't see any difference in its behavior. Esc still toggles back and force between edit/cmd modes.

I installed your branch:
```
pip3 install git+https://github.com/JSpenced/jupyter_contrib_nbextensions.git@update-for-jupyter-and-remove-redundant-keyspip3 install git+https://github.com/JSpenced/jupyter_contrib_nbextensions.git@update-for-jupyter-and-remove-redundant-keys
```
then enabled Navigation-Hotkeys, restarted jupyter. The hotkeys provided by this module work. And Esc is still a problem.

JSpenced:
For installation, you could just copy over the modified main.js file into your `nbextensions/navigation-keys` folder. Or copy the navigation keys to `nav-keys-modified` and replace the `main.js`. Activate that plug-in. I can't see why those would be combined into one binding as I completely removed the command from the file. Jupyter just overwrites bindings so at least one should be set I think.

Yeah, I was confused about those keybindings as they were written in a non-standard way in the javascript file. I wasn't really sure what they were originally because I wasn't using the plug-in before. I got rid of split-cell-at-cursor as that is redundant with the jupyter binding just as enter edit mode was binded to escape.

[...]

PR #1378, opened 2018-12-20
Update Navigation HotKeys Extension to work with command palette on new versions of jupyter 

JSpenced:
This updates so it doesn't use `Ipython.` anymore and the new `actions.register` and `add_shortcut` commands.

If you use this in the most recent versions of jupyter, it loads the functions improperly like this:
image

This has been fixed to have a proper group called `navigation_hotkeys`.

Also, second commit I removed two functions because those functions are already implemented in Jupyter as default bindings. They can be set in your `notebook.json` file, which I think is better.

#1334 Also solves this problem as this was one of the removed default bindings.


JSpenced:
The last commit makes it completely backwards compatible plus adds in so you can toggle some of the settings on and off to fix request #1334 .

juhasch:
Thank you for your contribution. Looks good to me.

PR #1390, opened 2019-01-28
making it work for remote (public) jupyter notebook

twpark:
removed 'localhost' from custom.css file path check so that it would work with remote access. Might be some other workarounds without doing this, though.


jcb91:
this should probably use require to resolve what the custom/custom.css URI should actually be, as otherwise we just match anything that ends like that. But, I think this satisfies a legitimate use case, so fair enough 👍

PR #1394, opened 2019-02-07
TOC2: catch headers in text in output cells created using Markdown()

hottwaj:
Make TOC2 also catch headers in text in output cells that was created using Markdown().

This allows programmatic creation of the TOC e.g. the following will now create an entry in the TOC when "run_alternative_scenario" is True

```
if run_alternative_scenario:
    display(Markdown('# Results from alt scenario')
```

Although this change makes TOC2 add headers from Markdown() text, the TOC is not automatically updated when such cells are added/changed. (You have to click the TOC2 "refresh" button to trigger an update)

Would appreciate help in getting automatic updates working!

Thanks

### DIFF
--- a/src/jupyter_contrib_nbextensions/nbextensions/toc2/toc2.js
+++ b/src/jupyter_contrib_nbextensions/nbextensions/toc2/toc2.js
@@ -583,7 +583,7 @@
         // excepting any header which contains an html tag with class 'tocSkip'
         // eg in ## title <a class='tocSkip'>,
         // or the ToC cell.
-        all_headers = $('.text_cell_render').find('[id]:header:not(:has(.tocSkip))');
+        all_headers = $('.text_cell_render,.output_markdown').find('[id]:header:not(:has(.tocSkip))');
         var min_lvl = 1 + Number(Boolean(cfg.skip_h1_title)),
             lbl_ary = [];
         for (; min_lvl <= 6; min_lvl++) {


jcb91:
Thanks @hottwaj for the contribution. Given that this changes the default behaviour though, I think it ought to be made optional with a parameter. Could you implement that?

hottwaj:
Hey sure I can check that out.  Any thoughts on the trigger that is needed to update the TOC when you press ctrl+enter on a cell that contains Markdown?  Thanks!

jfbercher:
@hottwaj Digging a little, I don't see a signal emitted when the `Markdown()` function is issued (nor any `display` function). 
"Ctrl-enter" trigs the 'execute.CodeCell' event, but listening to this one in addition to 'rendered.MarkdownCell' would add some overload (table of contents would be called for any cell exec). 
Perhaps could you try to fake markdown rendering by trying something like (not tried)
```
display(Markdown('# Results from alt scenario')
display(Javascript("Jupyter.events.trigger('rendered.MarkdownCell')"))
```
Otherwise I agree with @jcb91 that an option for this would be useful.

hottwaj:
@jfbercher Thanks for checking that out, maybe I can look into that if it becomes more necessary - at the moment the contents generally gets re-rendered at some point whenever a few cells are executed, so I haven't found it too problematic.

I haven't found time to look into making it an optional feature, but will try to (likely timescale: several weeks at least).

Cheers

hottwaj:
appreciate that this is 2 years later and likely this repo is not going to see more maintenance... but I finally made that minor change (added a menu option)!   Thanks

PR #1399, opened 2019-02-14
Add exporter name for toc2 nbconvert

wcable:
I've found myself in an unfortunate position where a list of 6 `custom (.html)` exporters appears in the "Download As" menu; perhaps I am not alone.

(All the other converters in the same directory should probably also get `export_from_notebook` display-name attributes, but I haven't looked into what they do / what appropriate names woudl be)


jcb91:
thanks!

Issue #1402, opened 2019-02-26
[toc2] toc2.js load failed with working in non-live notebook


zthxxx:
When use `jupyter nbconvert -y --to html --template toc2 ...` and view the html file in the browser,

and all titles which in toc wrapper are not collapsible, no fa-caret-down icon or cannot collapse by click.

![image](https://user-images.githubusercontent.com/15135943/53392030-1aadef00-39d3-11e9-8368-96897f436657.png)

In the console, i found this error :

```js
[toc2] working in non-live notebook

toc2.js:222 Uncaught TypeError: Cannot read property 'notebook' of undefined
    at create_navigate_menu (toc2.js:222)
    at Object.table_of_contents (toc2.js:651)
    at nbcv-test.html:11788
    at Object.execCb (require.min.js:29)
    at $.check (require.min.js:18)
    at $.enable (require.min.js:23)
    at $.init (require.min.js:17)
    at require.min.js:26
```

![image](https://user-images.githubusercontent.com/15135943/53391472-6c557a00-39d1-11e9-8698-156ddcc856c0.png)

some ref [source code here](https://github.com/jcb91/jupyter_contrib_nbextensions/blob/master/src/jupyter_contrib_nbextensions/nbextensions/toc2/toc2.js#L57)

zthxxx:
I just created a PR (#1403) to fix this bug.

Any suggestions?  @juhasch

PR #1403, opened 2019-02-26
fix: fix toc2.js load failed with working in non-live notebook

zthxxx:
check all `IPython` and `Jupyter` call with `liveNotebook` flag

ref issuse: #1402

### DIFF
--- a/src/jupyter_contrib_nbextensions/nbextensions/toc2/toc2.js
+++ b/src/jupyter_contrib_nbextensions/nbextensions/toc2/toc2.js
@@ -56,6 +56,11 @@
      */
     var read_config = function () {
         var cfg = default_cfg;
+
+        if (!liveNotebook) {
+            return cfg;
+        }
+
         // config may be specified at system level or at document level.
         // first, update defaults with config loaded from server
         $.extend(true, cfg, IPython.notebook.config.data.toc2);
@@ -643,7 +648,7 @@
         });
 
         // update navigation menu
-        if (cfg.navigate_menu) {
+        if (cfg.navigate_menu && liveNotebook) {
             var pop_nav = function() { //callback for create_nav_menu
                 $('#navigate_menu').empty().append($('#toc > .toc-item').clone());
             }


zthxxx:
hi~ @jcb91 @juhasch  I make a pr again~ 

thanks for your awesome extensions! I try to fix a little bug in `toc2.js`

more reason details see #1403 

plz review and merge,  Thanks alot.

zthxxx:
@jcb91 
Just a little change, 3 effective lines

zthxxx:
@jcb91   Any suggestions about this pr?

zthxxx:
@jcb91

src/jupyter_contrib_nbextensions/nbextensions/toc2/toc2.js
```diff
@@ -643,7 +648,7 @@
         });
 
         // update navigation menu
-        if (cfg.navigate_menu) {
+        if (cfg.navigate_menu && liveNotebook) {
```
jfbercher: I agree with that. That is the reason of the issue you encountered: if the navigate menu is enabled in the live notebook, then this config is stored in the notebook metadata and the js will try to create this menu even in the non live situation. 

Alternatively, we could also fix in hard the value `navigate_menu=false` in the toc2 template `toc2.tpl` which is used by nbconvert.

src/jupyter_contrib_nbextensions/nbextensions/toc2/toc2.js
```diff
@@ -56,6 +56,11 @@
      */
     var read_config = function () {
         var cfg = default_cfg;
+
+        if (!liveNotebook) {
+            return cfg;
+        }
+
```
jfbercher: This is not desirable. The idea (I am the author) was to be able to read the configuration stored in the notebook (line 66): the toc can be floating, the contents name may have been changed, etc

zthxxx: at line 66, the variable `IPython` referenced before assignment ([catch require do nothing with variable](https://github.com/ipython-contrib/jupyter_contrib_nbextensions/blob/7144c90048a33197755503064ab1a2a496480b11/src/jupyter_contrib_nbextensions/nbextensions/toc2/toc2.js#L84-L100))

Thus, the idea, read some config, is meaningless while in no-live-notebook

zthxxx:
@jfbercher 

at [line 66](https://github.com/ipython-contrib/jupyter_contrib_nbextensions/blob/7144c90048a33197755503064ab1a2a496480b11/src/jupyter_contrib_nbextensions/nbextensions/toc2/toc2.js#L66), the variable IPython referenced before assignment (catch require do nothing with variable, [line 84-99](https://github.com/ipython-contrib/jupyter_contrib_nbextensions/blob/7144c90048a33197755503064ab1a2a496480b11/src/jupyter_contrib_nbextensions/nbextensions/toc2/toc2.js#L84-L99))

Thus, the idea, read some config, is meaningless while in no-live-notebook

jfbercher:
Sorry, I was remembering falsely that the toc2 template (used for generating the html "non-live notebook" version) uses read_config, but that is not the case; It directly reads the config from notebook's metadata; So your analysis is correct and this fixes the mentioned issue. 

For completeness, it could also be useful to also add `cfg.navigate_menu=false;` in `toc2.tpl`  
```
        var cfg = {{ nb.get('metadata', {}).get('toc', {})|tojson|safe }};
        cfg.navigate_menu=false;
```
which also fixes the problem.

jfbercher:
Thanks for this @zthxxx !

zthxxx:
Thanks for your review! @jfbercher 

i think your solution is more fundamental.

Issue #1386, opened 2019-01-07
code_prettify: option to disable js alert for unsupported kernels


kafonek:
In [code_prettify/kernel_exec_on_cell.js](https://github.com/ipython-contrib/jupyter_contrib_nbextensions/blob/master/src/jupyter_contrib_nbextensions/nbextensions/code_prettify/kernel_exec_on_cell.js), there is an alert pop-up when the extension is loaded in a notebook that is not Python, R, or javascript; afterwards, the icon and keyboard shortcut is removed.  The alert shows up even when the option for "show alerts for errors in the kernel prettifying calls" is turned off.

Could a nbextension parameter please be included to handle that situation more "quietly", removing the icon and keyboard shortcut while not generating a js alert? 

Thank you.

jfbercher:
Fixed (in principle) by #1409

PR #1409, opened 2019-04-03
Code prettify (and all KernelExecOnCells plugins) --> option to disable alert if kernel is unsupported

jfbercher:
- added an option for displaying an alert if kernel is not supported and turned it off by default (instead issue a warning in the js console). This addresses #1386 @kafonek   @riordan
  - also, in passing, corrected an issue in configs merge.


jfbercher:
Interested people shoud have had the time to test this. 
Thus merging for now. 
Open an issue if necessary.

PR #1410, opened 2019-04-03
[toc2] Updates

jfbercher:
- Complement fix for issue #1402 by disabling Navigate menu in toc2 template (Navigate menu  cannot be shown in non-livenotebook conditions)
- Changed toc2.py exporter to embed images in the html output, rather than extract them
- Updated history


jfbercher:
Interested people shoud have had the time to test this. 
Thus merging for now. 
Open an issue if necessary.

Issue #1275, opened 2018-03-30
Large variables cause varInspector to slow down entire notebook.


babymastodon:
Hey, I found a bug in varInspector/main that causes the entire notebook to slow down in certain situations. The fix is pretty easy -- I might send a patch if I find some time.

Steps to reproduce:

- Enable varInspector/notebook extension, and open a new notebook
- Read a 4 GB file into memory and store it as a variable (with open("filename") as f: raw_data = f.read())
- Try to execute any other cell in the notebook (for example, print("hello world")). After the cell gets executed, the notebook gets stuck in the "queued" state for about 5 to 10 seconds. Also, CPU usage for the kernel gets pegged at 100%.
- use pyflame to identify that the notebook is getting stuck in the "var_dic_list" function

![image](https://user-images.githubusercontent.com/950674/38122629-fc308118-338a-11e8-8716-cdd15df5b2ec.png)

Impact:

Basically, if there is an extremely large variable in the notebook, every single cell will take a very long time to execute (> 5 seconds), no matter how simple the cell contents are.

Root cause:

Here's the code to the offending function: https://github.com/ipython-contrib/jupyter_contrib_nbextensions/blob/master/src/jupyter_contrib_nbextensions/nbextensions/varInspector/var_list.py#L41

```
'varContent': str(eval(v))[:200]
```

Culprit is the str() function. It serializes the entire variable into the string, even though only it only needs the first 200 variables.

Potential fix:

Replace the str() function with something smarter. For example:

```
def get_fast_repr(obj, length):
    if isinstance(obj, six.string_types):
        return str(obj[:length])
    if isinstance(obj, list):
        ...
   if isinstance(obj, dict):
      ...
   etc for all built-in types
```

ijoseph:
This extension is unusuable for me until this is fixed; the entire reason for using this for me is to help manage my memory by noticing large `DataFrame`s that I no longer need but forgot to `del`. However, this makes the entire notebook excruciatingly slow all the time, which defeats the purpose.

franzoni315:
[...]

PR #1411, opened 2019-04-06
[varInspector] updated var_list.py to address #1275

jfbercher:
Tentative fix to #1275 
As analyzed by @babymastodon for large variables, printing the beginning of content of the variable using str(var) [:200] still converts the full content to str even though only some characters are printed. For very large variables, this causes a slow down of the notebook each time varInspector is called. 
The fix consists in testing if the variable has a length property and then extracts the first elements of the variable before converting it to string. 
Tested here with a 2Gb file read as a binary variable without noticing any slow down. 
Further tests welcome.


jfbercher:
I think that interested people shoud have had the time to test this. 
Thus merging for now. 
Open an issue if necessary.

PR #1427, opened 2019-06-30
Fix command to install using pip

hsm207:
See https://pypi.org/project/jupyter-contrib-nbextensions/


hsm207:
@DrRogger The issue I encountered was without the underscore, pip will throw an error saying it cannot find the package. However, if I use the pip that came with Anaconda, this problem does not appear.

PR #1452, opened 2019-08-30
Use `black` formatter by default

lucatrv:
Based on: https://gist.github.com/MarvinT/a072aa992e977496974aaf492287b08c#file-black_code_prettify-json

### DIFF
--- a/src/jupyter_contrib_nbextensions/nbextensions/code_prettify/code_prettify.js
+++ b/src/jupyter_contrib_nbextensions/nbextensions/code_prettify/code_prettify.js
@@ -24,16 +24,17 @@ define(['./kernel_exec_on_cell'], function(kernel_exec_on_cell) {
     cfg.kernel_config_map = { // map of parameters for supported kernels
         "python": {
             "library": ["import json",
-            "def yapf_reformat(cell_text):", 
-            "    import yapf.yapflib.yapf_api",
-            "    from yapf import file_resources",
-            "    import os",
+            "library": ["import json",
+            "def black_reformat(cell_text):",
+            "    import black",
             "    import re",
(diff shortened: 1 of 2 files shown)


jcb91:
Although I like this idea, it looks like the black API has changed recently, so this doesn't always work. On my current installation (black 19.3b0), this throws 

    TypeError: format_str() takes 1 positional argument but 2 were given

while the method signature appears to expect a filemode dataclass object (see 
 https://github.com/psf/black/blob/19.3b0/black.py#L664). It looks like your version would have worked with the previous API from 18.9b0 (see https://github.com/psf/black/blob/18.9b0/black.py#L621-L623). I guess this is what we get for black still being "beta". A small try/except could probably cover both versions?

Thanks for the PR, btw 👍

lucatrv:
That's strange, it's working on me with black 19.3b0. I'm on Windows 7, with Miniconda on "conda-forge" channel.

jcb91:
Hmm, curiouser & curiouser. I'm on macOs, using 19.3b0 from PyPi. I'll try to give it a shot on windows at some point. Alternatively, perhaps it would be worth adding python tests for these snippets...

shunkakinoki:
Really great idea!

lucatrv:
Now it works with the new Black API.

inarighas:
Thanks for this PR. Is it ready to be merged ?

lucatrv:
I think it has been for a while already...

inarighas:
For jupyter lab, it s the case (on jupyterlab-code-formatter extension).But,  I dont believe it is the case for jupyter notebooks. code-prettify still use yapf by default, and autopep8 uses autopep8 ofc ... Using black to format still need manual tweaking.

Issue #1461, opened 2019-09-28
zenmode: not working properly with jupyter themes


luckylsk34:
Hello there,

Toggling zenmode off while using jupyter themes doesn't restore the previous background and CSS properties. As a result, the following happens.

before using zenmode:
![Screenshot_2019-09-28 Untitled](https://user-images.githubusercontent.com/25878722/65814967-b3754900-e206-11e9-841e-0ca47305e807.png)

toggling zenmode on:
![Screenshot_2019-09-28 Untitled(1)](https://user-images.githubusercontent.com/25878722/65814976-c9830980-e206-11e9-846d-1fb5d635a6fe.png)

toggling zenmode off:
![Screenshot_2019-09-28 Untitled(2)](https://user-images.githubusercontent.com/25878722/65814980-cf78ea80-e206-11e9-9546-9914dc125559.png)

Toggling zenmode off is expected to revert to the first image but as it shows in the third image, it does not.
I am not using any URLs for background images, my "Use builtin backgrounds in addition to any specified by URL" is checked OFF in zenmode settings. The grey background around the code cell also stands out from the body's background.

luckylsk34:
Maybe we can leave the background as is when no images or urls are used.

luckylsk34:
Closed by #1462 .

PR #1462, opened 2019-09-28
Fix for issue #1461

luckylsk34:
Fixed the gray background for the notebook-container problem by making the opacity 0. Commented out cell CSS to make it more suitable to the specified theme.

When zenmode it toggled on, the old background is no longer stored and nothing is stored in it anyway, and the same properties are stored in respective properties.
When zenmode is toggled off, the image is removed and nothing is changed.

Since we are making notebook-container transparent when toggling zenmode on and not changing it back when toggling zenmode off, this will not completely restore the look of the original theme for some themes where the notebook-container has a different background than the normal background.


juhasch:
Thank you.

PR #1474, opened 2019-11-12
Fix duplicate call of extension "init_cell"

juice500ml:
There exists a possibility of calling function `run_init_cells()` twice on function `run_init_cells_asap()`.
Kernel exists only as two cases: (1) Kernel is already ready, and (2) Kernel is not yet ready, and in both cases, `run_init_cells()` should be called only once.
On case (1), there is no need to hook an event on `kernel_ready`, as cells are already initialized.
If the event hook exists, in some specific cases, `run_init_cells()` is called twice, due to racing conditions.


juhasch:
Makes sense, thanks for catching this bug.

PR #1514, opened 2020-07-02
Added Goto Error extension

teticio:
Hi

I've created an extension which makes it possible to click on the filenames in a stack trace and open up a window showing you the code with the line highlighted, even if this is buried in some installed pacakge. I find this very useful for debugging and to understand better how other packages work. I got the idea from Google Colab, which has similar functionality. Also, a button "Search on Stack Overflow" is added to the stack trace which will search for your error on Stack Overflow.

I'd be really chuffed if this got included in the offering.

Thanks everyone for a great repo!

Rob


teticio:
Hello. Just wondering if there is any chance of this being merged... It would kinda make my day. Thanks!

PR #1535, opened 2020-10-20
Support non-live notebook such as nbconvert output

cailiang9:
For long notebook snapshots, we need toc as a quick navigation.
a simple way is to render with toc2 as a nbconvert template.
but when open the rendered html with browsers, we found some js errors because toc2 does not support non-live notebook.
this PR fixes the errors and supports non-live notebook.


juhasch:
Thank you

PR #1599, opened 2022-07-09
Bump version to 0.6.0

juhasch:


bnavigator:
0.6.0 has never been tagged or published on PyPI

PR #1532, opened 2020-09-29
replace template_path with template_paths

demotu:
Nbconvert 6.0 replaced 'template_path' with 'template_paths' (see https://nbconvert.readthedocs.io/en/latest/changelog.html#significant-changes).
This change in Nbconvert causes errors in jupyter_latex_envs and in jupyter_contrib_nbextensions (see  https://github.com/ipython-contrib/jupyter_contrib_nbextensions/issues/1529).

Replacing all instances of 'template_path' with 'template_paths' in both jupyter_latex_envs and in jupyter_contrib_nbextensions fixes the error.

PR #1601, opened 2022-07-12
Remove dependency on jupyter_latex_envs

juhasch:
Should "fix" the last "Config option template_path not recognized by LenvsLatexExporter. Did you mean one of: template_file, template_name, template_paths" warning message remaining from Issue #1532.

@jfbercher : I would remove this dependency, i.e. automatically installing your extension if you have no objections. It can be installed seperately, anyway.


danetrata:
The `jupyter_latex_envs` module is still being imported in other parts of the code.

PR #1604, opened 2022-08-18
fix typos

astromancer:
Hi! Small PR to fix a couple of typos in the install script.


juhasch:
Makes sense. Thank you.

PR #1612, opened 2022-11-14
Remove the `scripts` portion of setup.py

bkad:
Scripts is deprecated by entry_points and when both exist it causes errors for some PEP-compliant installers.

Some details here: https://github.com/bazelbuild/rules_python/issues/765#issuecomment-1193000233


juhasch:
Thank you.

PR #1614, opened 2022-12-03
`limit_outputs`: do not truncate HTML outputs

deeenes:
- Truncating HTML (such as pandas data frames) resulted in unclosed tags and the corrupted HTML was often illegible below the affected cells
- New Boolean parameter `limit_html_output`
- New default behaviour is to keep outputs intact if they have HTML content


juhasch:
Makes sense, thank you.

PR #1616, opened 2022-12-06
Reconcile format for absolute time

### LINKED ISSUE #1615: ExecuteTime's default absolute format disagrees with its documentation and description file
Both the documentation and description file for *ExecuteTime* give `'YYYY-MM-DD HH:mm:ss'` as the default format for displaying absolute time:

https://github.com/ipython-contrib/jupyter_contrib_nbextensions/blob/92bb755152c870caf9cac1acebb168f71f9dae6c/src/jupyter_contrib_nbextensions/nbextensions/execute_time/readme.md?plain=1#L76-L81

https://github.com/ipython-contrib/jupyter_contrib_nbextensions/blob/92bb755152c870caf9cac1acebb168f71f9dae6c/src/jupyter_contrib_nbextensions/nbextensions/execute_time/ExecuteTime.yaml#L30-L33

However, the actual default in the code places the time before the date:

https://github.com/ipython-contrib/jupyter_contrib_nbextensions/blob/92bb755152c870caf9cac1acebb168f71f9dae6c/src/jupyter_contrib_nbextensions/nbextensions/execute_time/ExecuteTime.js#L39

To use `'YYYY-MM-DD HH:mm:ss'`, one must explicitly set that format in a configuration file. Trying to set it within the configurator web page has no effect, because the page detects no change from the displayed value.

stanwest:
Closes #1615 by making the default format consistent with that in the documentation and description file.

### DIFF
--- a/src/jupyter_contrib_nbextensions/nbextensions/execute_time/ExecuteTime.js
+++ b/src/jupyter_contrib_nbextensions/nbextensions/execute_time/ExecuteTime.js
@@ -36,7 +36,7 @@ define([
         clear_timings_on_clear_output: false,
         clear_timings_on_kernel_restart: false,
         default_kernel_to_utc: true,
-        display_absolute_format: 'HH:mm:ss YYYY-MM-DD',
+        display_absolute_format: 'YYYY-MM-DD HH:mm:ss',
         display_absolute_timings: true,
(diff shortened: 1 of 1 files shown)


juhasch:
Thanks. Makes sense.

PR #1617, opened 2022-12-09
In Collapsible Headings, get notebook version from sys_info if possible

### LINKED ISSUE #1611: Collapsible Headings alters vertical position of tooltip for signature and docstring
With the Collapsible Headings extension enabled in versions 6.5.1 and 6.5.2 of [*notebook*](https://github.com/jupyter/notebook), but not in v. 6.4.12, the tooltip that displays the current object's signature and docstring when one hits Shift + Tab appears at the wrong height within the browser window. As far as I've seen, the tooltip appears too low when the view is scrolled near the top of the notebook and too high when near the bottom. The screen-shots below illustrate the behavior.

low

high

For a long enough notebook scrolled down somewhat, the tooltip may have so high a position that it is outside the window, giving the impression that the notebook is ignoring Shift + Tab (at least until the fourth press raises the pager).

Prior related issues are  #509 and #949, and the possibility of breakage is [mentioned](https://github.com/ipython-contrib/jupyter_contrib_nbextensions/blob/2248f6e15753d0eb4339215313ae21ba09cbdf74/src/jupyter_contrib_nbextensions/nbextensions/collapsible_headings/readme.md?plain=1#L85-L91) in the extension's *readme* file.

stanwest:
Closes #1611.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@ readthedocs stuff gets all the shiny link goodness. Thanks!
 
 Unreleased (aka. GitHub master)
 -------------------------------
+
+- In *Collapsible Headings*, get the notebook version from `sys_info` if possible, resolving an incompatibility with *notebook* versions 6.5.x [#1617](https://github.com/ipython-contrib/jupyter_contrib_nbextensions/pull/1617).
+
 0.7.0
 -----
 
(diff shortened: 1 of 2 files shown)


juhasch:
Thanks @stanwest. Looks good to me.

PR #1649, opened 2023-08-30
Update for notebook 7.x

charmoniumQ:
Fixes `ModuleNotFoundError`; some submodules of `notebook` that moved to `jupyter_server` were finally removed in `notebook` 7.x. We try importing from `notebook` and fall back on `jupyter_server` (perhaps this should be the other way around?).


WesleyTheGeolien:
@charmoniumQ I think we can use the shim_notebook provided by Jupyter please see #1657

